PR #10, opened 2015-10-23
Wip/bewest/dev

bewest:
Tools to work with Nightscout.

These tells help translate pump-history to the treatments format Nightscout suspects.

It has a couple of nice features, including the ability to join pair wise `TempBasal` and `TempBasalDuration` commands, as well as the ability to combine `BolusWizard` events that are nearby their `Bolus` events into a single treatment event.  The combined bolus events allow classifying into *Correction* vs *Carb* vs *Meal* treatments, which is kind of nice.

I also demonstrate how other people can augment/extend these capabilities by enhancing the `notes` field.

### DIFF
--- a/bin/mm-format-ns-profile.sh
+++ b/bin/mm-format-ns-profile.sh
@@ -1,0 +1,128 @@
+#!/bin/bash
+
+# Author: Ben West
+
+SETTINGS=${1-monitor/settings.json}
+CARBS=${2-monitor/carb-ratios.json}
+BASALRATES=${3-monitor/active-basal-profile.json}
+SENSITIVITIES=${4-monitor/insulin-sensitivities.json}
+TARGETS=${4-monitor/bg-targets.json}
+OUTPUT=${2-/dev/fd/1}
+# DIA
+# CARBRATIO
+#TZ=${3-$(date +%z)}
+
+function usage ( ) {
+cat <<EOF
+EOF
+}
+
+function dia ( ) {
+  cat $1 | json insulin_action_curve
+}
+
+function fix-schedule ( ) {
+  field=$1
+  json -e "this.time = this.start; this.value = this.$field.toString( );"
+}
+
+function carb-ratios ( ) {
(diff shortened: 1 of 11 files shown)


bewest:
image

bewest:
https://github.com/nightscout/cgm-remote-monitor/blob/dev/lib/client/careportal.js#L57-L75
Thanks, for the pointers, @jasoncalabrese 
image

### treatment api notes
It seems like there are a few rules:

`Meal Bolus`, `Correction Bolus`, and `Temp Basal` are special keywords for `eventType`.  I tried using `Temp Basal Start` but couldn't get it to work, even though the code looks like it should.

If there is a `glucose` property on a `Bolus` record, the classic dual hemisphere won't work.

### oref0/Medtronic notes

* attempts to join paired `normal`, `square` bolus type records, including optional `wizard` into single treatment events
* translates into `Meal`, `Correction`, and `Carb Correction` events depending on combined insulin/carb reporting.
* annotates the `notes` field with description of some of the interesting fields.

jasoncalabrese:
Nice need to figure out why setting `glucose` isn't working for you.  It does work from the careportal, maybe `glucoseType` or `units` is needed.

an example from the school nurse today:

```
{ "_id" : ObjectId("562a88029ded9683421ebc5f"), "enteredBy" : "Kris", "eventType" : "Meal Bolus", "glucose" : 148, "glucoseType" : "Finger", "carbs" : 35, "insulin" : 2.05, "units" : "mg/dL", "created_at" : "2015-10-23T18:45:00.000Z" }
```

scottleibrand:
I'm successfully looping with this branch on my dev Pi and test pump, so I'm ready to merge it to dev if you are, @bewest.

bewest:
Some more pictures for fun:
These are only possible because I'm uploading manually, still need some more tools to figure out last treatment, and cull the set since then.

Imgur

Imgur

Imgur

bewest:
openaps evening out

PR #11, opened 2015-10-25
Wip/bewest/dev

bewest:
more tools to manage an NS site.  May not belong here, we'll see.


scottleibrand:
I think this is ready to merge into dev.  Anything else you can think of we need to fix first?

scottleibrand:
@bewest / others: if no objections by tonight, I'll go ahead and merge these so we can get folks like @tghoward uploading without having to run a wip branch.

PR #14, opened 2015-10-28
Reset git

scottleibrand:
Script to Delete git lock / history if necessary to recover from corrupted .git objects


package.json
```diff
@@ -26,6 +26,7 @@
     "oref0-mint-max-iob": "./bin/oref0-mint-max-iob.sh",
     "oref0-ifttt-notify": "./bin/oref0-ifttt-notify",
     "oref0-reset-usb": "bin/reset-usb.sh",
+    "oref0-reset-git": "bin/reset-git.sh",
```
bewest: Let's rename the file so everything matches.

PR #24, opened 2015-11-16
ns-uploader-setup.sh

scottleibrand:


jmatheson:
Feedback from my run through using the script tonight
* had to create directory ~/src then put oref0 in there (due lines 44/45)
* had to run sudo easy_install openaps-contrib (for timezones install to work)
* assumes named our devices pump and cgm (I did not, so it created extras)
* typo line 90: should read read_insulin_sensitivies
* may consider including instructions to create hashkey “echo -n "<secret>" | sha1sum”

If these all sound reasonable I can take a stab at cleaning them up.

scottleibrand:
Actually read_insulin_sensitivies is the typo'd version. easy_install -ZU the latest openaps and it should be fixed. 
Looks like I was assuming a dev setup, so need to add openaps-contrib to the "normal" oref0 install script perhaps, and figure out a better way to reference the oref0 install directory instead of ~/src. 
The pump and cgm thing is "as designed": this is really for a brand new setup, not an existing custom one. If you can think of an elegant way to handle arbitrary custom setups, I'm all ears, but that was out of scope for v1. 
Happy to support user entry of unhashed secret if you want to do the patch. 

-Scott

    _____________________________

jmatheson:
Only thought would be to check for existence of openaps.vendors.medtronic / dexcom as existing device. But, that assumes you only have one of each device. Clean install is probably fine, just initial docs send you down a setup and play with it path before getting here.

scottleibrand:
Yeah, I'm thinking we'll need to have this script either be much earlier in the docs, or a separate path for upload-only. 

-Scott




On Mon, Nov 16, 2015 at 10:46 PM -0800, "James Matheson" <notifications@github.com> wrote:












Only thought would be to check for existence of openaps.vendors.medtronic / dexcom as existing device. But, that assumes you only have one of each device. Clean install is probably fine, just initial docs send you down a setup and play with it path before getting here.

scottleibrand:
Fixed the reference to ~/src/oref0 and also pushed https://github.com/openaps/docs/commit/3fc24c19b9c0060883ec0cc09ea367a4d8e9ea46 to install openaps-contrib in the initial quick-packages.sh setup script.  I think this is now ready to merge.

danamlewis:
+1 ready for merge with some fixes I gave @scottleibrand verbally on crontab inclusion. 

Note: this should not be included or referenced in main openaps doc workflow, recommending this instead only for advanced users who have already finished phase 1 and may be on a second build.

PR #27, opened 2015-11-18
a couple basic IOB test to make sure 3hr and 4hr dia's work

jasoncalabrese:


scottleibrand:
I like it.  Should we also split the 4h and 3h DIA stuff into two functions?  Right now they're all under `it('should calculate IOB using a 4 hour duration', function() {`

jasoncalabrese:
think after3hDIA* is just named wrong, should be after3h*, after4h*

scottleibrand:
Ok, in that case we should add some more tests to capture all the different DIA tests we were doing today.  @eyim might want to help with that...

jasoncalabrese:
yeah, main this was setting dia to 4h, then making sure there was some left at 3h, and none at 4h

Would be great to have some more exact expectations for this

PR #35, opened 2015-12-04
A bit of formatting if/else

LorelaiL:
rt.reason avgdelta added toFixed
move console.log result
add if/else
add maxSafeBasal into reason

### DIFF
--- a/bin/oref0-determine-basal.js
+++ b/bin/oref0-determine-basal.js
@@ -52,11 +52,14 @@ if (!module.parent) {
         console.error(reason);
         return 1;
     }
-    console.error(JSON.stringify(glucose_status));
-    console.error(JSON.stringify(currenttemp));
-    console.error(JSON.stringify(iob_data));
-    console.error(JSON.stringify(profile));
+    console.log(JSON.stringify(glucose_status));
+    console.log(JSON.stringify(currenttemp));
+    console.log(JSON.stringify(iob_data));
+    console.log(JSON.stringify(profile));
     rT = determinebasal.determine_basal(glucose_status, currenttemp, iob_data, profile);
+    console.log(rT.reason);
+    console.log(JSON.stringify(rT));
+    rT=rT;
 
 }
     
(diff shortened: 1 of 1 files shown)


bin/oref0-determine-basal.js
```diff
@@ -146,8 +159,14 @@ function init() {
         var naive_snoozeBG = Math.round( naive_eventualBG + 1.5 * bolusContrib );
         // adjust that for deviation like we did eventualBG
         var snoozeBG = naive_snoozeBG + deviation;
-        console.error("BG: " + bg + tick + " -> " + eventualBG + "-" + snoozeBG + " (Unadjusted: " + naive_eventualBG + "-" + naive_snoozeBG + ")");
-        if (typeof eventualBG === 'undefined') { console.error('Error: could not calculate eventualBG'); }
+        
+        console.log("BG: " + bg +"(" + tick + ","+glucose_status.avgdelta.toFixed(1)+")"+ " -> " + eventualBG + "-" + snoozeBG + " (Unadjusted: " + naive_eventualBG + "-" + naive_snoozeBG + ")");
+        
+        if (typeof eventualBG === 'undefined') { 
+            console.error('Error: could not calculate eventualBG');
```
cjo20: Duplicate printing of error because of rT being printed after return?

scottleibrand:
@LorelaiL I like the avgdelta tofixed and adding the maxsafebasal adjustment to reason. Thanks. Unfortunately the log/error changes would break everything. Use of oref0-determine-basal requires that the output be on stdout to go into the suggestedtemp.json. Putting anything else in there (what currently goes to error) or otherwise changing it will cause it to no longer be valid json, and break openaps. The else stuff I also removed because if you return early, that terminates the program, so by removing else clauses I was able to simplify my if/else hierarchy a lot.

scottleibrand:
Now that you have #36 opened to track the whole refactor, we should also make a new commit with just the substantive changes, such as https://github.com/openaps/oref0/pull/35/files#diff-40d4a549ff39a9078d8421ac054a8fdfR150, https://github.com/openaps/oref0/pull/35/files#diff-40d4a549ff39a9078d8421ac054a8fdfR196, https://github.com/openaps/oref0/pull/35/files#diff-40d4a549ff39a9078d8421ac054a8fdfR213, https://github.com/openaps/oref0/pull/35/files#diff-40d4a549ff39a9078d8421ac054a8fdfR261, https://github.com/openaps/oref0/pull/35/files#diff-40d4a549ff39a9078d8421ac054a8fdfR316, and any others I missed.  Those will be much easier to merge with minimal testing.

PR #38, opened 2015-12-05
Merge @LorelaiL's extract_module refactor and my fixes to dev

scottleibrand:


scottleibrand:
Tests are passing and the output now looks the same as from dev (on my current set of inputs).  Will need someone to check that this still works correctly in an actual oref0 loop implementation before we can merge it to dev.  Should also get someone to do a code review and make sure they don't see anything there that we shouldn't merge.

bin/oref0-determine-basal.js
```diff
@@ -56,7 +56,17 @@ if (!module.parent) {
...
+    var setTempBasal = require('../lib/basal-set-temp'); 
+    
+    rT = determinebasal.determine_basal(glucose_status, currenttemp, iob_data, profile, undefined, setTempBasal);
+
+    if(typeof rT.error === 'undefined') {
+        console.log(JSON.stringify(rT));
+    } else {
+        console.error(rT.error);
+    }
+    rT=rT;
```
jasoncalabrese: `rT=rT;` ?

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "description": "openaps oref0 reference implementation of the reference design",
   "scripts": {
-    "test": "make test"
+    "test": "mocha tests/*.js"
```
jasoncalabrese: moving away from using make for running the tests?

bin/oref0-determine-basal.js
```diff
@@ -67,276 +77,9 @@ function init() {
         , label: "OpenAPS Determine Basal"
         , pluginType: 'pill-major'
```
jasoncalabrese: looks like this is trying to be a NS plugin...  probably don't need this now

package.json
scottleibrand: @LorelaiL what was your reason for changing from make to mocha?  Was make (the way we had it configured) just not running all of the tests?

bin/oref0-determine-basal.js
scottleibrand: Fixed.

bin/oref0-determine-basal.js
scottleibrand: Removed.

scottleibrand:
@jasoncalabrese @amazaheri @LorelaiL anything else, or are we good to merge to dev?

PR #39, opened 2015-12-07
Expected delta

scottleibrand:
This introduces the concept of "expected delta" as a replacement for BGI/2, so that rather than a binary threshold, we gradually phase in high temps when the PWD is not falling fast enough, or low temps when not rising fast enough.  Fast enough is defined as BGI, adjusted for whatever additional delta (rate of rise/fall) would be required to get to target over the course of DIA.  So if you're 300 with a 120 target, 40 ISF, 3h DIA, and no IOB / insulin activity, you'd need to be falling at 5mg/dL per 5m before it completely phases out the high temp.  This should help with situations where IOB is way off from what it "should" be, especially when riding high without any correction bolus IOB.

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -60,7 +60,8 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     } else { 
         tick = glucose_status.delta; 
     }
-    
+    var minDelta = Math.min(glucose_status.delta, glucose_status.avgdelta);
+    //var maxDelta = Math.max(glucose_status.delta, glucose_status.avgdelta);
 
     
     //calculate BG impact: the amount BG "should" be rising or falling based on insulin activity alone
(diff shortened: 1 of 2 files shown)


scottleibrand:
Until #38 is merged, https://github.com/openaps/oref0/compare/wip/refactor...expected-delta?expand=1 is a much cleaner view of what's changing (diff against the refactored version).

scottleibrand:
Now that #38 is merged I think we can merge this to dev unless @jasoncalabrese or @amazaheri have seen any issues in their testing so far.  Will still have plenty more time for testing on dev, of course, before releasing.

jasoncalabrese:
It's looking good to me, going to let it run tonight

jasoncalabrese:
Ran this the last 2 nights, and I'm pretty sure it helped last night.  This it's ready to merge to dev.

scottleibrand:
Should we do a release of what's already in dev (
https://github.com/openaps/oref0/pull/41) first, or merge in expected-delta
first?

jasoncalabrese:
Not sure, the ns-status update tool is on dev, and would be nice to have more people try it, and would make it easier for people just starting.  But I don't get the impression that many loopers are looking to jump on new a release, since several are still running the old openaps-js.

I guess we've tested the refactor enough by running `expected-delta`, so it should be safe.

amazaheri:
I agree, lets merge it.

scottleibrand:
I'm not reading a clear preference one way or the other from either of
you.  Are you leaning toward merging expected-delta to dev first before a
release, then?

amazaheri:
I would merge into Dev.

jasoncalabrese:
Would have been nice to get the ns-status tool out before merging the refactor, but I have some doubts about releasing with the refactor since I skipped over it and went directly to `expected-delta`

scottleibrand:
Happy to cherry-pick ns-status out and do a patch release of that if you
want.  But if we don't care about the refactor without expected-delta, I'll
go ahead and merge expected-delta now.

PR #44, opened 2015-12-10
Profile undefined check cleanup

jmatheson:
Fixed typo in checking for defined value


jmatheson:
Looks like i did something wrong with this pull request

scottleibrand:
Try targeting dev instead of master.

PR #45, opened 2015-12-10
Fix typo for profile.min_bg

jmatheson:
Check for both min and max exist. Typo was just checking max twice.


scottleibrand:
Weird, thought we already fixed that.

PR #46, opened 2015-12-12
touch monitor/glucose.json before trying to cp it

jasoncalabrese:
If the `monitor/glucose.json` file happens to be deleted the `cp` would fail until `openaps get-bg` is run manually, by adding the touch a file will be created if it doesn't already exist.

### DIFF
--- a/bin/ns-uploader-setup.sh
+++ b/bin/ns-uploader-setup.sh
@@ -102,7 +102,7 @@ openaps alias add monitor-pump "report invoke monitor/clock.json monitor/temp_ba
 openaps alias add get-settings "report invoke settings/model.json settings/bg_targets.json settings/insulin_sensitivities.json settings/basal_profile.json settings/settings.json" || die "Can't add get-settings"
 openaps alias add get-bg '! bash -c "openaps monitor-cgm 2>/dev/null || ( openaps get-ns-glucose && grep -q glucose monitor/ns-glucose.json && mv monitor/ns-glucose.json monitor/glucose.json )"' || die "Can't add get-bg"
 openaps alias add gather '! bash -c "rm monitor/*; ( openaps get-bg && openaps get-settings >/dev/null && openaps monitor-pump ) 2>/dev/null"' || die "Can't add gather"
-openaps alias add wait-for-bg '! bash -c "cp monitor/glucose.json monitor/last-glucose.json; while(diff -q monitor/last-glucose.json monitor/glucose.json); do echo -n .; sleep 10; openaps get-bg >/dev/null; done"'
+openaps alias add wait-for-bg '! bash -c "touch monitor/glucose.json; cp monitor/glucose.json monitor/last-glucose.json; while(diff -q monitor/last-glucose.json monitor/glucose.json); do echo -n .; sleep 10; openaps get-bg >/dev/null; done"'
 
 # upload treatments to nightscout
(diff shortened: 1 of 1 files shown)


scottleibrand:
Wouldn't it be better for the cp to fail, so the diff fails and it continues straight on without waiting?

jasoncalabrese:
That's not what was happening, cp failed and it didn't even try the diff

jasoncalabrese:
this is what I was seeing:

```
Dec 12 11:53:55 indy2 openaps-loop: cp: cannot stat ‘monitor/glucose.json’: No such file or directory
Dec 12 11:53:55 indy2 openaps-loop: diff: monitor/glucose.json: No such file or directory
Dec 12 11:53:55 indy2 openaps-loop: removed ‘predict/oref0.json’
Dec 12 11:53:59 indy2 openaps-loop:
Dec 12 11:53:59 indy2 openaps-loop: module.js:340
Dec 12 11:53:59 indy2 openaps-loop:     throw err;
Dec 12 11:53:59 indy2 openaps-loop:           ^
Dec 12 11:53:59 indy2 openaps-loop: Error: Cannot find module '/home/pi/indy/monitor/glucose.json'
Dec 12 11:53:59 indy2 openaps-loop:     at Function.Module._resolveFilename (module.js:338:15)
Dec 12 11:53:59 indy2 openaps-loop:     at Function.Module._load (module.js:280:25)
Dec 12 11:53:59 indy2 openaps-loop:     at Module.require (module.js:364:17)
Dec 12 11:53:59 indy2 openaps-loop:     at require (module.js:380:17)
Dec 12 11:53:59 indy2 openaps-loop:     at Object.<anonymous> (/usr/local/lib/node_modules/oref0/bin/oref0-determine-basal.js:33:24)
Dec 12 11:53:59 indy2 openaps-loop:     at Module._compile (module.js:456:26)
Dec 12 11:53:59 indy2 openaps-loop:     at Object.Module._extensions..js (module.js:474:10)
Dec 12 11:53:59 indy2 openaps-loop:     at Module.load (module.js:356:32)
Dec 12 11:53:59 indy2 openaps-loop:     at Function.Module._load (module.js:312:12)
Dec 12 11:53:59 indy2 openaps-loop:     at Function.Module.runMain (module.js:497:10)
Dec 12 11:53:59 indy2 openaps-loop: calculate-iob://text/shell/predict/iob.json
Dec 12 11:53:59 indy2 openaps-loop: reporting predict/iob.json
Dec 12 11:53:59 indy2 openaps-loop: determine-basal://text/shell/predict/oref0.json
Dec 12 11:53:59 indy2 openaps-loop: reporting predict/oref0.json
Dec 12 11:53:59 indy2 openaps-loop: No action required
```

scottleibrand:
It did try the diff, and the diff failed because the file was not present, so it proceeded through the loop. There may be a bug in the logic where wait-for-bg doesn't return false if the diff fails, so it tries to enact regardless, but either way wait-loop should fail and it should proceed to retry with just loop. 

In any event, the wait-for-bg alias here in the read-only uploader is not referenced anywhere AFAICT, so should probably just be removed.

jasoncalabrese:
Saw that it wasn't used yet, and guess there's an other way to fix it.  Might be good to leave it since it's a good trick, and we could use it to upload sgvs to NS

PR #47, opened 2015-12-17
test capturing @audiefile scenario and overcompensatory quick-fix

scottleibrand:
This creates a unit test for #48, the issue @audiefile and @jasoncalabrese identified where expected-delta wouldn't low-temp when eventualBG < min but delta > expected delta (BG is falling more slowly than expected delta, mostly due to BGI).  Previously, I believe we were comparing delta to BGI/2, but in the expected-delta branch I changed it to compare to expected_delta (basically BGI), and haven't yet applied the newinsulinReq adjustment logic that I did for reducing high-temps.

This basic fix will disable the expected delta comparison completely (and low-temp) any time you’re falling with eventualBG < min.  That fixes the new unit test and doesn’t break any others, but I think the quick-fix will be too aggressive at low-temping when carbs are causing you to fall a lot slower than BGI, until we apply the newinsulinReq adjustment logic.

However, I think this is at least safer than the code currently running in dev.  If someone could give it a look through and confirm that, it might be worth merging this to dev even before I get the newinsulinReq adjustment logic in place.

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -162,8 +162,9 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
             // calculate 30m low-temp required to get projected BG up to target
             // use snoozeBG instead of eventualBG to more gradually ramp in any counteraction of the user's boluses
             var insulinReq = Math.min(0, (snoozeBG - target_bg) / profile.sens);
-            if (minDelta > 0 && minDelta < expectedDelta) {
-                var newinsulinReq = Math.round(( insulinReq * (1 - (minDelta / expectedDelta)) ) * 100)/100;
+            if (minDelta < 0 && minDelta > expectedDelta) {
+                // if we're barely falling, newinsulinReq should be barely negative
+                var newinsulinReq = Math.round(( insulinReq * (minDelta / expectedDelta) ) * 100)/100;
                 //console.log("Increasing insulinReq from " + insulinReq + " to " + newinsulinReq);
                 insulinReq = newinsulinReq;
             }
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -139,8 +139,6 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
                 rT.reason = rT.reason += "; no temp to cancel";
                 return rT;
                 }
-            } else { // BG is falling slower than expected delta
-                // TODO: possibly reduce low-temp by setting newinsulinReq
             }
         }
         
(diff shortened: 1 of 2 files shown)


lib/determine-basal/determine-basal.js
```diff
@@ -125,19 +125,22 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
...
-            } else {
-                rT.reason += ", but Avg. Delta " + glucose_status.avgdelta.toFixed(2) + " > Exp. Delta " + expectedDelta;
-            }
-            if (currenttemp.duration > 0) { // if there is currently any temp basal running
-                rT.reason = rT.reason += "; cancel";
-                return setTempBasal(0, 0, profile, rT, offline); // cancel temp
-            } else {
-            rT.reason = rT.reason += "; no temp to cancel";
-            return rT;
+            if (minDelta > 0) { // if 5m or 15m avg BG is rising faster than expected delta
```
jasoncalabrese: comment isn't correct anymore, should change it or just remove it

jasoncalabrese:
Does seem like it will low-temp too aggressively, but would be a way to do a super bolus, and take some future basal as bolus and allow the loop to temp to 0.

In addition the scaling options that have been discussed and similar to what we do for high temps, an other option might to use a forecasted BG an hour out instead of looking at the full/max effect of the IOB.

lib/determine-basal/determine-basal.js
scottleibrand: I think it is. That's the scenario I originally tested for: BG is low and rising faster than expected delta, but eventualBG is still less than min.

jasoncalabrese: Think it's just the location of the comment the confused me, it makes sense at the previous location before `if (minDelta > expectedDelta) {`

scottleibrand: I was trying to split up "rising faster" vs. "falling slower". Suggestions/edits welcome.

scottleibrand:
I don't think we want to encourage super bolusing. It uses up all of your safety margin, so much better to just do a full correction bolus and then let OpenAPS manage the glide path once bolus snooze expires, or do a partial correction bolus and let oref0 expected-delta continue to high-temp to get the rest. With meal assist we may be able to shorten bolus snooze more, though.

scottleibrand:
And with that we are now properly scaling low-temps, so this should be good to merge into dev for further testing there.

jasoncalabrese:
Ran this last night, working good and solves some issues I was seeing. :+1:

amazaheri:
Ditto, lets ship it! :+1:

PR #51, opened 2016-01-09
set deltas in tests to 4 to enable wtf assist

jasoncalabrese:


scottleibrand:
Probably should disable the debug line too.

jasoncalabrese:
Did you fix the debug line, was thinking we should add another test with the delta of 1 to verify wtf-assist doesn't kick in. Unless there's already a test for that.

scottleibrand:
Not yet: haven't touched a PC yet today. Good idea.

PR #53, opened 2016-01-10
Corrected positional arguments for NS profile creation

ktomy:
It seems there were some typos, positional variables were overlapping each other. Also I have put a default output file name.


scottleibrand:
LGTM. @bewest / @jasoncalabrese, any reason not to merge this?

PR #60, opened 2016-01-25
add device field to NS status

jasoncalabrese:
using the format openaps://hostname


jasoncalabrese:
related to https://github.com/nightscout/cgm-remote-monitor/pull/1487

PR #63, opened 2016-02-02
safe require the NS stus inputs

jasoncalabrese:
so 1 missing file doesn't cause the entire status uload to fail


scottleibrand:
Oh, that is slick. Didn't know you could do it that easily. We should do that everywhere.

PR #64, opened 2016-02-05
Added shell scripts to format dexcom reading into a NS preferred format

ThomasEmge:
I have added two scripts:
1. formatting glucose and iter_calibrations readings to sgv and mbg
2. formatting sensor_insertions to treatments

I also modified nightscout.sh to allow for culling of the dexcom readings

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,31 @@ API_SECRET="..." NIGHTSCOUT_HOST=localhost:1337 ns-upload-entries <input> <outpu
 
 ```
 
+
+### `format-ns-glucose`
+
+Format OpenAPS dexcom readings (glucose,calibrations) to a format that Nightscout prefers.
+The `TYPE` parameter specifies if you would like formatting for glucose readings `sgv` or calibrations readings `mbg`. If the input file is result of `openaps use cgm iter_glucose` please specify `sgv`, if the input is the result of `openaps use cgm iter_calibrations` then use `mbg` as the `TYPE` parameter.
+The `MODEL` parameter is a string that can be used to identify the dexcom model used to retrieve the readings.
+
+```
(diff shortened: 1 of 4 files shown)


bewest:
Looking good.

scottleibrand:
@inform880 and @jasoncalabrese can you test this and see whether it at least partially addresses https://github.com/openaps/oref0/issues/88 ?

jasoncalabrese:
I didn't find a way to get the cal records, with the slope and intercept.  We need that to use the sensor records.  The calibrations are mbgs.

scottleibrand:
@jasoncalabrese does that just affect raw?  Do these scripts look good for getting SGVs at least?

jasoncalabrese:
Yeah, that's just for raw, only issue I see with this related to the device name.  I think we should stick with the openaps:// prefix, something like openaps://dexcom.

jasoncalabrese:
Think we want a different treatment eventType so NS will use it for the sage plugin, need to look that up, but on phone now.

jasoncalabrese:
We should use `Sensor Start` for the `eventType` https://github.com/nightscout/cgm-remote-monitor/blob/dev/lib/plugins/sensorage.js#L80

scottleibrand:
Not sure why this got stalled.  But is it still needed in light of `oref0_glucose`?

ThomasEmge:
I haven't been actively using it but the code itself is sitting in my loop in case I want to plug in the dexcom to the rpi. Where can I find `oref0_glucose`?

scottleibrand:
I don't think this is needed any longer in light of oref0_glucose and https://github.com/openaps/docs/wiki/CGM-Loop-with-openxshareble-scratchpad - closing.

PR #65, opened 2016-02-06
Automatically convert units to mg/dL

oskarpearson:


bin/oref0-get-profile.js
```diff
@@ -35,15 +35,34 @@ if (!module.parent) {
     var cwd = process.cwd()
     var pumpsettings_data = require(cwd + '/' + pumpsettings_input);
     var bgtargets_data = require(cwd + '/' + bgtargets_input);
+
+    if (bgtargets_data.units == 'mmol/L') {
+      /* Convert to mg/dL */
+      bgtargets_data.units = 'mg/dL'
+      for (a in bgtargets_data.targets) {
+         var targetarea = bgtargets_data.targets[a];
```
scottleibrand: This creates a new variable for the translated targets, but never changes or replaces references to bgtargets_data.

bin/oref0-get-profile.js
```diff
@@ -35,15 +35,34 @@ if (!module.parent) {
...
+         targetarea.high = Math.round(targetarea.high*18.0);
+         targetarea.low = Math.round(targetarea.low*18.0);
+      }
+    }
+
     if (bgtargets_data.units !== 'mg/dL') {
```
scottleibrand: This will never trigger, because you changed it above.

bin/oref0-get-profile.js
```diff
@@ -35,15 +35,34 @@ if (!module.parent) {
...
+       for (a in isf_data.sensitivities) {
+         isf_data.sensitivities[a].sensitivity = Math.round(isf_data.sensitivities[a].sensitivity*18.0);
+       }
+    }
     if (isf_data.units !== 'mg/dL') {
-      console.log('ISF is expected to be expressed in mg/dL.'
-                 , 'Found', isf_data.units, 'in', isf_input, '.');
+      console.log('Insulin sensitivity data was expected to be expressed in mg/dL.',
+                  'Found', isf_data.units, 'in', isf_input, '.');
       process.exit(2);
```
scottleibrand: This will still bail on incorrect units, as you never did a mg/dL conversion for isf_data like for targets.

scottleibrand:
Doesn't look like this will do what you intend. In addition to fixing the things I identified above, we would also need to add some unit tests to validate the new code is correctly converting everything and using the corrected units to generate a sensible recommendation.

oskarpearson:
This involves adding tests to something in bin/ - which we are going to do by refactoring things.

I'm going to close this PR in the meantime

PR #67, opened 2016-02-10
Include mmtune info in NS Status if available

jasoncalabrese:


bin/ns-status.js
```diff
@@ -41,8 +41,28 @@ function requireWithTimestamp (path) {
   return resolved;
 }
 
+function fixRecFlag (enacted) {
+  if (enacted) {
+    enacted.recieved = enacted.recieved || enacted.duration === enacted.requested.duration
```
jasoncalabrese: @scottleibrand, with this we don't need the hack in our ini's, removing that seems to have fixed the empty enacted.json filee

scottleibrand:
I haven’t had a chance to test this yet, but I like the idea, so if you've tested and it looks good, then :+1:

jasoncalabrese:
I've been running since last night tested without mmtune also

PR #69, opened 2016-02-11
a tool to support filling in raw glucose values

jasoncalabrese:
I want to be able to use raw bg in some times, but also want to be safe about it.  In this first pass theres a `maxRaw` that defaults to 150.  The tool will always add a `raw` field, but will only set `glucose` to `raw` when `glucose` is missing or `< 40` and `raw < maxRaw` to prevent a lot high temping from raw.

Might be good to add some limits around the delta to ignore big swings.

This should be helpful with the wixel too.

### DIFF
--- a/tests/with-raw-glucose.test.js
+++ b/tests/with-raw-glucose.test.js
@@ -26,7 +26,7 @@ describe('IOB', function ( ) {
     entry.raw.should.equal(115);
   });
 
-  it('should add raw glucose, but set missing glucose above maxRaw', function ( ) {
+  it('should add raw glucose, but not set missing glucose above maxRaw', function ( ) {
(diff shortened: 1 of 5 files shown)


tests/with-raw-glucose.test.js
```diff
@@ -0,0 +1,45 @@
...
+
+  it('should add raw glucose and set missing glucose', function ( ) {
+    var entry = {unfiltered: 113680, filtered: 111232, noise: 1};
+    withRawGlucose(entry, cal);
+
+    entry.glucose.should.equal(115);
+    entry.raw.should.equal(115);
+  });
+
+  it('should add raw glucose, but set missing glucose above maxRaw', function ( ) {
```
scottleibrand: Missing a not?  "it('should add raw glucose, but not set missing glucose above maxRaw'"?

PR #87, opened 2016-03-24
Wip/bewest/help output

bewest:
Bless all tools with `-h` flag.

Allows using something like this to generate docs:

```
$ cat mktool-ref.sh 
#!/bin/bash


OPENAPS_HOME=${1-~/Documents/openaps}
PREFIX=${2-openaps}
OUTPUT=${3-source/reference}
OUTDIR=$OUTPUT/$PREFIX
compgen -c $PREFIX | while read tool ; do
  if [[ -x $(which $tool) ]] ; then
  OUTFILE=$OUTDIR/$tool.md
  (
    mkdir -p $OUTDIR

    echo "# \`$tool\`"
    echo
    echo "## Help"
    (cd $OPENAPS_HOME
      $tool --help
    )

  ) | tee $OUTFILE
  fi
done


```


bewest:
Bringing branch up to date with `dev` for consideration.

This mainly tries to make sure that every single tool does something simple when `-h` or `--help` is given.

bewest:
See also:

* https://github.com/openaps/oref0/pull/94
* https://github.com/openaps/oref0/pull/93

bin/mm-format-ns-profile.sh
```diff
@@ -2,18 +2,32 @@
...
 # Author: Ben West
 
+self=$(basename $0)
 SETTINGS=${1-monitor/settings.json}
 CARBS=${2-monitor/carb-ratios.json}
 BASALRATES=${3-monitor/active-basal-profile.json}
 SENSITIVITIES=${4-monitor/insulin-sensitivities.json}
 TARGETS=${5-monitor/bg-targets.json}
-OUTPUT=${6-settings/ns-profile.json}
+OUTPUT=${6-/dev/fd/1}
```
scottleibrand: What is this change?

scottleibrand:
One question (line comment) on bin/mm-format-ns-profile.sh.  Otherwise :+1:

bin/mm-format-ns-profile.sh
bewest: Where to `cat` the results (`-` aka, stdout is also named `/dev/fd/1`)  Many tools allow for an `--output -` or any other file, this is the bashism to do that.

PR #92, opened 2016-03-28
Lower max override from 100 to 90

CrushingT1D:
Change from 100 to 90 as it is meant to be a minimum bound in case of user entry error. In this case a user (me) is trying to set it to 90 on purpose which is a reasonable value.

### DIFF
--- a/lib/profile/targets.js
+++ b/lib/profile/targets.js
@@ -25,7 +25,7 @@ function lookup (inputs) {
 
 function bound_target_range (target) {
     // hard-code lower bounds for min_bg and max_bg in case pump is set too low, or units are wrong
-    target.max_bg = Math.max(100, target.high);
+    target.max_bg = Math.max(90, target.high);
     target.min_bg = Math.max(90, target.low);
     // hard-code upper bound for min_bg in case pump is set too high
     target.min_bg = Math.min(200, target.min_bg);


scottleibrand:
I'm fine with this.  At the time we set these limits, we thought there was some value to having a "range" instead of a single target.  Since then we've been running with min_bg = max_bg, so I have no problem setting the max_bg = min_bg = 90 as a lower bound.

Does anyone see any reason not to merge this?

danamlewis:
I am also fine with this given what we've seen across the user base.

On Mon, Mar 28, 2016 at 2:11 PM, Scott Leibrand <notifications@github.com>
wrote:

> I'm fine with this. At the time we set these limits, we thought there was
> some value to having a "range" instead of a single target. Since then we've
> been running with min_bg = max_bg, so I have no problem setting the max_bg
> = min_bg = 90 as a lower bound.
>
> Does anyone see any reason not to merge this?
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/openaps/oref0/pull/92#issuecomment-202582296>
>



-- 
Dana Lewis | http://diyps.org | @DanaMLewis
<http://www.twitter.com/danamlewis> | http://www.linkedin.com/in/danalewis
"Doing something for someone else is more important than anything you would
do for yourself."

PR #93, opened 2016-03-30
Wip/bewest/autoconfigure ns

bewest:
Provide some tools to set up nightscout devices easily.

This actually provides a single `ns` device that can both download and upload any content to/from NS:

`openaps use ns shell get entries.json "count=10"`, vs `openaps use ns shell upload treatments.json recently/created-treatments.json` should both work as expected, as well as many variants.  Note, this relies on certain fixes to the `remainder` argument in `openaps` : https://github.com/openaps/openaps/pull/80

### DIFF
--- a/bin/nightscout.sh
+++ b/bin/nightscout.sh
@@ -15,18 +15,92 @@ $self <cmd>
 * latest-openaps-treatment
 * cull-latest-openaps-treatments
 
-* ns-get
-* ns-upload
-* ns-dedupe-treatments
-* ns-status
-* ns-upload-entries
+* get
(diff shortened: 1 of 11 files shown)


bewest:
Something like this:

```bash
$ nightscout autoconfigure-device-crud myhost.com foboar
process://ns/nightscout/ns myhost.com 143505794bdd283f7daed50f42d201b6926892c8

```

Now allows this as a result:

```
openaps use ns shell get entries.json "count=10"

openaps use ns shell upload treatments.json recently/created-treatments.json
```

PR #54, opened 2016-01-14
Extended last glucose reading date validation

ktomy:
As discussed on gitter, old  versions of ns uploader have different date format. I made a better date validation to be sure that in case of "invalid date" determine-basal will return an error.

PR #103, opened 2016-04-29
Fix whitespaces issue for pull request 54 Extended last glucose reading date validation

eyim:


scottleibrand:
Calling all Medtronic CGM users (and anyone else) for some testing. @eyim just created this PR to pull in @ktomy's openaps/oref0#54 changes. would like to have a few people check out @eyim's oref0 dev branch (https://github.com/eyim/oref0/tree/dev) and see if it works in all the different old-BG situations you can think of (and doesn't break anything in normal usage)

eyim:
Hmm still didn't fix the issue I am having with old bg info.  Hopefully it fixes errors for other folks

scottleibrand:
@eyim want to open an issue with details on what you're seeing?  we should be able to update these checks to catch it.

bewest:
Couple of notes:
Does this cover https://github.com/openaps/oref0/issues/104 or need to?

* `Date.parse(isoformat)` generates a timestamp in ms... many formatted date strings can be converted to js date using `new Date(Date.parse(isoformatish))`.
* using http://sugarjs.com/dates is wonderful.

I see that some of the fixes I came up with conflict, I've posted a resolution to the conflicts here: https://github.com/openaps/oref0/pull/103.  Would love to see some of the `sugarjs` date stuff come into play.  Should we enable a `--dateField` switch to allow oref0 to select which field to read the date from?

scottleibrand:
So where are we with this now that fix-cli has been merged to dev?

eyim:
Sorry I am not qualified to pass judgement on this.  I simply fixed the white space issue. I can only say that it didn't fix my particular issue.  

Edward

PR #111, opened 2016-05-13
Rename max_iob.json to preferences.json

mddub:
First step. I think this is the best way to start introducing more configuration into the `profile` (e.g. things like `disable_neutral_temps`).

### DIFF
--- a/bin/oref0-get-profile.js
+++ b/bin/oref0-get-profile.js
@@ -18,7 +18,7 @@
 
 var generate = require('oref0/lib/profile/');
 function usage ( ) {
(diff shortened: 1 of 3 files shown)


scottleibrand:
👍 
We'll also need to update https://github.com/openaps/oref0/blob/dev/bin/oref0-mint-max-iob.sh#L12 and the openaps/docs: https://github.com/openaps/docs/search?q=max_iob

mddub:
Yep, I'll fix the docs once this ships. Just fixed `oref0-mint-max-iob` and one other reference.

PR #112, opened 2016-05-14
Make neutral temp basals configurable

mddub:
Docs update to follow.

### DIFF
--- a/bin/oref0-get-profile.js
+++ b/bin/oref0-get-profile.js
@@ -97,7 +97,7 @@ if (!module.parent) {
     , basals: basalprofile_data
     , isf: isf_data
     , max_iob: preferences.max_iob || 0
-    , set_neutral_temps: preferences.set_neutral_temps || false
+    , skip_neutral_temps: preferences.skip_neutral_temps || false
     , carbratio: carbratio_data
     };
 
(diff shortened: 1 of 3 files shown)


scottleibrand:
A whole bunch of tests are failing because you changed the default behavior to `preferences.set_neutral_temps || false`.  I think the simplest and least-surprise method of implementing this configurability would be to set the default to true.  If we have some reason to think that's a bad idea, we could instead make the default true if meal-assist or auto-sens are enabled, but false otherwise.  In that case we'd probably need to update and add more tests.

Thoughts?

mddub:
I don't agree that it would be the least-surprise method, since anyone who updates to latest after starting on oref0 before ~February will be as surprised by new behavior as I was. I see this as restoring the default behavior.

That said, I do agree it would be simpler, and I have no interest in cleaning up those tests. I changed it to `skip_neutral_temps` instead.

scottleibrand:
Ok, let's see what input we get from other folks who've switched to dev awhile ago or recently.  If most people want skip_neutral_temps to be the default (at least when meal assist and auto-sens are disabled) I'm happy to set the default conditionally and update the tests accordingly.

PR #120, opened 2016-05-22
Generate a warning, not a fatal error, on insufficient glucose data for autosens

### LINKED ISSUE #117: determine-basal report not working if auto-sens.json report not available
I have encountered an issue where inadequate blood glucose data to produce an auto-sens.json file is stopping determine-basal reporting from producing a report, and therefore stopping looping. 

It would appear to be more logical to continue to have determine-basal produce a report if auto-sens.json is missing or has no data, rather than stopping the report from proceeding at all.

scottleibrand:
This should fix #117.

edison@edison-rl ~/openaps-dev $ openaps report invoke settings/autosens.json
detect-sensitivity://text/shell/settings/autosens.json
Optional feature autosens disabled: not enough glucose data to calculate sensitivity
reporting settings/autosens.json
edison@edison-rl ~/openaps-dev $ cat settings/autosens.json
{ "ratio": 1, "reason": "not enough glucose data to calculate autosens" }

### DIFF
--- a/bin/oref0-detect-sensitivity.js
+++ b/bin/oref0-detect-sensitivity.js
@@ -37,8 +37,8 @@ if (!module.parent) {
         var cwd = process.cwd();
         var glucose_data = require(cwd + '/' + glucose_input);
         if (glucose_data.length < 72) {
-            console.log('Error: not enough glucose data to calculate autosens.');
-            process.exit(2);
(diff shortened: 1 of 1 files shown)


johnmales:
Hi @scottleibrand , yes this fix is working correctly. Deliberately reducing the glucose data to 2 hours of data only resulted in this response:
`{"ratio":1,"reason":"not enough glucose data to calculate autosens"}` and determine-basal continued correctly. Thanks for the fix.

PR #123, opened 2016-05-26
Added detail to error message when time is out of sync between devices.

Spazholio:


eluhrs:
Excellent idea.  I just struggled with the same thing.  I suggest swapping "Dexcom" for "CGM" since some folks use the Enlite.

PR #121, opened 2016-05-22
Merge dev back into AMA via wip/merging-ama

scottleibrand:

PR #119, opened 2016-05-22
Merge AMA into dev via wip/merging-ama

bewest:
Rebased https://github.com/openaps/oref0/compare/dev...advanced-meal-assist?expand=1 to be able to merge with `dev`.  If people would like, we can do some training on how to maintain branches/avoid conflicts going forward.

Most conflicts were simple, just tweaking the help output.  A few were more complicated, it looked like @scottleibrand intended to remove code/comments in various places after refactoring, but it's hard to tell.  I went with what looked like was intended by the commit message itself, in theory this perfectly matches `advanced-meal-assist` branch, but can be merged into dev.

It includes all the nice `nightscout` tools and all the other updates to oref0.

Issue #114, opened 2016-05-14
Dev to AMA upgrade path 


scottleibrand:
Need to make it possible to switch from dev to AMA branch without it breaking existing loops.

amazaheri:
I am waiting for our second Edison to arrive and can help with validating the upgrade process.

bewest:
I rebased `advanced-meal-assist` onto dev over here: https://github.com/openaps/oref0/pull/119.
I tried using it, without the new features set up to see if it would issue the same results/behavior.  I had some trouble verifying if updating without "enabling" any new features would turn out the same behavior or if it should by design.  Looks like we might need to adjust how the features are turned on/off, perhaps by continuing to rely on `yargs`.

PR #125, opened 2016-05-29
Advanced meal assist

bewest:
Dev pathway for ama happens here.


bewest:
This seems to be main place for AMA feature to live, and was able to catch up to dev via https://github.com/openaps/oref0/pull/119, https://github.com/openaps/oref0/pull/121.  This should co-ordinate with https://github.com/openaps/oref0/issues/114

scottleibrand:
Checks are now passing, so this is a candidate for merge if no one flags any issues.  I believe @cjo20 said things looked good for him.  Anything else we need to do before merging to dev?

cjo20:
Yup, things looked good for me when I upgraded. I'm not sure if the reordering of the yargs stuff in oref0-get-profile has introduced any upgrade problems, but it should be minor

scottleibrand:
Yeah, I'm not too worried about anything that only affects people who've added a feature in the last week or two.  :-)

scottleibrand:
One last call for any objections to merging AMA to dev.

bin/oref0-determine-basal.js
```diff
@@ -105,7 +106,7 @@ if (!module.parent) {
             console.error(msg.msg);
             // console.log(JSON.stringify(msg));
             if (!params['missing-meal-ok']) {
-              errors.push(msg);
+              warnings.push(msg);
```
bewest: FWIW, `--no-missing-meal-ok` means "strict" mode which I would prefer throw an error (since it's being asked for).

scottleibrand: If you'd like a flag for strict mode that's fine, but the default needs to be to warn, not break the loop. This has caused us issues several times due to network comms issues (like wifi captive portals) when configuration is fine.

bin/oref0-get-profile.js
```diff
@@ -37,19 +56,15 @@ if (!module.parent) {
...
       usage( );
-      process.exit(0)
+      process.exit(0);
     }
     var bgtargets_input = params._.slice(1, 2).pop()
     var isf_input = params._.slice(2, 3).pop()
     var basalprofile_input = params._.slice(3, 4).pop()
     var preferences_input = params._.slice(4, 5).pop()
     var carbratio_input = params._.slice(5, 6).pop()
+    var temptargets_input = params._.slice(6, 7).pop()
```
bewest: maybe this should go as a switch `--temp-targets`?  In general, for future-compat, all positional args except last are mandatory.

bin/oref0-meal.js
```diff
@@ -21,7 +21,7 @@
 
 var generate = require('oref0/lib/meal');
 function usage ( ) {
-        console.log('usage: ', process.argv.slice(0, 2), '<pumphistory.json> <profile.json> <clock.json> [carbhistory.json]');
+        console.error('usage: ', process.argv.slice(0, 2), '<pumphistory.json> <profile.json> <clock.json> <glucose.json> <basalprofile.json> [carbhistory.json]');
```
bewest: For future compatibility, in situations like these like to do a ` --with-extra-foo` pattern so that additional options/features can be included without breaking previous optional arguments.

bewest:
Left a few notes regarding tips on maintaining positional arguments vs option switches over the long haul.  Overall I'm excited to get this into dev and subsequently do a release.

scottleibrand:
Sounds like we should merge to dev and then do some additional yargs conversion (in addition to the stuff that has recently been added) before releasing to master. If no objections, I'll do the merge to dev today.

jasoncalabrese:
I'd be ok requiring the `--auto-sens`, `--meal`, and other options, maybe the merge to dev is the time to do that, I updated my rigs to use the options and it's not a big deal.

bin/oref0-determine-basal.js
jasoncalabrese: for interactive/manual mode strict is nice, since it will help spot issues, but when running a loop it's a problem

I added `--missing-meal-ok` to my rigs to prevent issues, but I expect others to miss that and have their loop fail.

PR #131, opened 2016-06-14
Change basal rounding to match values the pump can deliver

cjo20:


jasoncalabrese:
I think rounding is different for x23's

cjo20:
Do you know what the minimum scroll rate is for an x23?

lib/determine-basal/determine-basal.js
```diff
@@ -16,14 +16,19 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     var rT = { //short for requestedTemp
     };
 
+    //40 gives basal rates that match a 0.025u scroll rate.
+    //80 would match 0.05, 10 would match 0.1u
+    //Was 1000 before this change (round to 2 dp)
```
scottleibrand: Was 100 before

jasoncalabrese:
For the bolus wizard it's configurable, I forget what the basal steps are,
but I know they are smaller for x23's.

scottleibrand:
The problem with waiting until the last minute to round things is you get lost of really ugly 17-digit numbers being reported in the UI.  I like the change in the PR, and am happy to merge it as soon as we verify the basal resolution of the x23s.  @loudnate may know offhand, or we can go find where he does the same thing in his code.

lib/determine-basal/determine-basal.js
cjo20: Was 100 or 1000, depending on where it was in the code

cjo20:
eww, it actually turns out to be nasty. I thought it was the same as the bolus scroll rate, it turns out it is dependent on how large the basal rate is:

Increments:
• 0.025 units for basal amounts in the range of 0.025 to 0.975 units
• 0.05 units for basal amounts in the range of 1 to 9.95 units
• 0.1 units for basal amounts of 10.0 units or larger

For both the x23 (https://www.medtronicdiabetes.com/sites/default/files/library/download-library/user-guides/Paradigm_Revel_user_guide.pdf) and x54 (https://www.medtronic-diabetes.co.uk/sites/uk/medtronic-diabetes.co.uk/files/veo-x54_ifu_updated_26.04.2013.pdf)

scottleibrand:
So looks like 1/40 (0.025) will work, right?  If we're >= 1U/hr the pump will further round it to 1/20 (which it's already doing today).  Any reason we can't just let it keep doing that?  When the basal rate is that high, a slight mismatch between requested and enacted is no big deal: we just observe what was actually delivered and incorporate that into IOB.

cjo20:
I think a scale factor of 40 is better than what is there at the moment. I could just write a function to do the correct rounding (for x23 and x54 at least)

lib/determine-basal/determine-basal.js
Pazoles: Should it be "20 would match 0.05"?

cjo20: Yes. This line of code is disappearing in my next commit though :)

PR #144, opened 2016-06-24
Various determine-basal changes

cjo20:
- Round to doses the pump can provide - accurate for x23 and x54. For doses < 1u it will produce values between ones x22 pumps can provide, but this shouldn't be any worse than the current behaviour.
- As a result of rounding basal doses better, it should be possible to replace most of the checks that ballpark within 0.1u/hr.
- Convert some of the output numbers to mmol/l. This is mostly where actual blood glucose values are displayed, not delta or BGI values (because of the loss of resolution when rounding)
- Add tests for basal rounding
- get-profile now includes the pumps preferred unit setting
- Alter the require structure to allow access to round_basal outside of the .js file for tests.

### DIFF
--- a/bin/oref0-determine-basal.js
+++ b/bin/oref0-determine-basal.js
@@ -183,7 +183,7 @@ function init() {
     };
     
     determinebasal.getLastGlucose = require('oref0/lib/glucose-get-last');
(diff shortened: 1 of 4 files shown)


scottleibrand:
LGTM.  If someone can confirm that they're looping successfully on this code using mg/dL, I think we're good to merge.

PR #150, opened 2016-07-04
Change > /dev/stderr to >&2 for portability and weird file truncation

cjo20:
I run openaps with all output being appended to a log file. I was getting repeated truncation of the logfile whenever nightscout.sh was writing to /dev/stderr. Changing to >&2 seems to fix the problem, and should be more portable.


scottleibrand:
LGTM. @bewest any gotchas to know about?

PR #151, opened 2016-07-04
Round basal

cjo20:
Previous attempts at rounding basal were thwarted by another set of rounding in SetTempBasal. Pull the round basal function out in to a separate file to make it easier to call in multiple places.

Add model.json to the profile produced by get-profile. This model is now used by round_basal to choose an appropriate precision for doses in the sub-1u region. 

Undo previous changes to how determine-basal is require()'d by other files - they are no longer needed with round_basal in another file.

### DIFF
--- a/bin/oref0-get-profile.js
+++ b/bin/oref0-get-profile.js
@@ -18,7 +18,7 @@
 
 var generate = require('oref0/lib/profile/');
 function usage ( ) {
-        console.log('usage: ', process.argv.slice(0, 2), '<pump_settings.json> <bg_targets.json> <insulin_sensitivities.json> <basal_profile.json> [<preferences.json>] [<carb_ratios.json>]');
+        console.log('usage: ', process.argv.slice(0, 2), '<pump_settings.json> <bg_targets.json> <insulin_sensitivities.json> <basal_profile.json> [<preferences.json>] [model.json] [<carb_ratios.json>]');
 }
 
 if (!module.parent) {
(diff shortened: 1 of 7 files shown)


bin/oref0-get-profile.js
```diff
@@ -18,7 +18,7 @@
 
 var generate = require('oref0/lib/profile/');
 function usage ( ) {
-        console.log('usage: ', process.argv.slice(0, 2), '<pump_settings.json> <bg_targets.json> <insulin_sensitivities.json> <basal_profile.json> [<preferences.json>] [<carb_ratios.json>]');
+        console.log('usage: ', process.argv.slice(0, 2), '<pump_settings.json> <bg_targets.json> <insulin_sensitivities.json> <basal_profile.json> [<preferences.json>] [model.json] [<carb_ratios.json>]');
```
scottleibrand: This will break everyone passing carb_ratios until they modify their loop configs. Need to instead add it as the last argument, or use yargs to make it a non-positional one.

cjo20:
Yeah, I was on the fence about that. I chose the way round I did because it seemed more likely that people starting with the loop would be more likely to want to have their basal rates set correctly before they move on to the carb_ratios stuff. 

I'll change it round.

scottleibrand:
Any idea what the merge conflicts are?

cjo20:
No. As far as I can tell, I was at the head.

"This branch has conflicts" without saying what they aren't isn't the most helpful thing github could do *glares at github* Anything I can run on my end that might give me some insight?

scottleibrand:
You can try a git checkout dev, git pull, git merge dev. If that doesn't work, click on the command line instructions for merging and they should become apparent. You can then push back to your branch once you've resolved them, rather than trying to push to origin dev.

cjo20:
Conflict was with the change to the floating point precision when printing.

cjo20:
Current state seems to work with / without --model model.json in the "remainder" argument in openaps.ini

If I put a garbage file where it expects carb_ratios.json, I get the error messages I would expect. Not currently got carb_ratios set up, so I'll test more fully when I've got a file that parses correctly.

cjo20:
Tested with a proper carb_ratios.json and it seems to work.

scottleibrand:
LGTM. Anything else before we merge?

cjo20:
Think that's everything I was going to do for this pull request. Ready for merging as far as I can see.

PR #153, opened 2016-07-06
Better check for whether model is present

cjo20:
Saw a report of determine-basal crashing because `typeof profile.model` was `Object` - profile.json contained `'model':{}`. Moved to a stricter check.


scottleibrand:
Can you confirm the unit tests still pass for you? Looks like the CI build failed for some other npm reason...

cjo20:
I triggered a Travis rebuild, seems to have got through this time.

PR #163, opened 2016-07-10
Bring back export-defaults and update-preferences commands

sulkaharo:
Using the new yags commands format


scottleibrand:
This breaks when --model argument is not provided.  Reverted until someone has a chance to fix.

Issue #164, opened 2016-07-11
Support tiny basals


scottleibrand:
Currently determine-basal will allow a temp to continue running if it's within 0.1U/hr of the desired temp, or if insulinReq is within 0.1U.  Rather than using a hard-coded value, which is too small for basals >1U/hr and possibly too large for very small basal rates, it would be better to let the current temp run if it is within a certain percentage (perhaps 20%) of the desired temp/quantity.

@cjo20 thoughts?

cjo20:
Sounds like something worth trying. Gut feeling says that 20% is a little high, but I'll think about that a bit more this evening.

scottleibrand:
We get this kind of thing all the time: https://www.dropbox.com/s/bc84762zti9yyj4/Screenshot%202016-07-11%2014.03.40.png?dl=0

I'd rather reduce the number of temp basal events NS has to deal with, by not changing the basal unless the current temp is far enough off that we aren't likely to want to reverse our decision in 5 minutes with the next BG data point.

cjo20:
I'm wondering whether this change doesn't just mask another issue. I can think of two cases that could lead to this behavior:
1. determine-basal being weighted too heavily towards the delta over the last 5 minutes; if the sensor readings wobble slightly and the last 5 minutes are treated as very significant then small sensor errors will translate in to toggling between max basal rate and 0. I think I remember you saying something about changing it to look at the past 30 minutes?
2. Underlying basal issues - not sure what happened around the time of the image you posted, but if the basal at that time was too low then it would produce the ratcheting effect seen in that image as oref0 tries to catch up.

The issue I have with waiting to change the dose is it effectively dampens the response time of the CGM - if you are genuinely in the middle of a long rise, you'll be giving, on average, 10% less insulin than is needed. And on the way down, you'll be giving, on average, 10% more than is needed. My concern with that is the lower limit on basal rate - once it gets to 0u/hr, we're stuck and have to rely on the user to fix anything that happens.

Ultimately, I have no objection to swapping to the %-based system, I think it's definitely better than the current system at lower basal rates, I just think there might be a better solution to the problem it's trying to solve. I'd suggest starting at 10% and seeing what impact that has first (but that might just be me being overly cautious).

scottleibrand:
Making this a bit more generic for @moomoobloo's to track other incompatibilities with their use case of a 0.025U/hr basal (which currently prevents us from ever doing low glucose suspend unless we first high temp up to 0.125U/hr).

[...]

PR #165, opened 2016-07-12
Use 10m instead of 5m lookback for deviation

scottleibrand:
Per discussion in https://github.com/openaps/oref0/issues/164, it looks like it'd be better to smooth out our reaction to CGM variation by making many of our decisions on the 10m delta instead of the 5m delta.


scottleibrand:
@cjo20 or anyone else, any thoughts on this approach, in addition to the other stuff in #164?

scottleibrand:
Have been running this all week without issue.  Merging if no objections.

PR #167, opened 2016-07-14
add autosens prefs for ratio limits and if targets should be adjusted

jasoncalabrese:


lib/profile/index.js
```diff
@@ -5,14 +5,17 @@ var isf = require('./isf');
 var carb_ratios = require('./carbs');
 
 function defaults ( ) {
-    var profile = {        
-          max_iob: 0 // if max_iob is not provided, never give more insulin than the pump would have
-        // , dia: pumpsettings_data.insulin_action_curve        
-        , type: "current"
```
scottleibrand: Not related to your PR, but does anyone know what this type: "current" does?

scottleibrand:
I like it.  What testing have you done so far?

lib/profile/index.js
jasoncalabrese: I was wondering what that was too

jasoncalabrese:
I'm running it in production now, and allowing autosens to adjust targets again since it's easy to disable if it becomes a problem.  Only part I haven't seen get triggered is the logging when autosens rate gets limited.

scottleibrand:
Cool.  Merging.

PR #168, opened 2016-07-15
Support display_time in getLastGlucose

cjo20:
This needs to be tested with non-MDT cgm. MDT CGM output has the date in a different format, this will choose `date` if it is present, otherwise it will look for `display_time`.

### DIFF
--- a/lib/glucose-get-last.js
+++ b/lib/glucose-get-last.js
@@ -27,8 +27,12 @@ var getLastGlucose = function (data) {
         then = data[0];
     }
     change = now.glucose - then.glucose;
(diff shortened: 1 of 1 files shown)


scottleibrand:
Code LGTM.  👍  for merge after we test it with all different CGM sources (USB-connected Dexcom, openxshareble, NS share bridge, and xDrip).  @jasoncalabrese can you test the xDrip one?

jasoncalabrese:
I'm only using openxshareble now, but this looks ok.  It would be better to have something that processes the MM cgm data first that would do any conversions, but we're doing this type of thing in other places.

PR #169, opened 2016-07-15
Print max_bg, not profile.max_bg

cjo20:
Outputting the wrong target value when eventualBG > target.


jasoncalabrese:
updating rig with this now

PR #170, opened 2016-07-17
Pass the "BG data out of date" message on to nightscout

cjo20:
NS wasn't giving a reason for loop not running and it turned out to be the "BG data out of date" case that was causing it. This change allows the error message to be passed to Nightscout.

### DIFF
--- a/bin/oref0-determine-basal.js
+++ b/bin/oref0-determine-basal.js
@@ -148,13 +148,6 @@ if (!module.parent) {
     } else { console.error("Could not determine last BG time"); }
     var minAgo = (systemTime - bgTime) / 60 / 1000;
 
-    if (minAgo > 10 || minAgo < -5) { // Dexcom data is too old, or way in the future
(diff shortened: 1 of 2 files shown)


scottleibrand:
I installed this branch on one of my rigs and confirmed it doesn't break anything, and that the reason field gets uploaded to the NS devicestatus.json.  But at least with multiple rigs it's just saying "Looping" for the one that is getting BG too old, not showing me anything in the OpenAPS pill.  Have you tested that part and confirmed whether it displays something useful there about BG too old when no enacts are occurring?

cjo20:
I have had the "BG data out of date" message passed on NS and displayed in the OpenAPS pill - it should be part of the "Suggested" upload in the status. i don't know if this is because I'm running this change on NS: https://github.com/nightscout/cgm-remote-monitor/pull/1856

jasoncalabrese:
Should we even upload status when determine-basal fails because of stale data like this?  I see it with my rigs and have considered it a bug in loop.

cjo20:
I'm not sure how else to tell why a loop is failing without console/ssh'ing to the rig.

cjo20:
image
How it looks on my nightscout

PR #174, opened 2016-07-21
Fix rounding error for small basal rates.

moomoobloo:
Fix rounding error that was preventing small changes to small basal rates. Use round_basal instead. #164


scottleibrand:
Added the +/- 20% stuff and merged to dev.

PR #178, opened 2016-07-26
predict IOB out to DIA plus 30m to account for any running temp basals

scottleibrand:


scottleibrand:
This is a pretty minor change, but it will result in AMA being slightly less likely to temp to zero with lots of COB, as it takes into account an extra half hour of carb absorption.  We've been testing it in production for several days and like the new behavior.  I think this moves in the direction everyone wants (not counteracting meal boluses), so I'll probably merge this in a couple days if no one objects.

cjo20:
Is the requirement to look out past the end of the DIA the result of assuming 3mg/dl/5min carb absorption when there isn't any BG data to inform deviation calculations?

scottleibrand:
It helps in that case, but it is also useful when observed carb impact is low enough that not all carbs absorb over DIA, or when a temp basal is set and you want to see the full impact of it, at DIA past the end of the temp.

PR #179, opened 2016-07-26
support multiple receivers, and use all recent data points for avgdeltas

scottleibrand:
calculate and use new short_avgdelta (using avg of deltas from all data points 5-15m ago) and long_avgdelta (using avg of deltas from all data points 20-40m ago) vs. "now" (avg of all data points within last 2.5m)

This should mean we can have multiple rigs uploading BGs from multiple receivers and it won't affect our deltas.  It should also mean that our deltas are less jumpy, as we'll be using the average of several deltas in each case rather than just picking one arbitrarily.

Before merge, need to actually test this in production to verify we get expected behavior with a single rig, and then add uploads from a second receiver and verify everything stays happy.


scottleibrand:
@jasoncalabrese could you take a first look at this one, and see if you can follow my approach and whether you think it should be good from a theoretical perspective?

PR #196, opened 2016-08-22
add new carbratio_adjustmentratio preference

jasoncalabrese:
I'm using this so we can set higher than normal carb ratios on the pump, and cause the bolus wizard to calculate a partial bolus, but then correct the carb ratio used for AMA so we can catch up if/when there's a rise.

Might need to also adjust the min up CI from the current 3mg/dL/5m, but planning on experimenting without that first.

### DIFF
--- a/lib/determine-basal/cob-autosens.js
+++ b/lib/determine-basal/cob-autosens.js
@@ -98,8 +98,8 @@ function detectSensitivityandCarbAbsorption(inputs) {
         // if bgTime is more recent than mealTime
         if(bgTime > mealTime) {
             // figure out how many carbs that represents
-            // but always assume at least 3mg/dL/5m absorption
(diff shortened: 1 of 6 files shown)


jasoncalabrese:
Started testing using `"carbratio_adjustmentratio": .9` and raising the carb ratio about 10% for all times

jasoncalabrese:
Make that `"carbratio_adjustmentratio": 0.9`, without the 0 is invalid...

jasoncalabrese:
Also added new `profile.min_5m_carbimpact` pref, going to test with that at `4` tomorrow instead of the default 3 combined with the `0.9` for `profile.carbratio_adjustmentratio`.

The lower boluses helped some today, but I'd like low temps to end a little sooner and high temp to start sooner.

Issue #95, opened 2016-04-03
Erroneous negative IOB calculated when basal schedule rises >2x


scottleibrand:
@iainct reports at https://github.com/openaps/openaps/issues/88#issuecomment-204859700 that oref0 erroneously calculated a negative IOB based on a basal_profile.json that increases from 0.2 U/hr to 1.4 U/hr over the course of 60 minutes:

```
{ "i": 17, "start": "17:00:00", "rate": 0.2, "minutes": 1020 },
{ "i": 18, "start": "18:00:00", "rate": 0.8, "minutes": 1080 },
{ "i": 19, "start": "19:00:00", "rate": 1.4, "minutes": 1140 }
```

As oref0 currently calculates IOB based on the assumption that the currently scheduled basal profile is "close enough" to use for the last DIA hours' worth of pumphistory, this kind of basal profile is dangerous to use with oref0.

For now, we should detect when ( max_daily_basal / min_daily_basal ) > 3, and error out in that case.  In the long term, we should probably make oref0-calculate-iob smarter about using the basal scheduled at the time of each entry in pumphistory to calculate IOB as well.  I suspect we might want to keep the max/min > 3 check in place even after doing that, though.

scottleibrand:
Coded up the 3x limit.  Can I get some feedback on whether to merge #96 to dev as a fix for this?

loudnate:
You may want to consider using mmhistorytools to normalize temp basals (and suspends) against a basal schedule. It makes no estimations or assumptions about rates staying similar.

cjo20:
This just hit the top of my "stuff to look at" list. Using this issue to track discussion about the best way of integrating mmhistorytools.

thebookins:
Any reason why we can't use the iob report from the last loop run as a starting point for the iob calculation? Then we only need to assume that the current basal profile is "close enough" for the time since last run, not over the whole DIA.  Would also degrade nicely to current behaviour in the event of a break in looping. Sounds like mmhistorytools is the best solution going forward, but something like this might work in the meantime?

It means we're storing some state in the openaps implementation, which might be dangerous? Is it a deliberate design to not do this?

scottleibrand:
[...]

PR #198, opened 2016-08-25
Wip/accoutforbasalprofile

sulkaharo:
Changes IOB calculation take full basal profile into account. Fixes potential issues with fetching the basal rate (previous code was relying on the basal profile data being sorted, when it wasn't actually guaranteed).

### DIFF
--- a/bin/oref0-calculate-iob.js
+++ b/bin/oref0-calculate-iob.js
@@ -20,7 +20,7 @@
 
 var generate = require('oref0/lib/iob');
 function usage ( ) {
-    console.log('usage: ', process.argv.slice(0, 2), '<pumphistory.json> <profile.json> <clock.json> <basal-profile.json>');
+    console.log('usage: ', process.argv.slice(0, 2), '<pumphistory.json> <profile.json> <clock.json>');
 
 }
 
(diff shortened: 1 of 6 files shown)


thebookins:
While we are at it, it wouldn't hurt to fix the sorting function at the end of history.js:
`     all_data.sort(function (a, b) { return a.date > b.date });`
Since `a.date` and `b.date` are integers, it's sufficient to write `return a.date - b.date` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort). The current code doesn't sort properly. I'm not sure any of the downstream code depends on sorted data, but it may in the future.

jasoncalabrese:
Tested last night and installed on all our rigs

jasoncalabrese:
For sorting we should use lodash more, and get rid of the custom/fragile sort code.

thebookins:
The iob calculation will be incorrect when temp basals cross basal profile boundaries (see https://github.com/openaps/oref0/issues/95#issuecomment-242865676).

jasoncalabrese:
Updated main rig to the latest with the lodash sort, looking going I'm going to merge

PR #200, opened 2016-08-28
Fix ama oscillation

scottleibrand:
Opening PR for tracking and to see unit test info.  Have been running it offline for a few days, and am now pushing it to a couple rigs for testing.  Fixed the one existing unit test affected, and added one more, but not sure how to properly test AMA stuff without bringing in full real-world scenarios.  The oscillation is fairly rare, so planning to just wait for the situation to occur and make sure it does better than before.


scottleibrand:
Have been running this in prod all through Europe.  The determine-basal stuff looks good for the corner cases it covers.  There may be more to do to fix more cases, but I think this is good to merge.  The clockset fixes have also been quite useful, so I'd like to get them into dev ASAP.  @sulkaharo or @jasoncalabrese can you give this a once-over and merge if it looks good to you?

PR #203, opened 2016-09-05
Use correct ISF for any time of day when calculating autosens

sulkaharo:
Not dealt with properly: meal impact overlapping ISF change


scottleibrand:
This looks good running on our rigs.  Merging to dev.

PR #205, opened 2016-09-11
Use correct ISF for BGI; new oref0-crun tool

scottleibrand:
@sulkaharo any reason not to go ahead and merge this?  It's looking good to me...


sulkaharo:
Should be good to go!

PR #211, opened 2016-09-16
Validate carb ratios are reasonable

### LINKED ISSUE #210: Validate carb ratios are reasonable
x23 pumps allow 1.0 to 200 g/U
x15 pumps allow 3 to 150 g/U

In order to catch when openaps/decocare return impossible ratios per https://github.com/openaps/openaps/issues/111, and avoid AMA being way too aggressive as a result, we should probably limit the carb ratios oref0 accepts to the range used by the x15 (3 to 150).

scottleibrand:
Fix #210. This should keep AMA from being too aggressive due to https://github.com/openaps/openaps/issues/111
Cc @ryanchen26

### DIFF
--- a/bin/oref0-meal.js
+++ b/bin/oref0-meal.js
@@ -53,6 +53,12 @@ if (!module.parent) {
         return console.error("Could not parse input data: ", e);
     }
 
+    // disallow impossibly high/low carbRatios due to bad decoding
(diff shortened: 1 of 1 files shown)


scottleibrand:
Confirmed that this behaves as expected with a ratio of 0.26.  The resulting meal.json is:
```
{ "carbs": 0, "mealCOB": 0, "reason": "carb_ratio 0.26 out of bounds" }
```

PR #212, opened 2016-09-16
disable autosens and AMA by default in oref0-setup.sh

scottleibrand:


scottleibrand:
Going to merge this today if no objections.  In a future version I'll probably move to enabling autosens and (advanced) meal assist in preferences.json, but that requires changes to the main code (and perhaps updates to existing loops) so I think this is a better first step.

PR #214, opened 2016-09-18
oref0-setup: always set pump serial number; don't use the one from the export

scottleibrand:


PieterGit:
can confirm that it works with carelink stick now (specified in CLI with --serial=XYZ).
`openaps use pump model` works fine and shows pump model.

PieterGit:
i can also confirm that it works with a TI stick. ready for merging?

Issue #215, opened 2016-09-18
new oref0-setup branch does not work with mmol/L units (openaps get-settings fails silently and openaps monitor-pumps is broken)


PieterGit:
The new great oref0-setup scripts don't play nice with mmol/L units yet.

```
$ openaps get-settings
pump://JSON/model/settings/model.json
reporting settings/model.json
pump://JSON/read_bg_targets/settings/bg_targets.json
reporting settings/bg_targets.json
pump://JSON/read_insulin_sensitivities/settings/insulin_sensitivities.json
reporting settings/insulin_sensitivities.json
pump://JSON/read_selected_basal_profile/settings/basal_profile.json
reporting settings/basal_profile.json
pump://JSON/read_settings/settings/settings.json
reporting settings/settings.json
pump://JSON/read_carb_ratios/settings/carb_ratios.json
reporting settings/carb_ratios.json
get-profile://text/shell/settings/profile.json
reporting settings/profile.json
```

Looks quite good, but unfortuately:

```
$ cat settings/profile.json
BG Target data is expected to be expressed in mg/dL. Found mmol/L in settings/bg_targets.json .
```

The settings/bg_targets.json is in indeed in mmol/L units

```
$ cat settings/bg_targets.json
{
  "units": "mmol/L",
...
 ```

Should the oref0-script add the necessary stuff like
```
openaps device add units units
```

and convert the bg_targets_raw to bg_targets.json?

**Comment by PieterGit** [comment]
This causes monitor-pump to fail

``` 
$ openaps monitor-pump
pump://JSON/read_clock/monitor/clock.json
reporting monitor/clock.json
pump://JSON/read_temp_basal/monitor/temp_basal.json
reporting monitor/temp_basal.json
pump://JSON/iter_pump_hours/monitor/pumphistory.json
reporting monitor/pumphistory.json
tz://JSON/rezone/monitor/pumphistory-zoned.json
reporting monitor/pumphistory-zoned.json
tz://JSON/clock/monitor/clock-zoned.json
reporting monitor/clock-zoned.json
iob://text/shell/monitor/iob.json

module.js:485
    throw err;
          ^
SyntaxError: /home/pi/pi3n/settings/profile.json: Unexpected token B
    at Object.parse (native)
    at Object.Module._extensions..json (module.js:482:27)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
[...]

PR #216, opened 2016-09-18
Support both mmol/L and mg/dL

scottleibrand:
Fixes https://github.com/openaps/oref0/issues/215 (thanks @pietergit !)

I've tested that this works backwards compatibly with mg/dL.

@sulkaharo (or anyone else on mmol/L) you want to test this and confirm it works for you?

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -75,10 +75,8 @@ case $i in
 esac
 done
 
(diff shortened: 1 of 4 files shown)


PieterGit:
seems to work fine with current checkout.
I'm running in another issue https://github.com/openaps/oref0/issues/218 but that seems not related to mmol/L or mg/dL conversion.

PR #220, opened 2016-09-21
Fixed wrong test for $NIGHTSCOUT_HOST echo

SebastienLussier:
When running the setup script, entering a valid $NIGHTSCOUT_HOST resulted in
  "Ok, no Nightscout for you"
being printed if no TTY port was specified for mmeowlink


scottleibrand:
Thanks.  Looks like I fixed that in one place but not the other.

PR #222, opened 2016-09-23
Update oref0-setup.sh

Bender1061:
Add CGM type for Interactive walk through setup


bin/oref0-setup.sh
```diff
@@ -101,6 +101,9 @@ if [[ -z "$DIR" || -z "$serial" ]]; then
     read -p "What is your pump serial number? " -r
     serial=$REPLY
     echo "Ok, $serial it is."
+    read -p "What Kind of CGM are you using? (i.e. G4, G5, mdt) " -r
```
scottleibrand: I would capitalize MDT for consistency here.

Bender1061: very true

PR #223, opened 2016-09-23
Update device.json

ceben80:
Due to azure 24 h quota 288 BG query should be sufficient otherwise azure limit will exceed while querying 1000 BGs per minute

### DIFF
--- a/lib/oref0-setup/device.json
+++ b/lib/oref0-setup/device.json
@@ -17,7 +17,7 @@
     "name": "ns-glucose",
     "extra": {
       "fields": "",
-      "cmd": "bash -c \"curl -m 30 -s $NIGHTSCOUT_HOST/api/v1/entries/sgv.json?count=1000 | json -e \\\"this.glucose = this.sgv\\\"\"",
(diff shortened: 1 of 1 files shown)


scottleibrand:
Looking at this more closely, I don't think this device is used any longer, so that can't be what caused you to exceed your azure limits.  Per https://gitter.im/nightscout/intend-to-bolus?at=57eef1d2be5dec75500d5b50 we need someone to start working on switching the ns-loop upload stuff over to use the new ns device, and while we're at it we should start removing devices like this that are no longer used.

ceben80:
before oref0.setup.sh I was working with that's a good approach?

[device "ns"]
fields = oper remote input
cmd = nightscout
args = ns https://xxx.azurewebsites.net xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
								
[report "monitor/glucose.json"]
oper = get
use = shell
remote = entries/sgv.json
reporter = JSON
json_default = True
device = ns
input = count=288
remainder =

scottleibrand:
oref0-setup does this:

https://github.com/openaps/oref0/blob/dev/lib/oref0-setup/report.json#L32-L43

scottleibrand:
That way, if you have multiple receivers, it will download more than 288 data points, and if you have one receiver with gaps, 24h will be less than 288.

ceben80:
oki, understood

PR #224, opened 2016-09-23
Dev

tim2000s:
Added mods to the determine_basal.js and glucose_get_last.js to account for medtronic CGM.


scottleibrand:
Looks like you did a Merge branch 'dev' of https://github.com/tim2000s/openaps into dev.  That merges the openaps python package into the oref0 javascript package, which won't work.  You'll probably want to save a copy of your edits outside of git and then do a git reset (or delete and re-fork) to get back to the previous state and try again.

PR #226, opened 2016-09-24
Radio locale ww support

PieterGit:
add possibility to use oref0-setup.sh with TI-stick and World Wide (WW) pumps

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -60,6 +60,11 @@ case $i in
     API_SECRET="${i#*=}"
     shift # past argument=value
     ;;
+    -r=*|--radio-locale=*)
+    radio_locale="${i#*=}"
+    shift # past argument=value
+
+    ;;
     -e=*|--enable=*)
     ENABLE="${i#*=}"
     shift # past argument=value
@@ -70,6 +75,10 @@ case $i in
     ;;
 esac
 done
+if [[ -z $TTY && radio_locale != "US" ]]; then
+    echo "World wide radio locale is only supported with TI stick currently"
+    echo "If you'd like to help add World Wide radio locale support, please contact @pietergit on Gitter"
+fi
 
 if ! [[ ${CGM,,} =~ "g4" || ${CGM,,} =~ "g5" || ${CGM,,} =~ "mdt" ]]; then
     echo "Unsupported CGM.  Please select (Dexcom) G4 (default), G5, or MDT."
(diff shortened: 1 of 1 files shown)


bin/oref0-setup.sh
```diff
@@ -70,6 +75,10 @@ case $i in
     ;;
 esac
 done
+if [[ -z $TTY && radio_locale != "US" ]]; then
+    echo "World wide radio locale is only supported with TI stick currently"
```
scottleibrand: Aren't there WW Carelink sticks for WW pumps? If so, they should Just Work, and no further code changes should be required to support WW pumps with Carelink...

bin/oref0-setup.sh
```diff
@@ -70,6 +75,10 @@ case $i in
     ;;
 esac
 done
+if [[ -z $TTY && radio_locale != "US" ]]; then
```
scottleibrand: This will only work right if you define radio_locale to be "US" by default above.

bin/oref0-setup.sh
```diff
@@ -194,6 +214,10 @@ if openaps vendor add --path . mmeowlink.vendors.mmeowlink 2>&1 | grep "No modul
         echo -n "Cloning mmeowlink dev: "
         (cd ~/src && git clone -b dev git://github.com/oskarpearson/mmeowlink.git) || die "Couldn't clone mmeowlink dev"
     fi
+	if [[ "$radio_locale" -eq "WW" ]]; then
+		echo -n "Patching mmtune.py for WW frequencies. " # temporary fix for https://github.com/oskarpearson/mmeowlink/issues/16
+		sed -i "s/def __init__(self, link, pumpserial, radio_locale='US'):/def __init__(self, link, pumpserial, radio_locale='WW'):/g" ~/src/mmeowlink/mmeowlink/mmtune.py || die "Couldn't patch mmtune.py"
```
scottleibrand: Would be nice to make this part of pump.ini so we don't have to hard-code it. In the mean time, though, if you want to do this we need to add a git commit of the change so it doesn't block pulling future updates to that file.

bin/oref0-setup.sh
```diff
@@ -113,6 +122,17 @@ if [[ -z "$DIR" || -z "$serial" ]]; then
         echo -n TTY $ttyport
     fi
     echo " it is."
+	# 
+	if [[ -n "$ttyport" ]]; then
+		read -p "Are you using a USA (US) or world wide (WW) Medtronic pump? If US, press enter, otherwise enter WW. " -r
+		radio_locale=$REPLY
+		if [[ $radio_locale =~ ^[Ww][Ww]$ ]]; then
```
scottleibrand: You should use the `${radio_locale,,} =~ "ww"` syntax here to do a case insensitive comparison.

scottleibrand: Your approach will also work fine, actually. I misread it on first pass.

scottleibrand:
@PieterGit LMK if you think you'll be able to address the comments above, or if I should merge it and do so.

bin/oref0-setup.sh
oskarpearson: That's correct. The Carelink sticks come in USA and WW variants - and need to match the pump. You don't need to handle it specifically.

bin/oref0-setup.sh
oskarpearson: I'd really like to avoid doing this via patch if possible. The mmeowlink code can be changed to make the parameter available as a config field instead... it's just not something I've got to yet.

Could we pause this for the moment and I'll try have a look one of the evenings this week? My time is a bit limited at the moment as I'm moving house.

scottleibrand:
Since per @oskarpearson we don't want to merge this yet, I'm going to create a new branch for it in openaps/oref0 and merge it there so we can work on it collaboratively before merging to dev.  That will also let me delete the oref0-setup branch and re-branch it cleanly.

scottleibrand:
Opened https://github.com/openaps/oref0/pull/232 to track this.

PR #227, opened 2016-09-26
Match input reports to meal.json with their correct categories (basal, glucose, carbs). 

elnjensen:
This should fix the warning from running the `monitor/meal.json` report.


scottleibrand:
`make` on my own box runs fine, so gonna merge this and see if travis still has a problem with it.

PR #229, opened 2016-10-01
openxshareble support

scottleibrand:
Tested and working on an Edison Explorer rig.


scottleibrand:
This should affect anything if you don't specify the new options, so merging.

PR #230, opened 2016-10-01
Simplify Copy of Newer Files / Script Cleanup

rkresha:
can just use cp -pu instead of the rsync command. Using cp makes it more
portable since it's part of the core tools, whereas rsync may require an additional install in some
distros.

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -192,7 +192,7 @@ else
     (cd ~/src && git clone -b dev git://github.com/openaps/oref0.git) || die "Couldn't clone oref0 dev"
 fi
 echo Checking oref0 installation
-( grep -q oref0_glucose_since `which nightscout` && oref0-get-profile --exportDefaults 2>/dev/null >/dev/null ) || (echo Installing latest oref0 dev && cd $HOME/src/oref0/ && npm run global-install)
+( grep -q oref0_glucose_since $(which nightscout) && oref0-get-profile --exportDefaults 2>/dev/null >/dev/null ) || (echo Installing latest oref0 dev && cd $HOME/src/oref0/ && npm run global-install)
 
 echo Checking mmeowlink installation
 if openaps vendor add --path . mmeowlink.vendors.mmeowlink 2>&1 | grep "No module"; then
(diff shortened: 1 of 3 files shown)


scottleibrand:
LGTM, just need someone to run the modified oref0-setup install and confirm everything still works on raspbian and ubilinux.

rkresha:
Verified it works with raspbian

bin/oref0-setup.sh
```diff
@@ -192,7 +192,7 @@ else
     (cd ~/src && git clone -b dev git://github.com/openaps/oref0.git) || die "Couldn't clone oref0 dev"
 fi
 echo Checking oref0 installation
-( grep -q oref0_glucose_since `which nightscout` && oref0-get-profile --exportDefaults 2>/dev/null >/dev/null ) || (echo Installing latest oref0 dev && cd $HOME/src/oref0/ && npm run global-install)
+( grep -q oref0_glucose_since $(which nightscout) && oref0-get-profile --exportDefaults 2>/dev/null >/dev/null ) || (echo Installing latest oref0 dev && cd $HOME/src/oref0/ && npm run global-install)
```
scottleibrand: What's the benefit of that over ` syntax?

scottleibrand:
I tested out the `cp -up` stuff with https://github.com/openaps/oref0/pull/235 and it worked fine for me, so I'm ready to merge that if you are.

bin/oref0-setup.sh
rkresha: per http://wiki.bash-hackers.org/scripting/obsolete

> \`COMMANDS\`	replaced by $(COMMANDS)	This is the older Bourne-compatible form of the command substitution. Both the \`COMMANDS\` and $(COMMANDS) syntaxes are specified by POSIX, but the latter is _greatly_ preferred, though the former is unfortunately still very prevalent in scripts. New-style command substitutions are widely implemented by every modern shell (and then some). The only reason for using backticks is for compatibility with a real Bourne shell (like Heirloom). Backtick command substitutions require special escaping when nested, and examples found in the wild are improperly quoted more often than not.

scottleibrand: Ah, yes.  Quotes are one of the banes of my existence.  :-)

PR #234, opened 2016-10-03
split clockset.sh into oref0-set-device-clocks.sh and oref0-set-system-clock.sh

scottleibrand:


Bender1061:
does not work 

`edison@Bender1061:~/johnopenaps1$ ~/src/oref0/bin/oref0-set-system-clock.sh     Waiting for ntpd to synchronize... No!
ntpd did not synchronize.
Restarting ntp (via systemctl): ntp.service.
Waiting for ntpd to synchronize... No!
ntpd did not synchronize.
Fri Dec 31 19:25:05 EST 1999
edison@Bender1061:~/johnopenaps1$`

Bender1061:
I think the issue has to do with the sort, I tried last night from using the tail -1, to a head -1 and it worked correctly. I believe it's because it's seeing the two number of seconds (since beginning of time) as a string with the first number being the one that it will sort with. 
in C I would do a StrtoNum and then just pick the greater of the two numbers and use that. but I know you are more familiar with the bash scripting.

scottleibrand:
Interesting. What were the two epoch time strings, and how did it sort them lexically?

The bash equivalent of that numeric sort would be `sort -g` - might be worth a try.

Bender1061:
edison@Bender1061:~/johnopenaps3$ (date +%s; date -d `( cat $CLOCK; echo ) | sed 's/"//g'` +%s) | sort
1475550880
946684887
edison@Bender1061:~/johnopenaps3$

Bender1061:
edison@Bender1061:~/johnopenaps3$ (date +%s; date -d `( cat $CLOCK; echo ) | sed 's/"//g'` +%s) | sort -g
946685364
1475550880
edison@Bender1061:~/johnopenaps3$ (date +%s; date -d `( cat $CLOCK; echo ) | sed 's/"//g'` +%s) | sort
1475550880
946685372
edison@Bender1061:~/johnopenaps3$


Yeah it looks like the sort -g works

scottleibrand:
Ah, so y2k is just under 1 billion.  Thanks: I'll update the sort to use -g.

Bender1061:
so the command oref0-set-system-clock works perfectly now, but if the command is never called automatically when the enact alias is called.

scottleibrand:
I made a fix for that in the oref0-setup branch.  Just cherry-picked it over: https://github.com/openaps/oref0/pull/234/commits/5fa9d937bf729eb740ae9500377c58a94c56975b - you'll need to re-run oref0-setup to update your loop accordingly.

scottleibrand:
So this is good to merge now, right?

PR #235, opened 2016-10-03
add openxshareble support (with NS upload)

scottleibrand:
based on https://github.com/openaps/docs/wiki/CGM-Loop-with-openxshareble-scratchpad

still need to figure out if we're going to do wired G4 and G5 NS upload the same way, or something else.

### DIFF
--- a/bin/bt-pan
+++ b/bin/bt-pan
@@ -1,0 +1,222 @@
+#!/usr/bin/env python2
+from __future__ import absolute_import, print_function
+
+import os, sys, time, types, subprocess, signal
+
+import dbus
+
+
+### ~bluezutils.py
+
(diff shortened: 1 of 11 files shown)


scottleibrand:
This also incorporates https://github.com/openaps/oref0/pull/234

scottleibrand:
Probably should figure out how to disable looping based on raw data by default before we merge this.

scottleibrand:
We can fix merge conflicts on this and merge to dev after we've released dev to master and cut a new oref0 version, and then we'll need to disable looping on raw data by default before doing another release.

scottleibrand:
If/once we're happy with https://github.com/openaps/oref0/issues/269, I think this is good to merge to dev.

PR #246, opened 2016-10-14
path problem running script with sudo and calling cat

tghoward:
I don't know why the 'cat' command decides to define $HOME as "root" but I assume it has to do the the fact that I'm calling the script with sudo. (because of the explorer board). Don't feel like you have to bring in this PR, but this small change fixed it for me.  I did not replace $HOME in other places that this also might be a problem but my permutations of the setting caused the script to skip.


scottleibrand:
Hmm, I do everything as root on the Edison when using the Explorer board, as there is (was?) a mraa limitation that required that. Do you not have to do that?

tghoward:
Yep, you are right. Works as root. For some reason I thought it would be the same with sudo calls. Sorry about that.

PR #250, opened 2016-10-21
grep mins $(which oref0-dex-is-fresh)

Bender1061:
Fix for Checking oref0 holdup

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -200,7 +200,7 @@ else
     (cd ~/src && git clone -b dev git://github.com/openaps/oref0.git) || die "Couldn't clone oref0 dev"
 fi
 echo Checking oref0 installation
-( cat $(which oref0-dex-is-fresh) | grep mins && grep -q oref0_glucose_since `which nightscout` && oref0-get-profile --exportDefaults 2>/dev/null >/dev/null ) || (echo Installing latest oref0 dev && cd $HOME/src/oref0/ && npm run global-install)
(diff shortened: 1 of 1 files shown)


bin/oref0-setup.sh
```diff
@@ -200,7 +200,7 @@ else
     (cd ~/src && git clone -b dev git://github.com/openaps/oref0.git) || die "Couldn't clone oref0 dev"
 fi
 echo Checking oref0 installation
-( cat $(which oref0-dex-is-fresh) | grep mins && grep -q oref0_glucose_since `which nightscout` && oref0-get-profile --exportDefaults 2>/dev/null >/dev/null ) || (echo Installing latest oref0 dev && cd $HOME/src/oref0/ && npm run global-install)
+( cat $(which oref0-dex-is-fresh) | grep mins $(which oref0-dex-is-fresh) && grep -q oref0_glucose_since `which nightscout` && oref0-get-profile --exportDefaults 2>/dev/null >/dev/null ) || (echo Installing latest oref0 dev && cd $HOME/src/oref0/ && npm run global-install)
```
scottleibrand: You don't need the cat at all. This line should start with the grep.

scottleibrand:
Thanks for testing this! If you can clean up the unnecessary cat I think this is ready to merge.

tghoward:
Hmm. It still stalls for me when that line begins like this:
( grep mins $(which oref0-dex-is-fresh) && grep -q oref0_glucose_since ....

PieterGit:
don't think this will work because  `which` gives empty string, so `grep mins $(which filethatdoesnotexist)` will also hang

scottleibrand:
Implemented https://github.com/openaps/oref0/commit/bb53c26ffaa67c2505c81710640915bc60d80db5 as a workaround for now.

scottleibrand:
This has now been fixed by checking for oref0 0.3.x.

PR #255, opened 2016-11-06
truncate git history at 2000 commits

scottleibrand:
This should prevent the .git directory from growing uncontrollably.  If garbage collection fails and it does grow to fill the disk, it should now go ahead and rm -rf .git and re-init.

### DIFF
--- a/bin/oref0-reset-git.sh
+++ b/bin/oref0-reset-git.sh
@@ -50,3 +50,6 @@ fi
 # if git repository is too corrupt to do anything, mv it to /tmp and start over.
 
 (git status && git diff) > /dev/null || (echo "Saving backup to: $BACKUP" > /dev/stderr; mv .git $BACKUP; openaps init . )
+
+#truncate git history to 1000 commits if it has grown past 1500
+oref0-truncate-git-history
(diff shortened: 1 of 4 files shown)


scottleibrand:
How to test this: https://gitter.im/nightscout/intend-to-bolus?at=581ebd3345c9e3eb4302dcf7

dougtakeuti:
Scott, this seems to work now. Last night, before updating it, I had 36M. I
did not see a change for the first hour, but now I have 7.6M. Thank you for
doing this!

2016-11-10 16:21 GMT-02:00 Scott Leibrand <notifications@github.com>:

scottleibrand:
@jasoncalabrese should we add a --no-truncate option to oref0-reset-git for people syncing their myopenaps repo's with github?  or would it be better to run oref0-truncate-git as a separate cron?

jasoncalabrese:
I guess a --no-truncate option would work, has anyone tried pushing to a remote repo after truncating?  What I'd really like is a way to rotate the repo and push to new github repos, but never high enough on the list to get to it.

scottleibrand:
From what I've read it shouldn't work to push from a truncated repo, but I'm not set up to try it.

scottleibrand:
Actually I think having it be a separate cron is a lot easier all around, so I did that.

PR #263, opened 2016-11-29
Fix activityContrib

warstar2187:
Before fix.
image

After fix.
image

### DIFF
--- a/lib/iob/calculate.js
+++ b/lib/iob/calculate.js
@@ -25,7 +25,7 @@ function iobCalc(treatment, time, dia) {
             var y = (minAgo-peak)/5;
             iobContrib = treatment.insulin * (0.001323 * y * y - .054233 * y + .55556);
             //activityContrib=sens*treatment.insulin*(2/dia/60-(minAgo-peak)*2/dia/60/(60*dia-peak));
-            activityContrib = treatment.insulin * (2 / dia / 60 - (minAgo - peak) * 2 / dia / 60 / (60 * dia - peak));
+            activityContrib = treatment.insulin * (2 / dia / 60 - (minAgo - peak) * 2 / dia / 60 / (60 * 3 - peak));
         }
 
         results = {


scottleibrand:
This fixes an incorrect insulin activity calculation for DIA <> 3h, which would result in negative insulin activity (and positive BGI) at the tail end of the DIA for DIA < 3h, and erroneously high activity (and too-low BGI) for DIA > 3h. Discussion at https://gitter.im/MilosKozak/AndroidAPS?at=583cd6c116207f7b0eb97fa2

cjo20:
Nice catch! We still really do need to move away from that code though.

scottleibrand:
@cjo20 or anyone else: if you'd like, we need someone to do the derivative of the IOB calculation function at https://github.com/Perceptus/iob/blob/master/insulin_on_board_pct.js and expose that in a way that we can replace this code with that...

cjo20:
Should activityContrib be the derivative of IOBContrib?

scottleibrand:
Yes, IIRC. I created the existing IOB formula as the integral of that bilinear insulin activity formula.

mikestebbins:
I was looking at calculate.js more before I tested the proposed solution (I have a DIA = 5). I haven't run test cases on the proposed solution (sorry, I don't know enough about Node to quickly build-up a test), but doesn't the fact that L5 (https://github.com/openaps/oref0/blob/dev/lib/iob/calculate.js#L5) hard-codes a 180 minute end-point and that the L24 `else if` block (https://github.com/openaps/oref0/blob/dev/lib/iob/calculate.js#L24)  that this proposed fix resides within only applies if `minAgo < end` mean that more changes are necessary to extend this to DIA != 3? For instance, should L5 be `var end = dia * 60;`? I apologize if I'm misunderstanding the code; it all seems fairly clear, except for the use of diaratio to stretch/shrink the curve (which is what I think it's doing).

cjo20:
It shouldn't require extra changes because of this line: https://github.com/openaps/oref0/blob/dev/lib/iob/calculate.js#L15 Which scales the minAgo value to be between 0 and 180 minutes.

mikestebbins:
Thank you, Chris, I see it now. I quick built it up in Python and can see how the scaling works. I plan on implementing tonight to test.

scottleibrand:
Still need more folks to test this, but based on initial results I'm going to merge to dev and we can test there before merging dev to master.

PR #265, opened 2016-12-02
fixes iob_data props check

fox1t:


fox1t:
Ok, i'll then work on dev branch(my fault). 
iob_data.bolussnooze is used 
`var bolusContrib = iob_data.bolussnooze * sens;` and bolusContrib is used to calculate `snoozeBG` that is then return. In my opinion is always better to check if iob_data.bolussnooze is not undefined, otherwise it will be NaN.

scottleibrand:
What conditions could cause it to be missing? What is the impact on temp basal decisions if snoozeBG is NaN?

I don't want to add any strict checks that might result in the loop stopping when it would be safer to continue looping despite the error condition, as halting the loop usually results in a reduction of overall safety to the user, unless it's a fatal error that makes it impossible to calculate a temp basal.

scottleibrand:
Closing this one.  Please feel free to open a new PR targeted against dev if you think additional checks are needed there.

Issue #172, opened 2016-07-18
COB calculation problem 


eyim:
Was without a sensor for 14 hours.  When it restarted and ran the AMA calculations, the COB seemed off.  Even thought the last meal was @8:16 EDT for 100 carbs.  

How this is dangerous is that the system tried to high temp even with a falling BG because it assumed there were a lot of COB when in reality there wasn't.

Here is the example:

```
Files monitor/last-glucose.json and monitor/glucose.json differ
{"carbs":100,"boluses":8.3,"mealCOB":73}
{"ratio":1}
{"iob":0.638,"activity":0.0399,"bolussnooze":0,"basaliob":-0.05,"netbasalinsulin                                                                             ":-0.05,"hightempinsulin":0,"time":"2016-07-18T04:06:24.000Z"}
{"delta":-15,"glucose":131,"avgdelta":30.667}
{"duration":30,"rate":0.8,"temp":"absolute"}
{"max_iob":3,"type":"current","dia":5,"skip_neutral_temps":false,"current_basal"                                                                             :0.8,"max_daily_basal":1.5,"max_basal":4,"min_bg":120,"max_bg":120,"sens":70,"ca                                                                             rb_ratio":14}
Carb Impact: 3 mg/dL per 5m; CI Duration: 20.3 hours
Accel. Carb Impact: 10 mg/dL per 5m; ACI Duration: 6.1 hours
determine-basal://text/shell/enact/suggested.json
reporting enact/suggested.json
[...]

PR #270, opened 2016-12-04
Prep next oref0 release

scottleibrand:
So far, the main new feature here is support for running openxshareble in a separate cgm-loop and uploading the retrieved BG values to NS.  I have never actually gotten openxshareble working reliably, though, with this setup or any other, so I'm not sure how useful that will be.  What will be useful, though, is using the same cgm-loop structure to support uploading to NS the CGM data retrieved from a USB-connected Dexcom receiver, as we're now doing with MDT CGM.


scottleibrand:
The other new feature that needs testing here is BG interpolation for autosens and AMA.  Currently, missing data will result in AMA's COB estimates to be erroneously high, as described at https://github.com/openaps/oref0/issues/172 .  By interpolating missing data points, we can calculate the total carbs absorbed during the period where CGM data was unavailable.

scottleibrand:
Ok, starting to look at a merge of dev to master and release.  We now have BT tethering working, as well as USB OTG connected Dexcom CGM support (uploading to NS).  Autotune has been tested by a bunch of people, and is an optional advanced feature anyway.  There are a bunch of other optional advanced features in here that have had various levels of testing, but probably shouldn't hold up a release.  Anything I'm missing?

danamlewis:
(other than forgetting to note the amazing new oref0-runagain.sh, I think that covers it! lots of awesome work by so many people in this release :))

cjo20:
Nothing I'm aware of that should prevent a release, but I don't currently use oref0-setup.

cjo20:
Given that the changes to mmtune that allow the use of mmtune-old.json to speed up mmtune were released, would it be worth add that change to the loop before releasing?

scottleibrand:
Yes, if we can do so easily and it's easy to test. You have an idea of what needs to be done? Want to start a PR (or at least an issue)?

PieterGit:
@scottleibrand 
The WW-stuff is also new and almost there.  I still have issues with my PI3/TI USB/WW-pump setup on latest dev (see https://github.com/openaps/oref0/issues/354 ). I'm working on getting those resolved. The current patch also uses the pump device from `pump.ini` instead of doing plain wrong stuff for ERF/SRF etc users.  I think a key feature for the new release would also be the WW-pump stuff that is (nearly) there. I don't know if I am the only one still struggling with it, or that others can confirm that their WW-pump setup works fine on the dev branch.

I noticed the coverage really decreases quite a lot. Shouldn't we focus a week or more on adding more tests, so that we don't decrease our coverage with 8.2% to 33.26%. Each oref0-release should have the same or higher unit test coverage. Same as the focus on the docs that helped a lot.

I would also want to suggest to use one or more (feature) freezes, similar to what other major distributions do, (e.g. https://wiki.ubuntu.com/ZestyZapus/ReleaseSchedule ).  My take for openaps/oref0:

Feature freeze: At this point we stop introducing new features, packages, and APIs, and concentrate on fixing bugs, adding tests and updating docs.

Final freeze Exceptions requiring confirmation:
- Safety critical bugs
- Release critical bugs
- Security critical bugs
- Exceptional circumstances

In both freeze periods (advanced) users are asked to thumbs up or thumbs down a release based on their configuration. It's up to the release managers ( @scottleibrand  and @danamlewis ) to define the point when it's ready to be released.

scottleibrand:
@PieterGit I think we do something similar to that "feature freeze" thing, but less formally and more organically.  At some point as we get close to wanting to release dev, we start leaving major feature work on feature branch and not merging it to dev, while we finish cleaning up the features we've already added (and adding minor stuff).

We could definitely use more unit testing, but I'm not prepared to hold up release to force people (including myself) to work on unit tests if they don't want to.  I'm pretty comfortable with the level of integration testing we do.  More unit tests would definitely help in the long run, but the features we're releasing are quite useful with or without tests IMO.

And if you have some more finishing up to do on the WW stuff, that might give someone a bit of time to get some tests in...

danamlewis:
I agree unit tests would be good; but we've had no one step up to do them; and we have tested each feature as it is added to dev, along with testing dev in several other people's brand new setups. I suggest we move forward with this release in the next 24h - and if need be, do patch releases from there. There's some features in this release (specifically related to bluetooth and network switching) that drastically improves real-world usability that I think is a higher priority to get out to master-branch-level-users.

PR #273, opened 2016-12-05
Handle xDrip errorstate 38

AdrianLxM:
While Dexcom uses values at 30 and below as error states, xDrip uses the values at 31 to 38 as additional error states (only 38 used/propagated) so far.
Those values are not used by Dexcom as they map LOW to 39 and the normal range it shows values starts with 40.
NightScout shows it as error "38?? ✖︎".

I don't use Oref0 but AndroidAPS, but I hope I have found all cutoffs. I've looked at all ocurrences of "30" in the sourcecode and checked if it was a cutoff or not.

### DIFF
--- a/lib/determine-basal/cob-autosens.js
+++ b/lib/determine-basal/cob-autosens.js
@@ -37,7 +37,7 @@ function detectSensitivityandCarbAbsorption(inputs) {
         } else if (glucose_data[i-1].dateString) {
             lastbgTime = new Date(glucose_data[i-1].dateString);
         } else { console.error("Could not determine last BG time"); }
-        if (glucose_data[i].glucose < 38 || glucose_data[i-1].glucose < 38) {
+        if (glucose_data[i].glucose < 39 || glucose_data[i-1].glucose < 39) {
             continue;
         }
         var elapsed_minutes = (bgTime - lastbgTime)/(60*1000);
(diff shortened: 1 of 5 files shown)


lib/determine-basal/cob-autosens.js
```diff
@@ -37,7 +37,7 @@ function detectSensitivityandCarbAbsorption(inputs) {
         } else if (glucose_data[i-1].dateString) {
             lastbgTime = new Date(glucose_data[i-1].dateString);
         } else { console.error("Could not determine last BG time"); }
-        if (glucose_data[i].glucose < 30 || glucose_data[i-1].glucose < 30) {
+        if (glucose_data[i].glucose < 38 || glucose_data[i-1].glucose < 38) {
```
scottleibrand: Would this need to be < 39 (<= 38)?

lib/determine-basal/determine-basal.js
```diff
@@ -66,7 +66,7 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
 
     var bg = glucose_status.glucose;
     // TODO: figure out how to use raw isig data to estimate BG
-    if (bg < 30) {  //Dexcom is in ??? mode or calibrating
+    if (bg < 38) {  //Dexcom is in ??? mode or calibrating
```
scottleibrand: Would this need to be < 39 (<= 38)?

AdrianLxM:
@scottleibrand yes, I too blindly replaced the 30 with 38 :/
corrected it.

PR #275, opened 2016-12-10
Changes to Allow BT tethering 

Bender1061:


bin/oref0-setup.sh
```diff
@@ -221,6 +221,14 @@ if openaps vendor add --path . mmeowlink.vendors.mmeowlink 2>&1 | grep "No modul
     echo Installing latest mmeowlink
     sudo pip install -U mmeowlink || die "Couldn't install mmeowlink"
 fi
+# Install Bluez for BT Tethering
```
scottleibrand: I think we want to make this conditional: no sense doing all the bluez stuff unless the user is trying to set it up.

scottleibrand: Perhaps condition it on https://github.com/openaps/oref0/blob/dev/bin/oref0-setup.sh#L75 ? Although that leaves a bit of a bootstrapping issue if you don't have bluez yet and can't look up your phone's MAC...

Bender1061: good, point, but you can get your bt mac ID from your Android phone under settings/about phone/status  and on the iphone settings/general/about

scottleibrand: Ah, perfect.

Bender1061:
just made the changes selected. 

now will the change I made on oref0-online do anything bad to the shareble?

scottleibrand:
I think the way you're forcing a restart of bluetoothd should be ok, as shareble doesn't (yet) run as a daemon, so it should just retry on the next run.

Bender1061:
Not working, yet, I can't get this check of a valid Mac address to work yet, I'm doing some more testing, but  this is a lot of googling for me to try to figure this out

Bender1061:
if you have any ideas

scottleibrand:
I would skip the validity checking for now and just check if that variable is set at all.

Bender1061:
Ok, I got it to work now. Not sure it that's the best way to check for an empty string though.  I know it works great in C, or my robot code of Rapid.

scottleibrand:
The best way to do that in bash is http://serverfault.com/questions/7503/how-to-determine-if-a-bash-variable-is-empty

Bender1061:
yeah that looks a little cleaner, its up and working right now, I also tried running without the bt mac and it worked as expected. Now we need to get someone to test with shareble to make sure I did not mess up that.

bin/oref0-setup.sh
```diff
@@ -281,12 +293,8 @@ elif [[ ${CGM,,} =~ "shareble" ]]; then
     fi
     sudo apt-get -y install bc jq libusb-dev libdbus-1-dev libglib2.0-dev libudev-dev libical-dev libreadline-dev python-dbus || die "Couldn't apt-get install: run 'sudo apt-get update' and try again?"
     echo Checking bluez installation
-    if ! bluetoothd --version | grep -q 5.37 2>/dev/null; then
```
scottleibrand: Did you remove the ! here on purpose?

Bender1061: yes, cause we don't need need to copy that .conf for just using the Tethering, so it's already checking if shareble, so it should have already installed bluez, this will allow that copy of the .conf. otherwise it would not have copied it.
now does that cause a problem if someone is just re-running the script, I don't know. someone using shareble needs to test.

PR #276, opened 2016-12-13
Add radio_locale support

oskarpearson:
Related change: https://github.com/openaps/openaps/pull/112

I don't currently have a box to test this on, but "it should work" (TM). I'll try sort one out this evening.

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -40,6 +40,10 @@ case $i in
     serial="${i#*=}"
     shift # past argument=value
     ;;
+    -rl=*|--radio_locale=*)
+    radio_locale="${i#*=}"
+    shift # past argument=value
+    ;;
     -t=*|--tty=*)
     ttyport="${i#*=}"
     shift # past argument=value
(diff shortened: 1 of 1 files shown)


bin/oref0-setup.sh
```diff
@@ -377,14 +403,14 @@ if [[ "$ttyport" =~ "spi" ]]; then
...
-if [[ -z "$ttyport" ]]; then
+if [[ -z "$using_mmeowlink" ]]; then
     openaps device add pump medtronic $serial || die "Can't add pump"
     # carelinks can't listen for silence or mmtune, so just do a preflight check instead
     openaps alias add wait-for-silence 'report invoke monitor/temp_basal.json'
     openaps alias add wait-for-long-silence 'report invoke monitor/temp_basal.json'
     openaps alias add mmtune 'report invoke monitor/temp_basal.json'
 else
-    openaps device add pump mmeowlink subg_rfspy $ttyport $serial || die "Can't add pump"
+    openaps device add pump mmeowlink subg_rfspy $ttyport $serial $radio_locale || die "Can't add pump"
```
scottleibrand: We also need to update oref0-setup to install the version of openaps with radio_locale support before this will work, right?

oskarpearson:
I've made some additional changes here - so as to support MDT cgms. I think we can merge.

Currently the end-result doesn't work, since the distributed openaps version doesn't include the requisite changes, but the command run from the scripts is as follows:

`+ openaps device add cgm mmeowlink subg_rfspy /dev/ttyACM0 123123 WW`
and
`+ openaps device add cgm mmeowlink subg_rfspy /dev/ttyACM0 123123`

Currently these extra parameters are  just ignored and not written to the pump.ini / cgm.ini

scottleibrand:
Oh, nice.  I like backwards compatibility.  :)

scottleibrand:
I think what I'll do for now, given the backwards compatibility, is just merge this without any special code to ensure that the user has the openaps dev branch.  Anyone wishing to test this with oref0-setup will just need to check out and install the openaps dev branch manually before running oref0-setup.

scottleibrand:
FWIW, looks like the openaps stuff is actually *not* backwards compatible.  With the dev branch of oref0-setup, you also have to install the dev branch of openaps or it errors out when trying to pass the WW or US flag.

PR #277, opened 2016-12-14
Edison battery

Bender1061:
Script add for the Edison Voltage monitoring for NightScout. Used @cjo20 EdisonVoltage repo, puts voltage on your nightscout site.

note for right now you need to run 

cd ~/src/oref0/ && npm run global-install

if order for the 

oref0-ed-battery to run.


bin/oref0-setup.sh
```diff
@@ -364,6 +364,26 @@ elif [[ $ENABLE =~ autosens ]]; then
 elif [[ $ENABLE =~ meal ]]; then
     EXTRAS='"" monitor/meal.json'
 fi
+# Install EdisonVoltage
+if egrep -i "edison" /etc/passwd | grep "edison" 2>/dev/null; then
```
scottleibrand: Why two greps here?

bin/oref0-setup.sh
```diff
@@ -364,6 +364,26 @@ elif [[ $ENABLE =~ autosens ]]; then
...
+   else
+      echo "Installing EdisonVoltage"
+      cd ~/src && git clone -b master git://github.com/cjo20/EdisonVoltage.git || (cd EdisonVoltage && git checkout master && git pull)
+      cd ~/src/EdisonVoltage
+      make voltage_server
+      sudo ./voltage_server &
+      echo 1 > /tmp/battery_trigger
+      sudo make install
+   fi
+   cd && cd $directory || die "Can't cd $directory"
```
scottleibrand: Why two cd's here?

bin/oref0-setup.sh
```diff
@@ -393,6 +413,10 @@ fi
 (crontab -l; crontab -l | grep -q "cd $directory && oref0-reset-git" || echo "* * * * * cd $directory && oref0-reset-git") | crontab -
 #truncate git history to 1000 commits if it has grown past 1500
 (crontab -l; crontab -l | grep -q "cd $directory && oref0-truncate-git-history" || echo "* * * * * cd $directory && oref0-truncate-git-history") | crontab -
+# Install EdisonVoltage
+if egrep -i "edison" /etc/passwd | grep "edison" 2>/dev/null; then
```
scottleibrand: Same "two greps" question again.

cjo20:
The `voltage` command doesn't need the server running, it is a separate method of accessing the data. I would advise not using the server at all in this case; just running `/home/edison/src/EdisonVoltage/voltage json batteryVoltage battery` under sudo / as root is sufficient. The only `make` command you should need to run is `make voltage`.

baughaw:
Steps to get this to work.

1) make voltage
2) Add following devices/reports

[device "edison-battery"]
fields = output
cmd = sudo /root/loopy/EdisonVoltage/voltage
vendor = openaps.vendors.process
args =

[report "monitor/edison-battery.json"]
device = edison-battery
use = shell
reporter = JSON
output = json
remainder = batteryVoltage battery

3) Change the 2 following aliases
 a)  monitor-pump - report invoke monitor/edison-battery.json 
b) format-ns-status - add  --uploader monitor/edison-battery.json

baughaw:
You will obviously need to change the path in the above device definition
"cmd = sudo /root/loopy/EdisonVoltage/voltage"

Bender1061:
@cjo20 ~/src/EdisonVoltage/voltage json batteryVoltage battery will not work with out the sudo make install

baughaw:
That's correct.   That is step 1 I was mentioning.   You will need to run make voltage

Bender1061:
you're saying I need to change my order a bit

cjo20:
It doesn't just need to be changed. You can remove anything involving the `voltage-server` or `battery_trigger`. You also only need 1 call to make, which will be `make voltage`

Bender1061:
@cjo20  gotcha, changes made and it's working for me.

bin/oref0-ed-battery.sh
```diff
@@ -2,7 +2,6 @@
 
 echo Run oref0-ed-battery
 if ! /home/edison/src/EdisonVoltage/voltage json batteryVoltage battery | grep "-66" 2>/dev/null; then
-    echo start Voltage server and trigger
-    cd /home/edison/src/EdisonVoltage && sudo ./voltage_server &
-    cd /home/edison/src/EdisonVoltage && echo 1 > /tmp/battery_trigger
+    echo start Voltage
+    sudo /home/edison/src/EdisonVoltage/voltage
```
cjo20: Calling this here doesn't look necessary. You've already called it in the `if` condition (and then directed the output to NULL). What is this block of code supposed to do?

Bender1061: Whenever I tried running without doing this I would only get the the -66 for battery percentage after a reboot or cycle power on the Edison. 
when I get home from work, I will do further testing.

cjo20: That's because you're not running it as root in the if condition, which you need to do for the command to update the data. If you change the if condition to use sudo, it'll get proper data back, and the whole if statement will effectively do nothing. Every time you call `voltage` it has to be as root, or it will refuse to update the data it receives.

baughaw: Why are you installing this in crontab?  It can be added as a device 

[device "edison-battery"]
fields = output
cmd = sudo /root/loopy/EdisonVoltage/voltage
vendor = openaps.vendors.process
args =

Then just refresh the report inside of the monitor-pump alias

Bender1061:
Ok, thank you @cjo20 you were completely right. got rid of the oref0-ed-battery completely and added sudo to the call of edison-battery.

Bender1061:
What is everyone's thought on putting the shutdown in the crontab? Generally, it's not something I'm worried about with the edison, since it does not mess up the system like it does with the pi. but then there are people that use it. It would be easy enough to add, just wondering if everyone thinks it's needed?

scottleibrand:
I think I'd prefer to keep that a manual thing, where people can add it to their cron if they want.  I've never seen corruption on an Edison due to low-battery shutdown.

baughaw:
Why does reading the edison battery status need to be in cron?

baughaw:
Here is all that needs done to make this work without cron...  

`echo Adding Edison battery device
openaps device add edison-battery process sudo $HOME/EdisonVoltage/voltage || die "Can't add edison battery"
   
echo Replacing monitor-pump alias
#replace alias with one that has the voltage added
openaps alias add monitor-pump 'report invoke monitor/edison-battery.json monitor/clock.json monitor/temp_basal.json monitor/pumphistory.json monitor/pumphistory-zoned.json monitor/clock-zoned.json monitor/iob.json monitor/meal.json monitor/reservoir.json monitor/battery.json monitor/status.json'

echo Replacing monitor-pump alias
#replace alias with one that has voltage
openaps alias add format-ns-status '! bash -c \"ns-status monitor/clock-zoned.json monitor/iob.json enact/suggested.json enact/enacted.json monitor/battery.json monitor/reservoir.json monitor/status.json --uploader monitor/edison-battery.json > upload/ns-status.json'
`

Bender1061:
why not? this gives you the ability to change the amount of time it checks it, and reduces the need to edit other aliases to call it. At least that is my own opinion on the matter. as it is I wish I did not have to edit the format-ns-status. and it's really not part of monitoring the pump at all.

baughaw:
Jason's approach was to do it without cron.   He has been running this code since July.

Bender1061:
plenty of ways to skin a cat.

baughaw:
Your approach has no way to tell if something is failing with it when you run it using cron.

Bender1061:
This is good discussion. 
But you can tell that with the log file? 
Keeping it independent keeps your loop running. If something were to fail, I would much rather my Nightscout not update than to have my loop fail.

Bender1061:
realized, I pushed this to master. Should I have put this on Dev? @scottleibrand

Bender1061:
and so far everything is working well on my explorer boards too

scottleibrand:
Yes, PRs should go against oref0 dev.

scottleibrand:
Or we could do https://github.com/openaps/oref0/pull/278

Bender1061:
really think we should keep things like Battery separate from pump. We keep the Wifi and BT separate. and who knows about the future with hopefully someday getting other pumps in the whole works. Keeping system systems separate from pump systems seams more logical to me.

Bender1061:
but if the conscientious is to go the other route, we should probably load the changes the same way that we load everything else instead of doing it in the oref0-setup.

Bender1061:
and to keep things out of cron. make battery, part of  pump-loop or ns-loop

Bender1061:
I could make those changes pretty fast on this pull request

ehwest:
I've got three Intel Edison devices here under test for battery lifetime.
They are running openaps and nightscout with mongodb.
How can I add measures of the battery power?

Bender1061:
did you run the setup with the script from oref0-Setup?
if you did , you could just run this command and then run the oref0-setup just like from the manual

$ mkdir -p ~/src; cd ~/src && git clone -b EdisonBattery git://github.com/Bender1061/oref0.git || (cd oref0 && git checkout EdisonBattery && git pull)

emsamom:
Suggest you add instructions to enable the upbat plugin for Nightscout.  Nightscout documentation explains the upbat plugin is used to monitor the phone that is uploading data from CGM, but in our case we are using it for the explorer board.  Was a bit confusing to figure out based on nightscout documentation and no mention here.

Thanks to all for figuring this out, it's very needed!

scottleibrand:
Merged #281

PR #278, opened 2016-12-14
Monitor Edison battery WITHOUT cron

baughaw:


scottleibrand:
Are there any situations where the aliases we're replacing are different (perhaps for MDT CGM or something)?

baughaw:
Not that I could see.  The two aliases seem independent of sensor choice.

baughaw:
It is possible..  let me check :(

baughaw:
Yep it is missing.  

openaps report add monitor/edison-battery2.json JSON edison-battery shell json batteryVoltage battery

scottleibrand:
K. Can you commit that change to your branch so it gets added to the PR?

bin/oref0-setup.sh
```diff
@@ -356,6 +356,35 @@ elif [[ ${CGM,,} =~ "G4" || ${CGM,,} =~ "shareble" ]]; then
...
+      cd ~/src/EdisonVoltage
+      make voltage
+   fi
+ 
+    
+   echo Adding Edison battery device
+   openaps device add edison-battery process sudo $HOME/EdisonVoltage/voltage || die "Can't add edison battery"
+   
+   echo Adding Edison battery report
+   openaps report add monitor/edison-battery2.json JSON edison-battery shell json batteryVoltage battery
```
scottleibrand: I think you need to remove the 2?

baughaw:
I believe I committed it..  Did I do something wrong?

scottleibrand:
Just had only seen your comment not your commit when I posted that. Flaky LTE.

bin/oref0-setup.sh
baughaw: ha, yep, I was testing on my live rig. :)   Hang on

Bender1061:
how about to keep it out of a cron job, not have edison-battery as part of pump-loop alias or better yet ns-loop alias. That way it keeps it out of monitor-pump, which the Battery is not part of, and semi keeps it separate from the pump just like the BT and the wifi are?

Bender1061:
should probably also make the changes to load the changes from a Json file, like how we add all the other reports and alias from. that way it keeps the oref0-setup script as clean as we can (granted that is getting difficult with the amount of awesomeness everyone had been adding to it)

scottleibrand:
Yeah, I think @Bender1061's suggestion of putting edison-battery in the ns-loop alias, instead of in monitor-pump, makes more sense.  @baughaw you want to adjust that, or should I?

Bender1061:
add an echo to the edison-battery, so that you can see it posting to the ns-loop.log 

like this

echo Edison Battery at $(date): $(sudo ~/src/EdisonVoltage/voltage json batteryVoltage battery) && sudo ~/src/EdisonVoltage/voltage json batteryVoltage battery

leave the sudo in there, so it will work with people not running the explorer board.

baughaw:
@scottleibrand Hi Scott, I am traveling today..  So you can make the change if you like or wait for me until tomorrow.  I wont be offended :)

baughaw:
@Bender1061 Would this cause the code to run twice?

echo Edison Battery at $(date): $(sudo ~/src/EdisonVoltage/voltage json batteryVoltage battery) && sudo ~/src/EdisonVoltage/voltage json batteryVoltage battery

Seems like that is an expensive operation since the voltage code writes and reads from 2 files each time it runs.

scottleibrand:
Firing up new python environments for each `openaps` call is probably a lot more expensive than reading voltage twice. :-) That said, running it once and then doing a `cat` of the output file would ensure that what is in the log matches what is reported to NS.

PR #280, opened 2016-12-18
bluetoothd for shareble

tghoward:
I got sed to modify /etc/rc.local and incorporated them here.  This needs testing, especially to make sure there isn't conflict with the other ble language in there, but I think it is ok.  I'll be starting another Edison from scratch, hopefully tomorrow and will be able to more fully test. Others should too, however.


tghoward:
Ok, tested script with a fresh install. BLE to G4 worked for me after a reboot!

PR #282, opened 2016-12-20
Dexusb cgm loop

PieterGit:
- initial start for dexusb + world wide pump
- added oref0-find-ti-usb-device.sh which works on ti usb sticks


package.json
```diff
@@ -57,6 +58,8 @@
     "oref0-dex-is-fresh": "./bin/oref0-dex-is-fresh.sh",
     "oref0-dex-time-since": "./bin/oref0-dex-time-since.sh",
     "oref0-dex-wait-until-expected": "./bin/oref0-dex-wait-until-expected.sh",
+	"oref0-subg-ww-radio-parameters": "./bin/oref0-subg-ww-radio-parameters.sh"
```
scottleibrand: Missing commas on these.

PieterGit:
@scottleibrand : can you merge this to dev? I have this running on a rpi3 without problems

scottleibrand:
Sorry, I keep meaning to test this to see whether it's backwards compatible with the existing method, but haven't gotten around to it.  Can you describe how this interacts with that?

PieterGit:
It's backwards compatible. If you don't set dexusb everything works the same.

When you set the EXTRA bit dexusb the normal get-bg loop will be replaced with a python script that will write `raw-cgm/raw-entries.json` every minute (if there are new records on the Dexcom G4 connected via USB), and then call the `openaps get-bg`. The python script will also upload the new cgm entries to Nightscout so other rig's can use the cgm.

The `oref0-find-ti` will show the location of the TI-stick, which is usefull if you have both a TI-stick and Dexcom connected so that you can find out on which ttyACM[0-9] the TI is available.

The `bin/oref0-subg-ww-radio-parameters.sh` is for setting the subg ww parameters. It can timeout and stop the pump loop if the pump is not available, so therefore I created a timeout wrapper for it. I see that I still need to add the call to `oref0-subg-ww-radio-parameters-timeout` at the start of `mmtune` in `oref0-setup.sh`

bin/oref0-setup.sh
```diff
@@ -103,7 +103,7 @@ if ! ( git config -l | grep -q user.name ); then
     git config --global user.name $NAME
 fi
 if [[ -z "$DIR" || -z "$serial" ]]; then
-    echo "Usage: oref0-setup.sh <--dir=directory> <--serial=pump_serial_#> [--tty=/dev/ttySOMETHING] [--max_iob=0] [--ns-host=https://mynightscout.azurewebsites.net] [--api-secret=myplaintextsecret] [--cgm=(G4|shareble|G5|MDT)] [--bleserial=SM123456] [--blemac=FE:DC:BA:98:76:54] [--btmac=AB:CD:EF:01:23:45] [--enable='autosens meal'] [--radio_locale=WW]"
+    echo "Usage: oref0-setup.sh <--dir=directory> <--serial=pump_serial_#> [--tty=/dev/ttySOMETHING] [--max_iob=0] [--ns-host=https://mynightscout.azurewebsites.net] [--api-secret=myplaintextsecret] [--cgm=(G4|shareble|G5|MDT|xdrip)] [--bleserial=SM123456] [--blemac=FE:DC:BA:98:76:54] [--btmac=AB:CD:EF:01:23:45] [--enable='autosens meal dexusb'] [--radio_locale=()WW|US)]"
```
scottleibrand: Is the radio local () thing here a typo?

bin/oref0-setup.sh
```diff
@@ -503,6 +505,10 @@ fi
 echo Running: openaps report add enact/suggested.json text determine-basal shell monitor/iob.json monitor/temp_basal.json monitor/glucose.json settings/profile.json $EXTRAS
 openaps report add enact/suggested.json text determine-basal shell monitor/iob.json monitor/temp_basal.json monitor/glucose.json settings/profile.json $EXTRAS
 
+# create .profile so that openaps commands can be executed from the command line
```
scottleibrand: It would be better if we could eliminate the need for these environment variables and have oref0 use the `ns` device instead. If you don't want to take that on, probably should add a TODO to remove this workaround when it's done.

bin/oref0-setup.sh
```diff
@@ -531,7 +537,11 @@ elif [[ ${CGM,,} =~ "xdrip" ]]; then
     (crontab -l; crontab -l | grep -q "cd $directory && ps aux | grep -v grep | grep -q 'openaps monitor-xdrip'" || echo "* * * * * cd $directory && ps aux | grep -v grep | grep -q 'openaps monitor-xdrip' || ( date; openaps monitor-xdrip) | tee -a /var/log/openaps/xdrip-loop.log; cp -up $directory/xdrip/glucose.json $directory/monitor/glucose.json") | crontab -
     (crontab -l; crontab -l | grep -q "xDripAPS.py" || echo "@reboot python $HOME/.xDripAPS/xDripAPS.py") | crontab -
 elif ! [[ ${CGM,,} =~ "mdt" ]]; then
+  if ! [[ $ENABLE =~ dexusb ]]; then
```
scottleibrand: Would it be clearer if we reversed this and did "if dexusb" and the contents of your else clause, and then put the existing text in the else?

package.json
```diff
@@ -49,6 +49,7 @@
     "ns-status": "./bin/ns-status.js",
     "nightscout": "./bin/nightscout.sh",
     "ns-dedupe-treatments": "./bin/ns-dedupe-treatments.sh",
+	"oref0-find-ti": "./bin/oref0-find-ti.sh",
```
scottleibrand: Tabs vs. spaces?

bin/oref0-setup.sh
PieterGit: Yes it was a typo (or I can try to blame NotePad++). Fixed

bin/oref0-setup.sh
PieterGit: added TODO and created issue https://github.com/openaps/oref0/issues/299 for eliminate the environement variables

bin/oref0-setup.sh
PieterGit: fixed

package.json
PieterGit: fixed. also sorted the items under `bin`

scottleibrand:
Looks good. Ready to merge to dev?

PieterGit:
i think it ready... can install it without problems. please merge to dev

PR #284, opened 2016-12-22
restart networking completely instead just cycling wlan0

kenstack:
this has proved more stable for me across some wifi netoworks, BUT NEEDS TESTING BY OTHERS

### DIFF
--- a/bin/oref0-online.sh
+++ b/bin/oref0-online.sh
@@ -3,21 +3,20 @@ MAC=$1
 echo -n "At $(date) my IP is: "
 if ! curl -m 15 icanhazip.com; then
     echo -n "Error, cycling networking "
-   # sudo ifdown wlan0
-  #  sudo ifup wlan0
-  #  echo -n "and getting new wlan0 IP"
-   # ps aux | grep -v grep | grep -q "dhclient wlan0" && sudo killall dhclient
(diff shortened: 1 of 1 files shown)


bin/oref0-online.sh
```diff
@@ -2,13 +2,17 @@
...
-    sudo ifdown wlan0
-    sudo ifup wlan0
-    echo -n "and getting new wlan0 IP"
-    ps aux | grep -v grep | grep -q "dhclient wlan0" && sudo killall dhclient
-    sudo dhclient wlan0 -r
-    sudo dhclient wlan0
+    echo -n "Error, cycling networking "
+   # sudo ifdown wlan0
+  #  sudo ifup wlan0
+  #  echo -n "and getting new wlan0 IP"
```
scottleibrand: Does `/etc/init.d/networking stop` kill dhclient?  If not, it might be worthwhile keeping this part (and put it after the networking restart) in case we need still to kick DHCP.

kenstack: as far as I know it stops everything - Ive never had to restart DHCP but its a fair question - is there a way to confirm?

scottleibrand: There's no mention of DHCP or dhclient in /etc/init.d/networking, so probably worth leaving it just in case...

PR #286, opened 2016-12-24
Adding xdrip offline support

colinlennon:
Added support for xdrip as cgm source, using xDripAPS microservice.

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -462,7 +462,7 @@ fi
 
 # xdrip CGM (xDripAPS)
 if [[ ${CGM,,} =~ "xdrip" ]]; then
-    echo xdrip selected as CGM, so congifuring xDripAPS
+    echo xdrip selected as CGM, so configuring xDripAPS
     sudo apt-get install sqlite3 || die "Can't add xdrip cgm - error installing sqlite3"
     sudo pip install flask || die "Can't add xdrip cgm - error installing flask"
     sudo pip install flask-restful || die "Can't add xdrip cgm - error installing flask-restful"
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -447,6 +447,7 @@ if [[ ${CGM,,} =~ "xdrip" ]]; then
     sudo pip install flask || die "Can't add xdrip cgm - error installing flask"
     sudo pip install flask-restful || die "Can't add xdrip cgm - error installing flask-restful"
     git clone https://github.com/colinlennon/xDripAPS.git $HOME/.xDripAPS
+    mkdir -p $HOME/.xDripAPS_data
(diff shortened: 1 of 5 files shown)


colinlennon:
Thanks! As soon as I have time, I'll pull the docs and add some details. Should I pull the dev docs, or master?

PieterGit:
Hi Colin,

Looks good. Might also be usefull for my dex usb branch, see 
https://github.com/openaps/oref0/pull/282

I noticed a possible security issue on lines 
https://github.com/colinlennon/xDripAPS/blob/master/xDripAPS.py#L79 and 
https://github.com/colinlennon/xDripAPS/blob/master/xDripAPS.py#L123
You don't do any input checking as far as a I see, see
https://www.owasp.org/index.php/Testing_for_SQL_Injection_%28OTG-INPVAL-005%29

scottleibrand:
Most docs PRs go to master. We save dev for major restructurings that we need to preview in RTD and collaborate on.

sulkaharo:
Did xDrip support multiple REST upload targets? Wondering what the optimal setup for this would be. We have three active loops and sync everything to Nightscout, so what could work is xDrip pushing locally to all three if they're present (using .local domains) as well as Nightscout. @scottleibrand what's the default loop behavior if SGV data is available from multiple sources? Or was the script single-source as is?

jasoncalabrese:
You should be able to give xdrip space separated URIs, but at least with some old uploader versions if an upload failed to one of the destinations it would prevent/slow uploads to the others.

scottleibrand:
@sulkaharo we have updated oref0 to support multiple CGM sources uploading to NS, and routinely have both a rig and Dexcom Share data being uploaded there, sometimes from differently calibrated receivers. @jasoncalabrese has also tested using different sensors. When the readings are off by more than about 1 mmol, you can get weird oscillations depending on the timing of uploads, any upload gaps, and the timing of oref0 runs. But even then it tends to all average out to do the right thing. If they're pretty close it works great.

libxmike:
installing fresh with xdrip and getting error and exit:
Checking /home/pi/myopenaps: /home/pi/src/oref0/bin/oref0-setup.sh: line 231: openaps: command not found
Can't init /home/pi/myopenaps

colinlennon:
@libxmike Is it a completely fresh install? Looks like you might not have the OpenAPS toolset installed. If you're not sure, you could check by creating a test directory, cd into it and see if 'openaps init' works or not.

colinlennon:
@PieterGit Thanks Pieter. Yes, you're correct, I should add checks for SQL injection attacks to the code, that's on my to-do list.

libxmike:
@colinlennon you right, I did not run on this rig openaps install. now I can install. Got confused with many of my SD Card images I have.

PR #287, opened 2016-12-26
Remove get-ns-bg from ns-loop for MDT cgm

tmecklem:
oref0-setup provides a mechanism to fetch NS cgm in addition to device cgm so that the newer of the two can be used in the loop. This change removes that Nightscout entries fetch when using `MDT` cgm.

Reasons for the change:
1. NS fetch is unnecessary with MDT cgm. There is a very low likelihood that NS will have newer sgvs than the pump device when using `MDT` as the cgm source for oref0. 
2. Reliability of NS entries upload. The NS fetch has been observed to cause a problem with the `recent-missing-entries` report which causes duplicate entries and inconsistency in the NS uploaded entries.

Related PRs:
https://github.com/openaps/oref0/pull/288
https://github.com/openaps/decocare/pull/16

### DIFF
--- a/lib/oref0-setup/mdt-cgm.json
+++ b/lib/oref0-setup/mdt-cgm.json
@@ -48,6 +48,13 @@
       "command": "report invoke monitor/cgm-mm-glucosedirty.json cgm/cgm-glucose.json"
     }
   },
+    {
+    "type": "alias",
+    "name": "ns-loop",
+    "ns-loop": {
+      "command": "! bash -c \"echo Starting ns-loop at $(date): && openaps ns-temptargets && echo -n Refreshed temptargets && openaps ns-meal-carbs && echo \\\" and meal-carbs\\\" && openaps upload\""
+    }
+  },
   {
     "pump-loop": {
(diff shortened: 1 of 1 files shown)


Bender1061:
Here is my take on this. if you are also using a mmconnect to upload your bg to nightscout. (i know it's redundant) but what I really like about it is if you BG is over 400, nothing shows up for what your BG is and after 30 mins of being 400, basically the loop stop doing anything. When you have your mmconnect sending BG to nightscout, the BG just stays at 400, and the loop will keep running, thinking you are 400. So it then once it brings your BG back down you are golden. I know this is an extreme situation, but it does work very well for bringing you back down overnight. And brings you back down in a much safer fashion than guessing and giving yourself a giant bolus. 
What I would like more, is OpenAPS read over 400 as 400, so it would do the same thing as the mmconnect.

tmecklem:
@Bender1061 The high and low glucose sensor readings from openaps should be fixed by https://github.com/openaps/decocare/pull/16.

Bender1061:
Oh Wow, ok, now this I'm going to take a look at! that is awesome. Ok with that change, This is a great idea.

PR #289, opened 2016-12-26
Support loading ISF and BG targets in mmol/L

sulkaharo:
This will mean users with a pump in mmol mode don't need to add a conversion step to their loop script


scottleibrand:
Looks good, but probably should merge this to dev first rather than straight to master.

PR #296, opened 2016-12-28
Questioning why "type": = "current"

danamlewis:
@jasoncalabrese, @bewest do you know why this is in? Not seeing it do anything... if no one objects, will try removing it and seeing what it breaks

### DIFF
--- a/lib/profile/index.js
+++ b/lib/profile/index.js
@@ -7,7 +7,7 @@ var carb_ratios = require('./carbs');
 function defaults ( ) {
   var profile = {
     max_iob: 0 // if max_iob is not provided, never give more insulin than the pump would have
-    , type: 'current'
+   // , type: 'current'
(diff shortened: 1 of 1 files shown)


bewest:
IIRC, it's for nightscout, since NS has ability to use many profiles.

jasoncalabrese:
I looked at it once before and couldn't find a use for it, can probably be removed

sulkaharo:
Yeah just remove it. The Nightscout data format is as follows:

sulkaharo:
Just delete it? Nightscout stores a datetime field called startDate for each changed profile and the profile object you get from the API has a field called defaultProfile that points to the profile object that's supposed to be used currently. So looks to me like the idea behind adding this was good but the implementation would need refactoring to be meaningful.

danamlewis:
👍 all fingers point to merging the deletion, so doing so now.

PR #302, opened 2016-12-29
Install from source if the oref0 git clone is there. Use (updated) oref0 standard published package if there is no git oref0 directory

PieterGit:
i noticed the oref0 version was still 0.3.3 so it used the published package instead of the git version in ~/src/oref0 if the ~/src/oref0 is a git repository then install from source. 

i think the Cloning of oref0 should not be necessary when people use a published package

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -248,13 +248,16 @@ mkdir -p $HOME/src/
 if [ -d "$HOME/src/oref0/" ]; then
     echo "$HOME/src/oref0/ already exists; pulling latest"
     (cd ~/src/oref0 && git fetch && git pull) || die "Couldn't pull latest oref0"
+    echo (Re)installing oref0 installation from ~/src/oref0
+    cd $HOME/src/oref0/ && npm run global-install
 else
     echo -n "Cloning oref0: "
     (cd ~/src && git clone git://github.com/openaps/oref0.git) || die "Couldn't clone oref0"
+    echo Checking oref0 installation
+    npm list -g oref0 | egrep oref0@0.3. || (echo Installing latest oref0 && sudo npm install -g oref0)
 fi
-echo Checking oref0 installation
-npm list -g oref0 | egrep oref0@0.3. || (echo Installing latest oref0 && sudo npm install -g oref0)
-#(echo Installing latest oref0 dev && cd $HOME/src/oref0/ && npm run global-install)
+
+#(echo Installing latest oref0 dev && 
 
 echo Checking mmeowlink installation
 if openaps vendor add --path . mmeowlink.vendors.mmeowlink 2>&1 | grep "No module"; then
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -241,7 +241,7 @@ mkdir -p settings || die "Can't mkdir settings"
 mkdir -p enact || die "Can't mkdir enact"
 mkdir -p upload || die "Can't mkdir upload"
 if [[ ${CGM,,} =~ "xdrip" ]]; then
-	mkdir -p xdrip || die "Can't mkdir xdrip"
+    mkdir -p xdrip || die "Can't mkdir xdrip"
 fi
 
 mkdir -p $HOME/src/
@@ -251,17 +251,18 @@ if [ -d "$HOME/src/oref0/" ]; then
     OLD_HEAD=$(git rev-parse HEAD)
     (git fetch && git pull) || die "Couldn't pull latest oref0"
     NEW_HEAD=$(git rev-parse HEAD)
-    if [ "$OLD_HEAD" == "$NEW_HEAD" ];  then
+    if [ "$OLD_HEAD" != "$NEW_HEAD" ];  then
        echo "(Re)installing oref0 installation from ~/src/oref0. version=$NEW_HEAD"
-       npm run global-install
+       sudo npm install -g .
+       #npm run global-install
     else 
        echo "Not need to reinstall. version=$NEW_HEAD is still current"
     fi
 else
    echo -n "Cloning oref0: "
-    (cd ~/src && git clone git://github.com/openaps/oref0.git) || die "Couldn't clone oref0"
-    echo Checking oref0 installation
-    npm list -g oref0 | egrep oref0@0.3. || (echo Installing latest oref0 && sudo npm install -g oref0)
+   (cd ~/src && git clone git://github.com/openaps/oref0.git) || die "Couldn't clone oref0"
+   echo Checking oref0 installation
+   npm list -g oref0 | egrep oref0@0.3. || (echo Installing latest oref0 && sudo npm install -g oref0)
 fi
 
 echo Checking mmeowlink installation
@@ -280,9 +281,10 @@ fi
 cat preferences.json
 git add preferences.json
 
-# enable log rotation
-sudo cp $HOME/src/oref0/logrotate.openaps /etc/logrotate.d/openaps || die "Could not cp /etc/logrotate.d/openaps"
-sudo cp $HOME/src/oref0/logrotate.rsyslog /etc/logrotate.d/rsyslog || die "Could not cp /etc/logrotate.d/rsyslog"
+#enable log rotation
+#DEPRECATED: See https://github.com/openaps/oref0/pull/302
+#sudo cp $HOME/src/oref0/logrotate.openaps /etc/logrotate.d/openaps || die "Could not cp /etc/logrotate.d/openaps"
+#sudo cp $HOME/src/oref0/logrotate.rsyslog /etc/logrotate.d/rsyslog || die "Could not cp /etc/logrotate.d/rsyslog"
 
 test -d /var/log/openaps || sudo mkdir /var/log/openaps && sudo chown $USER /var/log/openaps || die "Could not create /var/log/openaps"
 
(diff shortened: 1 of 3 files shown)


PieterGit:
@scottleibrand can you review this patch and merge it to dev if you agree

bin/oref0-setup.sh
```diff
@@ -248,13 +248,16 @@ mkdir -p $HOME/src/
...
     echo -n "Cloning oref0: "
     (cd ~/src && git clone git://github.com/openaps/oref0.git) || die "Couldn't clone oref0"
+    echo Checking oref0 installation
+    npm list -g oref0 | egrep oref0@0.3. || (echo Installing latest oref0 && sudo npm install -g oref0)
 fi
-echo Checking oref0 installation
-npm list -g oref0 | egrep oref0@0.3. || (echo Installing latest oref0 && sudo npm install -g oref0)
-#(echo Installing latest oref0 dev && cd $HOME/src/oref0/ && npm run global-install)
+
+#(echo Installing latest oref0 dev &&
```
scottleibrand: Syntax error here?  Did you mean to truncate this commented line?

bin/oref0-setup.sh
```diff
@@ -248,13 +248,16 @@ mkdir -p $HOME/src/
 if [ -d "$HOME/src/oref0/" ]; then
     echo "$HOME/src/oref0/ already exists; pulling latest"
     (cd ~/src/oref0 && git fetch && git pull) || die "Couldn't pull latest oref0"
+    echo (Re)installing oref0 installation from ~/src/oref0
+    cd $HOME/src/oref0/ && npm run global-install
```
scottleibrand: Do we need to re-install every time?  Or is there some quicker way we can check if we're up to date?  Maybe only do global-install if `git pull` updated anything?

PieterGit: is there a better way then `git show | head -n 1` and store the checksum?

bin/oref0-setup.sh
PieterGit: meant to remove the comment line

scottleibrand:
One other thought: won't this mean always installing from source, since we always clone the oref0 repo in oref0-setup?  Maybe we should check and only install from source if the local oref0 clone is on a non-master branch...

PieterGit:
this should do the trick... i think. shell programming is not my cup of tea.
`sudo npm install -g oref0` will install the latest oref0 release package
if you don't checkout other branches or stuff like that that will be master and so next oref0-setup.sh won't reinstall . 

one possible loophole: if you have git cloned the repo yourselves it will only install oref0 if there are changes in the branch you checkout. 

Should step 1 on https://openaps.readthedocs.io/en/dev/docs/walkthrough/phase-2/oref0-setup.html be changed to use packaged version anyhow? `sudo npm install -g oref0` 

I'm going to stop for tonight.

bin/oref0-setup.sh
```diff
@@ -247,14 +247,22 @@ fi
 mkdir -p $HOME/src/
 if [ -d "$HOME/src/oref0/" ]; then
     echo "$HOME/src/oref0/ already exists; pulling latest"
-    (cd ~/src/oref0 && git fetch && git pull) || die "Couldn't pull latest oref0"
+    cd ~/src/oref0
+    OLD_HEAD=$(git rev-parse HEAD)
+    (git fetch && git pull) || die "Couldn't pull latest oref0"
+    NEW_HEAD=$(git rev-parse HEAD)
+    if [ "$OLD_HEAD" == "$NEW_HEAD" ];  then
```
scottleibrand: Does this need to be a != ?

PieterGit:
Why should a novice user want to use the git clone repo? I think the instructions for non-developpers should be changed to:
```
sudo npm install -g oref0
oref0-setup
```

and for the advanced users / building from src the current `git clone` and `~/src/oref0/bin/oref0-setup.sh`  instructions

bin/oref0-setup.sh
```diff
@@ -247,14 +247,22 @@ fi
...
+       npm run global-install
+    else 
+       echo "Not need to reinstall. version=$NEW_HEAD is still current"
+    fi
 else
-    echo -n "Cloning oref0: "
+   echo -n "Cloning oref0: "
     (cd ~/src && git clone git://github.com/openaps/oref0.git) || die "Couldn't clone oref0"
+    echo Checking oref0 installation
+    npm list -g oref0 | egrep oref0@0.3. || (echo Installing latest oref0 && sudo npm install -g oref0)
```
scottleibrand: If we're going to install from source any time HEAD changes, doesn't that make this part irrelevant?

bin/oref0-setup.sh
scottleibrand: Perhaps we should add a condition here that `git branch` is not on master.  If it is, we should be able to just install the packaged version...

scottleibrand:
The user doesn't have to clone oref0 - oref0-setup does it, for purposes of being able to do things like https://github.com/openaps/oref0/blob/dev/bin/oref0-setup.sh#L275-L277

bewest:
I think it would help to have a proper release workflow and life cycle, similar to what I did with openaps and other projects.  Using the package manager along with a good release cycle conforming to semantic versioning goes a long way and eliminates needing to `git clone ... && cd ... &`.  This also helps use the standard system locations and avoid special/magic paths, which is good practice as well.

Here's a tip:
`sudo npm install -g git://github.com/openaps/oref0.git#master` always installs the master branch, while `sudo npm install -g git://github.com/openaps/oref0.git#dev` always installs the dev branch, etc.  `oref0-setup.sh` should probably use these forms to avoid special directories and magic paths.  [npm manual](https://docs.npmjs.com/cli/install)

The git clone instructions are for other developers who are expecting to send pull requests back and actually need to set up a git repo.  Most users shouldn't need this.

PieterGit:
@scottleibrand i think the lines https://github.com/openaps/oref0/blob/dev/bin/oref0-setup.sh#L275-L277 should be installed by the `npm install -g` magic and not by `oref0-setup.sh`

I totally agree that most users shouldn't set up or clone a git repo to be able to install a software product.

I would also advice a bit of Python zen https://www.python.org/dev/peps/pep-0020/
```
There should be one-- and preferably only one --obvious way to do it.
Although that way may not be obvious at first unless you're Dutch.
``` 
So installing should not require `npm run global-install` but just use the standard `npm install -g`

Pieter (dutch btw 😃 )

scottleibrand:
If you know the magic incantation to do the logrotate thing with npm, I'd love to be able to move that over: please feel free to add it to the PR!  :-)

PieterGit:
according to npm docs `install` and `postinstall` will be called by npm. I don't know if we need install or postinstall. I moved the logrotate stuff to the postinstall.

@bewest @scottleibrand Can you review these changes?

I noticed a lot of dependencies on the git checkout, e.g. importing the jsons.

PieterGit:
@scottleibrand how to proceed with this pull request. will you merge the stuff that's already in here to dev, or will we do stuff diffently in future?

package.json
```diff
@@ -4,7 +4,9 @@
   "description": "openaps oref0 reference implementation of the reference design",
   "scripts": {
     "test": "make test",
-    "global-install": "npm install && sudo npm install -g && sudo npm link && sudo npm link oref0"
+    "deprecated-global-install" : "npm install && sudo npm install -g && sudo npm link && sudo npm link oref0",
+    "global-install": "Error: global-install is deprecated. Please use 'npm install -g oref0' for a packaged version, or 'npm install -g .' for a git checkout. See https://github.com/openaps/oref0/pull/302",
```
scottleibrand: I don't think this accomplishes what `global-install` is for: it doesn't do the `npm link` portion, so you end up having to re-run every time you change any little thing.

PieterGit: @scottleibrand 
I don't see what you mean. I deprecated the `global-install` to `deprecated-global-install`. 
Currently autotune branch always installs oref@0.3.3 with `oref0-setup.sh` while currently the branch is at 0.3.5-dev and so it would not install oref0-autotune in `/usr/local/bin`.
When I use `npm install -g` it will install and link the new files (e.g. `/usr/local/bin/oref0-autotune`. So I think we should stop using non stantard install procedures like `global-install`

See also: https://github.com/PieterGit/oref0/blob/oref0_version/bin/oref0-setup.sh#L247-L266

scottleibrand: My understanding is that `npm install -g` does *not* create symlinks in the globally installed directories pointing to the source directory, but rather copies the files there.  That is what `npm link` is for.  Is that incorrect?

PieterGit: @scottleibrand 
According to https://docs.npmjs.com/cli/link I believe your right. But why should we a normal user (non developer) link the files in the package? The current dev branch always installs the packaged version, see https://github.com/openaps/oref0/blob/dev/bin/oref0-setup.sh#L256 so we should not need a git clone.

Unfortuately there are quite still 10 references to $HOME/src/oref0. Maybe we should stop using (relative) links to files in the package in the script and use `curl` or `wget` to get those files from github.com.

Maybe @bewest knows a smart way that's easy for both the developer (so that you don't have to re-run every time you change something), and a standard deployment way.

I think git cloning and then using a packaged version is a confusing way of deployment.

scottleibrand: Sounds like we're trying to accomplish the same thing: we want something like `npm run global-install` (off a manual clone) for developers who need it, but for regular users we want everything to be installed via packages, with some other method (curl/wget would probably work, although not sure it's the cleanest) to pull in files we need that npm can't install.  The trick, I think, is making sure that if a developer does `npm run global-install`, that we don't overwrite that with a packaged install...

scottleibrand:
@PieterGit does https://github.com/openaps/oref0/pull/369 solve most of what we're trying to do here?  What else still needs to be done after that?

PR #303, opened 2016-12-30
oref0-autotune-test.sh script add

jyaw:
Added oref0-autotune-test.sh script to make testing oref0 autotune easier 😄. Script arguments allow the user to specify:

- OpenAPS Directory (--dir)
- Nightscout URL (--ns-host)
- Start Date (--start-date)
- End Date (--end-date)
- Number of Runs (--runs)

1. Note that the pump profile is pulled from the following location: /settings/profile.json in users's specified OpenAPS Directory. Pro Tip: this is the only directory structure requirement, so you could setup PARENT_FOLDER/settings/profile.json anywhere without other OpenAPS files.
2. Also note that End Date and Number of Runs are not required parameters, but the script will default to the day before today as the end date and 5 runs, so you may or may not want to use those. 

Example Usage: ./oref0-autotune-test.sh --dir=openaps --ns-host=<NS URL> --start-date=2016-12-9 --end-date=2016-12-10 --runs=2


danamlewis:
Merging now, will test on my end tomorrow!

PR #304, opened 2016-12-30
Added 'set -eu' to all bash scripts

ijustlovemath:
set -eu will prevent strange errors that occur when bash variables aren't set properly. There may be a need to remove the -e tag in the git-related scripts, as some of the git utilities may exit with an error when, for example, detecting corruption.


scottleibrand:
This seems like a really big change to just apply blindly.  Perhaps we should identify the places where it's actually needed, and start there, maybe with just `set -u`?

ijustlovemath:
`set -u` will just raise an error in $?, which doesn't matter much if you aren't regularly checking for exit status. I do think we should start there, but on basic format conversion scripts, having `-eu` will have no ill effects, and will actually help capture weird errors as they happen. I've tested the ones I can with no issues so far, but everything that's meant to interface with a pump or CGM is outside what I can do, as I have neither.

oskarpearson:
It's generally-accepted best practice to use `set -e` and `set -u`  http://kvz.io/blog/2013/11/21/bash-best-practices/

However... I'm sure that this won't be without problems - because some scripts will probably not catch errors that they should catch. IMHO it'd be the right decision in the long run, but will probably cause a period of instability.

My preference is to also use `set -o pipefail` btw

scottleibrand:
Do you think that any problems we see with `set -u` and/or `set -e` will be immediately obvious if we merge this to dev and try to loop on it?  Or are they likely to be more subtle things?  I'm most worried about the effect of those flags on openaps aliases, but not sure if these changes actually apply there, or just inside scripts...

ijustlovemath:
The problem you will see will be immediate, if anything is to happen at all. What this does is help avoid obscure problems you sometimes see in these custom bash scripts where somebody, for example, hasn't done an `export API_SECRET=...` in their bashrc. When those variables are not set, the script will exit and tell you why it exited (unbound variable). This is actually a good thing, as it's just another piece of information you can use to debug these sorts of problems. That's why, as Oskar mentioned, it's a standard practice of bash scripting.

If you loop on it and for some reason the `set -e` will cause problems, your initial setup will not work. Should you see that, you'll know that `-e` is maybe not the right option given the lack of error handling in openAPS scripts.

ijustlovemath:
I'd also like to mention that if you do test with the result of a command, like in a while loop or an if [[]] statement, the `-e` will not trap that. This comes directly from the man page for `set`, see here: http://askubuntu.com/questions/346900/what-does-set-e-do

scottleibrand:
I just merged this to a branch on one of my rigs and ran it there without issues.  Want to re-target the PR to the oref0 dev branch, so we can test it further there?

ijustlovemath:
I would, but I'm not sure how to do that from mobile. Gimme a bit to figure that out.

scottleibrand:
Hard part is finding this page: https://github.com/openaps/oref0/compare/dev...ijustlovemath:patch-1
;-)

scottleibrand:
Merged to dev via https://github.com/openaps/oref0/pull/309

PR #305, opened 2016-12-31
improvements to dexusb

PieterGit:
most important one to add 'python -u' for unbuffered output. otherwise
the logging will not be added when started from crontab for some reason


bin/oref0-setup.sh
```diff
@@ -549,7 +549,7 @@ elif [[ ${CGM,,} =~ "xdrip" ]]; then
     (crontab -l; crontab -l | grep -q "cd $directory && ps aux | grep -v grep | grep -q 'openaps monitor-xdrip'" || echo "* * * * * cd $directory && ps aux | grep -v grep | grep -q 'openaps monitor-xdrip' || ( date; openaps monitor-xdrip) | tee -a /var/log/openaps/xdrip-loop.log; cp -up $directory/xdrip/glucose.json $directory/monitor/glucose.json") | crontab -
     (crontab -l; crontab -l | grep -q "xDripAPS.py" || echo "@reboot python $HOME/.xDripAPS/xDripAPS.py") | crontab -
 elif [[ $ENABLE =~ dexusb ]]; then
-    (crontab -l; crontab -l | grep -q "@reboot	/usr/bin/python" || echo "@reboot	/usr/bin/python /usr/local/bin/oref0-dexusb-cgm-loop.py >> /var/log/openaps/cgm-dexusb-loop.log 2>&1" ) | crontab -
+    (crontab -l; crontab -l | grep -q "@reboot .*dexusb-cgm" || echo "@reboot /usr/bin/python -u /usr/local/bin/oref0-dexusb-cgm-loop >> /var/log/openaps/cgm-dexusb-loop.log 2>&1" ) | crontab -
```
scottleibrand: What does this .* syntax do? I've never seen that...

PieterGit: `.` means any character, `*` means zero or more 
so it's a way to find out if the line containes @reboot and dexusb-cgm
the script should only hit on @reboot lines with dexusb-cgm on it

PR #309, opened 2017-01-02
Exit scripts when variables under or functions fail

ijustlovemath:


scottleibrand:
Not only does this break a bunch of things in hard-to-detect ways, it doesn't even help with the intended use case:

Without `set -eu`:
```
root@edison-eb ~/myopenaps # ns-upload
Input file entries.json does not exist.
```
With `set -eu`:
```
root@edison-eb ~/myopenaps # ns-upload
/usr/local/bin/ns-upload: line 14: 5: unbound variable
```

I'm going to revert this change in its entirety rather than waste any more time tracking down and fixing all the things it broke.

Issue #261, opened 2016-11-26
Iterative autotuning of basals and ratios


scottleibrand:
One possibly easier way to implement https://github.com/openaps/oref0/issues/99 in oref0 would be to take an incremental approach to iteratively adjusting basal schedules and ratios.

To keep track of required adjustments to the pump's programmed basal schedule, we could create a long-lived autotune.json file that contains an entry for each hour of the day for which we've identified an adjustment.  For ISF and CSF, we could similarly track any required adjustments in that same file.  

For time periods without COB, and where basal insulin activity dominates, we could look at net deviations for each hour, and calculate how much less or additional basal insulin would have been required to eliminate the deviations.  We could then add an adjustment factor/multiplier to the autotune.json to adjust basal by a fraction of that, perhaps 10%.  That could be split across a few hours' basals: perhaps 5% of the required adjustment would go in the 2h-prior slot, and 2.5% each could go in the 1h-prior and 3h-prior ones.  The other 90% of adjustment would not be made at all unless subsequent days' outcomes justified further adjustments, so as to dampen oscillations and cause the system to react gradually to observed insulin needs.

After mealtimes, we could observe whether AMA's COB estimate drops to zero before or after post-prandial deviations drop to zero.  If the COB estimate hits zero first, that indicates that the (perceived carb) sensitivity factor is too low.  As with basals, we could adjust the CSF ratio (calculated from the pump's ISF and IC ratio) multiplier in autotune.json by a fraction (perhaps 10%) of the adjustment that would've been required to get COB to decay to zero at the time we saw deviations drop to ~0.  For meals where deviations drop to zero while COB is still positive, we'd want to subtract out any net positive deviations after the initial negative deviation, so we can account for any carbs whose absorption was delayed by post-meal activity etc.

For any post-correction periods and post-meal periods with significant insulin activity after COB and deviations have dropped to zero, we can calculate deviations for the period where bolus and high-temp insulin dominate basal insulin, and calculate what adjustments to ISF would have been necessary to bring deviations for that period to zero.  As with basals and CSF, we can gradually adjust the ISF multiplier by a fraction (perhaps 10%) of the observed deviations.

scottleibrand:
The existing lib/determine-basal/cob-autosens.js doesn't have all the info it needs to determine whether a meal is over yet: it just calculates carbsAbsorbed, and then lib/meal/total.js uses that to actually calculate COB.

It would probably make sense to create a new high-level autotune-prep.js, which would:
[...]

PR #301, opened 2016-12-29
New tool to sync profile data from OpenAPS to Nightscout

sulkaharo:
DO NOT MERGE YET! Someone who uses mg/dL needs to verify this works.

Note this expands the data contained in profile.json to contain a full pump profile and rounds some of the contained numbers to get rid of excessive fractions.

Issue #99, opened 2016-04-06
Basal tuning and ratio estimation


scottleibrand:
Per the discussion in #58 and many preceding and subsequent discussions in various fora, it would be very useful to implement an algorithm to provide suggestions for tuning a user's current basal schedule, insulin sensitivity factor (ISF), insulin to carb (IC) ratio, and duration of insulin action (DIA).

To do so, we could use the OpenAPS automatic sensitivity detection algorithm to calculate, for each 5m data point, how much BG deltas deviated from what would’ve been predicted based on net insulin activity (relative to current schedule basals) and the user's current ratios.

For each carb entry, we can determine the sum of subsequent positive deviations (above a threshold used to detect when carbs are absorbing, perhaps 0.5 mg/dL/min).  This can then be directly compared to the carb estimate to get a carb sensitivity factor (CSF, units of mg/dL per g), which can be combined with estimated ISF (units of mg/dL per U) to get IC ratio (units of g per U).  (Note: while it would appear that this relies on correct carb counting, it actually does not.  If the user's carb counts are biased high or low, the resulting ratios will essentially correct for that, resulting in what you might call perceived-carb ratios.)

To calculate ISF/DIA and tune basal schedules, we need to first exclude any data between carb announcement and 15-30 minute after the last carb absorption is detected.  We can then bucket the remaining data points according to whether their insulin activity is dominated by boluses / high-temps or by basal insulin.  The former can be used to calculate ISF, and the latter to tune basal schedules.  We can initially assume that ISF is constant, and then come back and fine-tune based on time of day or day-to-day variation, assuming we have sufficient data, after we’ve tuned basal schedules and DIA.

To calculate ISF, we need to take the subset of data with no carb absorption whose insulin activity is bolus / high-temp dominated as described above, and perform a parameter estimation function to solve for ISF.  One simple (but maybe not ideal) method to do so would be to perform a simple goal seek algorithm, recalculating the deviations for each new ISF to find the ISF that converges on a median or mean deviation (for this subset of data) of zero.

To determine the best DIA, we want to select the one that minimizes the sum (or perhaps the RMS?) of all the deviations, in order to choose the insulin curve that best fits the actual observed insulin activity profile over time.  This parameter could be estimated from the same subset of data used to estimate ISF, using simple goal seek or more sophisticated parameter estimation methods.

[...]

PR #313, opened 2017-01-03
Iterative autotuning of basals and ratios

scottleibrand:
Implements #261 

If autotune is selected as an advanced feature during oref0-setup, a cron job runs every night that downloads the previous day's entries and treatments data from NS, performs the autotune calculations, and updates the profile.json accordingly (subject to not being more than 20% off of the pumpprofile.json).  The updated profile.json is used to calculate IOB and by determine-basal.

Additionally, thanks to @jyaw and @pietergit, oref0-ns-autotune can easily be run manually to use retrospective data from NS to more quickly tune basals and ratios.  That output is intended for human review (i.e. via xlsx export), and is not used automatically by oref0.

Thanks also to @danamlewis for being awesome as always, and contributing so much in so many ways, on design, clarifying the code, making pumpprofile mandatory, code reviews, and so much more.  And special thanks to @sulkaharo and everyone who contributed code and ideas in #261, and in Gitter, and in Slack.


lib/autotune/index.js
```diff
@@ -0,0 +1,292 @@
...
+    p50bgi = percentile(bgis, 0.50);
+    p50ratios = Math.round( percentile(ratios, 0.50) * 1000)/1000;
+
+    // calculate what adjustments to ISF would have been necessary to bring median deviation to zero
+    fullNewISF = isf * p50ratios;
+    fullNewISF = Math.round( fullNewISF * 1000 ) / 1000;
+    // and apply 10% of that adjustment
+    var newISF = ( 0.9 * isf ) + ( 0.1 * fullNewISF );
+    if (typeof(pumpISF) !== 'undefined') {
+        var maxISF = pumpISF * 1.2;
```
PieterGit: I think 20% is quite a lot of change for an ISF to begin with. From safety perspective I would make this configurable and default it to 5% or 10%. For example we use `max_iob=0` as default value.

package.json
```diff
@@ -50,6 +53,7 @@
     "oref0-meal": "./bin/oref0-meal.js",
     "oref0-mint-max-iob": "./bin/oref0-mint-max-iob.sh",
     "oref0-normalize-temps": "./bin/oref0-normalize-temps.js",
+    "oref0-ns-autotune": "./bin/oref0-ns-autotune.sh",
```
PieterGit: Why not cal it `oref0-autotune`? The autotune will tune `oref0`. It will not tune anything to Nightscout.

bin/oref0-ns-autotune.sh
```diff
@@ -0,0 +1,173 @@
...
+for i in "${date_list[@]}"
+do 
+  curl "$NIGHTSCOUT_HOST/api/v1/entries/sgv.json?find\[date\]\[\$gte\]=`(date -d $i +%s | tr -d '\n'; echo 000)`&find\[date\]\[\$lte\]=`(date --date="$i +1 days" +%s | tr -d '\n'; echo 000)`&count=1000" > ns-entries.$i.json
+done
+
+echo "Running $NUMBER_OF_RUNS runs from $START_DATE to $END_DATE"
+sleep 2
+
+# Do iterative runs over date range, save autotune.json (prepped data) and input/output 
+# profile.json
```
PieterGit: Shouldn't we use a different file name as `profile.json`, e.g. `profile.autotune.js` and 
`cp profile.json profile.autotune.js` at the start of autotune and `cp profile.autotune.js profile.json` at then end of autotune, if the users want's to automatically tune profiles. This will be a simple way of implementing https://github.com/openaps/oref0/issues/99 

I would advise the default is only to generate  autotune files, and not to use the `profile.autotune.json` by default

bin/oref0-ns-autotune.sh
```diff
@@ -0,0 +1,173 @@
...
+    # output autotuned profile or what will be used as <autotune/autotune.json> in the next iteration
+    echo "~/src/oref0/bin/oref0-autotune.js autotune.$run_number.$i.json profile.json profile.pump.json > newprofile.$run_number.$i.json"
+    ~/src/oref0/bin/oref0-autotune.js autotune.$run_number.$i.json profile.json profile.pump.json > newprofile.$run_number.$i.json
+    
+    # Copy tuned profile produced by autotune to profile.json for use with next day of data
+    cp newprofile.$run_number.$i.json profile.json
+
+  done # End Date Range Iteration
+done # End Number of Runs Loop
+
```
PieterGit: We could do something like:
```
date="$(date --date="$date + 1 days" +%Y-%m-%d-%H-%M)"; 
cp profile.autotune.json profile.autotune.$date.json 
```
here to save the generated autotune profile on disk for later reference/backup.
Even better will be to store the profile in Nightscout, see https://github.com/openaps/oref0/pull/301

lib/autotune/index.js
danamlewis: Most people have been using 20% on ISF via autosens already, and we haven't had any issues with that - and watching the results of the tune, I feel comfortable continuing to have 20% be the safety cap for either one-off tuning or for use in the loop, given that it has the other safety parameters in concert with it, too. But yes, in the future we also plan to make it a ratio configurable in preferences.json, or to use the same ratio safety multipliers already there for autosens.

cjo20: I think it'd be interesting to have a mode that showed what the autotune would set the ratios to without any percentage restrictions - both for the shape of the graph and a visualisation of how far off it thinks your ratios are in total; it may make sense for this mode to only output in xlsx (or similar) to prevent it being accidentally used in a loop.

scottleibrand: Yeah, we should probably change the 20% limit from being hard-coded to using the autosens min and max from preferences.json.  Anyone want to take that on?  If not, I might be able to get to that later today.

@cjo20 you can auto-tune more than 20% from the pumpprofile by providing the autotuned profile.json as the pumpprofile argument.  That way it will only be limited to changing things 20% per day, and can converge as far from the initial values as you want.

package.json
scottleibrand: The oref0-autotune name is already taken, by bin/oref0-autotune.js.

I called it oref0-ns-autotune because it's tuning based on data downloaded from NS, not based on local pumphistory.json and glucose.json.  Any other suggestions for clearer names for oref0-ns-autotune, oref0-autotune-prep, and oref0-autotune?

bin/oref0-ns-autotune.sh
scottleibrand: By default, autotune will not be activated, but can be run manually.  It will only be turned on for automatic tuning if they enable autotune as an advanced feature in oref0-setup.  I don't see the need for an intermediate state where tuning happens nightly but doesn't actually do anything: if the autotune output is not copied to the profile.json for use in closed looping, that is essentially the same as running autotune retrospectively, which you can do manually any time you like.

bin/oref0-ns-autotune.sh
scottleibrand: That is a good idea.  Care to PR it?

package.json
PieterGit: i would rename `bin/oref0-autotune.js` to  `bin/oref0-autotune-core.js` and use `bin/oref0-autotune` for the cli program. That NS is used as a source confuses me, because you want the name to be clear about what it is auto tuning. And that is oref0.

scottleibrand: Ok, I'd be fine with. Want to PR a fix for all the names and dependencies?

scottleibrand: I went ahead and did this one.

bin/oref0-ns-autotune.sh
scottleibrand: Looking at what oref0-autotune does, it looks like all of this data is saved to disk already in the autotune directory.

scottleibrand:
Ok, I think this might be about ready for merge now.  Anything else?

scottleibrand:
If anyone can give me a :+1: on https://github.com/openaps/oref0/pull/319, I'll go ahead and merge it to autotune, and then I think autotune can be considered feature complete and ready for one last quick round of testing before merging this PR to dev.

scottleibrand:
I'll probably go ahead and merge this tomorrow (Sunday) either way, unless anyone wants me to wait.

scottleibrand:
Merged #319 - I think autotune is ready for one last quick round of testing before merging this PR to dev, and then we can start working on enhancements and more automated testing.

scottleibrand:
Installed this on several rigs that never had autotune, and verified that:
 - they looped successfully until the first autotune run
 - after the first autotune run, they looped successfully with the newly tuned ratios
 - the nightly autotune worked unattended, and they kept looping thereafter

So if no objections, I'm going to merge this to dev today.

jrussell88:
I'm running `oref0-autotune` on my laptop to tune my basal. 

Making daily basal adjustments, I run it against one day's data at a time, update profile.json with revised basal profile, and copy to pumpprofile.json and autotune.json. As more BG data builds up with the new basal profile, the current recommendations also build up. However it should give more accurate results if I could run against my changing basal profile over a longer time, rather than being unduly influenced by a single day.

Is it possible to manually run autotune against changing basal data?

scottleibrand:
Autotune is designed to run against one day's data at a time and incrementally update basals and ratios accordingly. Running it repeatedly against the same data will cause it to adjust things faster, but may cause it to overfit that data and over-adjust things.

jrussell88:
> Running it repeatedly against the same data
I don't think that's what I suggested. Autotune can be run against multiple days, with the same basal profile and combines the results. That has the advantage of using multiple days, and minimising the effects of one-off 'random' perturbations.

The same benefits could apply to a changing basal profile, if autotune could be run against a multi-day basal profile. 

Partly this depends what methods are used to estimate the parameters - maybe I need to go and read some code.

scottleibrand:
The assumption is that there is one relatively unchanging basal profile, and autotune is trying to find it. Each time you run autotune, you want to run it with your best estimate of the basal profile and ratios, so it can further tune those closer to the truth. I'm not sure what you mean by a multi-day basal profile.

jrussell88:
By multi-day basal profile I mean the basal profile over a period longer than one day. The assumption that the basal profile is static from day to day is contradicted by the adjustments which autotune is making. 

I understand that to get the benefit of longer data periods, the daily basal profile must be constant.

scottleibrand:
Autotune's adjustments don't imply the basal profile changes from day to day, just that it was wrong originally. We use autotune to converge on an average daily basal profile, and then handle day to day variations with autosens. You can read up on the differences in the OpenAPS docs.

PR #314, opened 2017-01-03
Add stdout log file and simple autotune recommendation report to oref0-ns-autotune.sh

jyaw:
This PR is intended to add 2 additional features to the oref0-ns-autotune.sh script.

**1. Terminal Logging** - Can be turned off using --log=false, defaulted to true. The autotune.<date/time stamp>.log file will be created in the OpenAPS autotune directory. It contains all the terminal output from this run of autotune. I've found the terminal output indispensable in understanding the output.
**2. Simple Recommendation Report** - (autotune_recommendations.log) Defaulted to enabled. Simple tabular output with Parameter, Current and Autotune Recommendation presented. Created in order to display the product of autotune in a quick old vs new view. The shell script that creates the report is a little slow (5-10 sec?), but I didn't think that was a big deal since the manual autotune can take quite some time. The updated oref0-ns-autotune.sh cats this report to the screen.


bin/oref0-ns-autotune.sh
```diff
@@ -112,6 +119,12 @@ cp settings/autotune.json autotune/profile.json && cat autotune/profile.json | j
 cd autotune
 # TODO: Need to think through what to remove in the autotune folder...
 
+# Turn on stdout logging, if enabled (default to true)
```
scottleibrand: Is it stdout we want, or stderr?

bin/oref0-ns-autotune.sh
```diff
@@ -83,7 +90,7 @@ esac
 done
 
 if [[ -z "$DIR" || -z "$NIGHTSCOUT_HOST" ]]; then
-    echo "Usage: oref0-ns-autotune <--dir=openaps_directory> <--ns-host=https://mynightscout.azurewebsites.net> [--start-date=YYYY-MM-DD] [--end-date=YYYY-MM-DD] [--runs=number_of_runs] [--xlsx=autotune.xlsx]"
+    echo "Usage: oref0-ns-autotune <--dir=openaps_directory> <--ns-host=https://mynightscout.azurewebsites.net> [--start-date=YYYY-MM-DD] [--end-date=YYYY-MM-DD] [--runs=number_of_runs] [--xlsx=autotune.xlsx] [--log=true|false(true)]"
```
scottleibrand: I would do this as something like `--log=[true]|false`

scottleibrand:
Looks good.  If you want to make any edits tonight great; otherwise I'll probably go ahead and merge and we can do any clean-up afterward.

bin/oref0-ns-autotune.sh
jyaw: I had assumed that stdout was where all of the messaging was coming from in the prep/autotune process. All that output is what's in the log files. We can change the terminology here if needed, no problem.

jyaw:
I don't mind cleaning up tomorrow evening if you haven't already. Thanks for the feedback!

scottleibrand:
I think that is all console.error, which goes to stderr. The stdout of the .js commands is just the json output.

jyaw:
@scottleibrand always learning about this stuff... did some quick googles and the initial search I did when updating the script to find the `exec` command syntax included `&>` and said it was for stdout only. This is not correct, `>` is stdout, `&>` is stdout and stderr. This jives with your explanation. Thanks. Will leave as is then, unless you have other objections.

scottleibrand:
Ok, cool.  So sounds like we just need to change the comments to reference stdout instead of stderr then.

PR #315, opened 2017-01-06
Process high and low events as type sgv

tmecklem:
This change causes the oref0 and nightscout scripts to treat GlucoseSensorDataLow (<40) and GlucoseSensorDataHigh (>400) events as type=sgv events. As before, oref0 continues to treat regular GlucoseSensorData events as type=sgv.

There is a relevant PR in decocare that adds the decoding of the newly discovered high and low events: https://github.com/openaps/decocare/pull/16.

### DIFF
--- a/bin/mm-format-ns-glucose.sh
+++ b/bin/mm-format-ns-glucose.sh
@@ -32,7 +32,7 @@ cat $HISTORY | \
   json -E "this.medtronic = this._type;" | \
   json -E "this.dateString = this.dateString ? this.dateString : (this.date + '$(date +%z)')" | \
   json -E "this.date = new Date(this.dateString).getTime();" | \
-  json -E "this.type = (this.name == 'GlucoseSensorData') ? 'sgv' : 'pumpdata'" | \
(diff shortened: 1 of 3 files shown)


scottleibrand:
Too me a minute to figure out that `this.name.indexOf('GlucoseSensorData') > -1` is a substring search to match on either GlucoseSensorData, GlucoseSensorDataLow, or GlucoseSensorDataHigh, so documenting that here for anyone else it confuses.

scottleibrand:
This isn't dependent on the decocare changes, right?  It can be merged right away and will be backwards compatible?

tmecklem:
Is there a way for me to use https://lodash.com/docs/4.17.4#startsWith in the scripts?

tmecklem:
Yeah, I made this independent of the decocare changes so it could be merged any time.

scottleibrand:
If you want to use lodash I think you can do it like @sulkaharo just did at https://github.com/openaps/oref0/pull/301/commits/ee1ecd6949d15b4b481f12b22c4fd16066827b00

scottleibrand:
LMK if you'd prefer to refactor with lodash, but I'm happier merging it as-is if code clarity is the only reason.

tmecklem:
It would be for better readability. I'm okay with it as it is. 👍

PR #316, opened 2017-01-06
profile.max_basal fix

MilosKozak:


sulkaharo:
Good catch!

MilosKozak:
No idea. I'm studying code and I thought it was only typo

scottleibrand:
It was definitely a typo, but we may have ended up relying on it anyway. I can't think of any reason to allow basal rates below 0.1U/hr at any time, so probably safest to check both...

sulkaharo:
At which point does this check happen? Our basal profile has rates under 0.1, wondering how it's worked so far.

sulkaharo:
Can you include an unit test that'd catch this?

cjo20:
Why would you want to disallow basal rates under 0.1u/hr?

MilosKozak:
pump usualy doesn't allow lower basal rate.

cjo20:
I can set mine all the way down to 0, at 0.025u/he increments.

MilosKozak:
i'm pretty sure MM pumps are not able to deliver only 0.025U/h

scottleibrand:
We can't support zero: that means OpenAPS can't do anything. (The maxSafeBasal multipliers will always be zero.) If you have a use case for a minimum for basal of 0.025 instead of 0.1, I'd be fine with that.

scottleibrand:
I believe 522 can do 0.05U/hr and 523 can do 0.025U/hr (albeit with not a lot of precision).

MilosKozak:
i think DanaR is the best and lowest possible basal rate is 0.04 or 0.06

cjo20:
img_1650

I don't think there's a good case for not lowering it to at least 0.025

MilosKozak:
well we're still talking about max not a single value. i believe if max is below 0.1 we can treat the profile as invalid

scottleibrand:
Yeah, I think we want to leave the typofixed max daily limit at 0.1. I think we're now discussing (agreeing?) to set the current basal limit to 0.025.

cjo20:
A max of 0.1u/hr seems pretty arbitrary. Why would you choose 0.1u/hr over, say, 0.125u/hr, or 0.175u/hr? What is the benefit of requiring that an openaps user has at least 1 basal rate a day over 0.1u/hr?

scottleibrand:
The reason for the check was to catch decocare bugs or bad pump comms that resulting in impossibly low readings. But if anyone really really needs basal profiles that low, they can change the code. It's working for 150+ people, including a 1-year-old.

cjo20:
If the only invalid value is 0, the check should be 
```
if (profile.max_basal == 0) {
      console.error("max_basal of 0 is not supported");
      return -1;
}
```

MilosKozak:
just looking i've in AndoridAPS the same limit i choosed independent to you. i feel that try to detect user's configuration mistakes is pretty important ...

scottleibrand:
0 is not the only invalid value. Anything the pump doesn't support is invalid. And our algorithm is not tested or supported for basals below 0.1U/hr. We're already talking about relaxing that check by 4x: no need to do more than that when there have been zero users who've needed it.

scottleibrand:
To do this right we'd probably need to error if any scheduled basal is less than 0.025, not just the current one. I think I'll merge this as is and we can do that as a separate thing.

PR #323, opened 2017-01-08
fix ww pump and dexusb with small changes

PieterGit:
with a reinstall i noticed two bugs that caused ww pump not to work and also a problem with the dexusb script which i thought was fixed in git, but wasn't

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -452,7 +452,7 @@ else
     openaps device add pump mmeowlink subg_rfspy $ttyport $serial $radio_locale || die "Can't add pump"
     openaps alias add wait-for-silence '! bash -c "(mmeowlink-any-pump-comms.py --port '$ttyport' --wait-for 1 | grep -q comms && echo -n Radio ok, || openaps mmtune) && echo -n \" Listening: \"; for i in $(seq 1 100); do echo -n .; mmeowlink-any-pump-comms.py --port '$ttyport' --wait-for 30 2>/dev/null | egrep -v subg | egrep No && break; done"'
     openaps alias add wait-for-long-silence '! bash -c "echo -n \"Listening: \"; for i in $(seq 1 200); do echo -n .; mmeowlink-any-pump-comms.py --port '$ttyport' --wait-for 45 2>/dev/null | egrep -v subg | egrep No && break; done"'
-    if [[ ${radio_locale,,} =~ "WW" ]]; then
+    if [[ "$radio_locale" =~ "WW" ]]; then
       # add subg-ww-radio-parameters script to mmtune for WW pump. See https://github.com/oskarpearson/mmeowlink/issues/51 or https://github.com/oskarpearson/mmeowlink/wiki/Non-USA-pump-settings for details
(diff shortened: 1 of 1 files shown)


bin/oref0-setup.sh
```diff
@@ -452,7 +452,7 @@ else
     openaps device add pump mmeowlink subg_rfspy $ttyport $serial $radio_locale || die "Can't add pump"
     openaps alias add wait-for-silence '! bash -c "(mmeowlink-any-pump-comms.py --port '$ttyport' --wait-for 1 | grep -q comms && echo -n Radio ok, || openaps mmtune) && echo -n \" Listening: \"; for i in $(seq 1 100); do echo -n .; mmeowlink-any-pump-comms.py --port '$ttyport' --wait-for 30 2>/dev/null | egrep -v subg | egrep No && break; done"'
     openaps alias add wait-for-long-silence '! bash -c "echo -n \"Listening: \"; for i in $(seq 1 200); do echo -n .; mmeowlink-any-pump-comms.py --port '$ttyport' --wait-for 45 2>/dev/null | egrep -v subg | egrep No && break; done"'
-    if [[ ${radio_locale,,} =~ "WW" ]]; then
+    if [[ "$radio_locale" =~ "WW" ]]; then
```
scottleibrand: I think the proper fix here is to change WW to ww - the ${radio_locale,,} syntax renders it in lowercase.

scottleibrand: That will allow us to honor whatever the user puts in, regardless of what case they use.  That's what we're doing for all the other user-entered variables.

PieterGit: there is also some code for upcasing it in the script 
```
  # Force uppercase, just in case the user entered ww
      radio_locale=${radio_locale^^}
```
but only for user input. Should we always upcase or aways lowercase.  What do you suggest?

scottleibrand: The convention I've been using thus far is to lowercase everything for comparison:

```
scott@scott-mbp ~/oref0 $ grep ,, bin/oref0-setup.sh
if ! [[ ${CGM,,} =~ "g4" || ${CGM,,} =~ "g5" || ${CGM,,} =~ "mdt" || ${CGM,,} =~ "shareble" || ${CGM,,} =~ "xdrip" ]]; then
    if [[ ${CGM,,} =~ "shareble" ]]; then
if [[ ${CGM,,} =~ "shareble" ]]; then
if [[ ${CGM,,} =~ "shareble" ]]; then
if [[ ${CGM,,} =~ "xdrip" ]]; then
if [[ ! -z "$BT_MAC" || ${CGM,,} =~ "shareble" ]]; then
if [[ ${CGM,,} =~ "g5" ]]; then
elif [[ ${CGM,,} =~ "shareble" ]]; then
    if [[ ${radio_locale,,} =~ "WW" ]]; then
if [[ ${CGM,,} =~ "mdt" ]]; then
if [[ ${CGM,,} =~ "xdrip" ]]; then
    if [[ ${CGM,,} =~ "shareble" ]]; then
    elif [[ ${CGM,,} =~ "xdrip" ]]; then
    elif ! [[ ${CGM,,} =~ "mdt" ]]; then # use nightscout for cgm
if [[ ${CGM,,} =~ "shareble" ]]; then
```

PieterGit: ok, will fix it tommorow. bed time for now

PR #328, opened 2017-01-10
Pancred addon

Bender1061:
add on @mddub 's Pancreabble to the oref0-setup.

the peb-urchin-status.sh is not by any means perfect. but it does work.  Easy to turn on and off the different types of statues you would like by changing the bools in pancreoptions.json
Ideas on what statues and notifacations to to add to this would be most helpful.
Almost need a way to know if a value has already been sent to the watch, to prevent duplicate notifications. This would also be an easy way to warn you that your battery is low. But I was unable to make the script write to a json file. 

follow https://github.com/mddub/pancreabble for instructions on Pairing Pebble to rig.
I have not tested with a Pi. 

Added the oref0-bluetoothup to simplify turning on the bluetooth when BlueZ is installed. 

Pebble Mac address can be sound in settings of Pebble. 

there is a bug of when an update is sent to the watch, an error message will show up on terminal if you are hooked up to SSH with a usb cable. this does not happen when using network. So it's annoying but you can still work around it.

### DIFF
--- a/bin/oref0-bluetoothup.sh
+++ b/bin/oref0-bluetoothup.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+
+if ! ( hciconfig -a | grep -q "PSCAN" ) ; then
+   sudo killall bluetoothd
+   sudo /usr/local/bin/bluetoothd --experimental &
+fi
+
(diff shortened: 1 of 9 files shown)


Bender1061:
Fixed to PR to Dev

PR #329, opened 2017-01-10
Pancre add

Bender1061:
add on @mddub 's Pancreabble to the oref0-setup.

the peb-urchin-status.sh is not by any means perfect. but it does work. Easy to turn on and off the different types of statues you would like by changing the bools in pancreoptions.json
Ideas on what statues and notifacations to to add to this would be most helpful.
Almost need a way to know if a value has already been sent to the watch, to prevent duplicate notifications. This would also be an easy way to warn you that your battery is low. But I was unable to make the script write to a json file.

follow https://github.com/mddub/pancreabble for instructions on Pairing Pebble to rig.
I have not tested with a Pi.

Added the oref0-bluetoothup to simplify turning on the bluetooth when BlueZ is installed.

Pebble Mac address can be sound in settings of Pebble.

there is a bug of when an update is sent to the watch, an error message will show up on terminal if you are hooked up to SSH with a usb cable. this does not happen when using network. So it's annoying but you can still work around it.

### DIFF
--- a/bin/oref0-bluetoothup.sh
+++ b/bin/oref0-bluetoothup.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+if ( ifconfig wlan0 | grep -q "inet addr" ) && ( ifconfig bnep0 | grep -q "inet addr" ); then
+   sudo killall bluetoothd
+fi
 
 if ! ( hciconfig -a | grep -q "PSCAN" ) ; then
    sudo killall bluetoothd
@@ -8,6 +11,7 @@ fi
 
 if ( hciconfig -a | grep -q "DOWN" ) ; then
    sudo hciconfig hci0 up
+   sudo /usr/local/bin/bluetoothd --experimental &
 fi
 
 if !( hciconfig -a | grep -q $HOSTNAME ) ; then
--- a/bin/oref0-bluetoothup.sh
+++ b/bin/oref0-bluetoothup.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+
+if ! ( hciconfig -a | grep -q "PSCAN" ) ; then
+   sudo killall bluetoothd
+   sudo /usr/local/bin/bluetoothd --experimental &
+fi
+
+if ( hciconfig -a | grep -q "DOWN" ) ; then
+   sudo hciconfig hci0 up
+fi
+
+if !( hciconfig -a | grep -q $HOSTNAME ) ; then
+   sudo hciconfig hci0 name $HOSTNAME
+fi
--- a/bin/oref0-bluetoothup.sh
+++ b/bin/oref0-bluetoothup.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
+MAC=$1
 
 if ( ifconfig wlan0 | grep -q "inet addr" ) && ( ifconfig bnep0 | grep -q "inet addr" ); then
-   sudo killall bluetoothd
+   sudo bt-pan client $MAC -d
 fi
 
 if ! ( hciconfig -a | grep -q "PSCAN" ) ; then
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -189,7 +189,7 @@ if [[ -z "$DIR" || -z "$serial" ]]; then
        BT_PEB=$REPLY
        echo "Ok, $BT_PEB it is."
     else
-       echo echo Ok, no Pancreabble for you.
+       echo Ok, no Pancreabble for you.
     fi
     read -p "Do you need any advanced features? y/[N] " -r
     if [[ $REPLY =~ ^[Yy]$ ]]; then
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -583,14 +583,13 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
         (crontab -l; crontab -l | grep -q "reset_spi_serial.py" || echo "@reboot reset_spi_serial.py") | crontab -
     fi
     (crontab -l; crontab -l | grep -q "cd $directory && ( ps aux | grep -v grep | grep -q 'openaps pump-loop'" || echo "* * * * * cd $directory && ( ps aux | grep -v grep | grep -q 'openaps pump-loop' || openaps pump-loop ) 2>&1 | tee -a /var/log/openaps/pump-loop.log") | crontab -
-    crontab -l
     if [[ ! -z "$BT_PEB" ]]; then
        (crontab -l; crontab -l | grep -q "cd $directory && ( ps aux | grep -v grep | grep -q 'peb-urchin-status $BT_PEB && openaps urchin-loop'" || echo "* * * * * cd $directory && ( ps aux | grep -v grep | grep -q 'peb-urchin-status $BT_PEB && openaps urchin-loop' || peb-urchin-status $BT_PEB && openaps urchin-loop ) 2>&1 | tee -a /var/log/openaps/urchin-loop.log") | crontab -
     fi
-    crontab -l
     if [[ ! -z "$BT_PEB" || ! -z "$BT_MAC" ]]; then
-       (crontab -l; crontab -l | grep -q "oref0-bluetoothup" || echo '* * * * * ps aux | grep -v grep | grep -q "oref0-bluetoothup" || oref0-bluetoothup >> /var/log/openaps/network.log' ) | crontab -
+       (crontab -l; crontab -l | grep -q "oref0-bluetoothup  $BT_MAC" || echo '* * * * * ps aux | grep -v grep | grep -q "oref0-bluetoothup  '$BT_MAC' || oref0-bluetoothup  '$BT_MAC' >> /var/log/openaps/network.log' ) | crontab -
     fi
+    crontab -l
 fi
 
 if [[ ${CGM,,} =~ "shareble" ]]; then
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -81,6 +81,10 @@ case $i in
     BT_MAC="${i#*=}"
     shift # past argument=value
     ;;
+    -p=*|--btpeb=*)
+    BT_PEB="${i#*=}"
+    shift # past argument=value
+    ;;
     *)
             # unknown option
     echo "Option ${i#*=} unknown"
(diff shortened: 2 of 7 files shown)


Bender1061:
@scottleibrand I closed previous pull request, I'm not seeing the change to pump-loop we had been talking about.

bin/oref0-bluetoothup.sh
```diff
@@ -0,0 +1,15 @@
...
+if ! ( hciconfig -a | grep -q "PSCAN" ) ; then
+   sudo killall bluetoothd
+   sudo /usr/local/bin/bluetoothd --experimental &
+fi
+
+if ( hciconfig -a | grep -q "DOWN" ) ; then
+   sudo hciconfig hci0 up
+fi
+
+if !( hciconfig -a | grep -q $HOSTNAME ) ; then
```
scottleibrand: Does this get passed as an environment variable, or do we need to pass it as a script argument?

bin/oref0-setup.sh
```diff
@@ -180,6 +184,16 @@ if [[ -z "$DIR" || -z "$serial" ]]; then
           echo "Ok, $BT_MAC it is."
        fi
     fi
+    if [[ ! -z $BT_PEB ]]; then
+       read -p "For Pancreabble enter Pebble mac id (i.e. AA:BB:CC:DD:EE:FF) hit enter to skip " -r
+       BT_PEB=$REPLY
+       echo "Ok, $BT_MAC it is."
```
scottleibrand: Is this extra?

bin/oref0-bluetoothup.sh
Bender1061: it's an environmental variable, accessible from anywhere.

bin/oref0-setup.sh
```diff
@@ -187,12 +187,9 @@ if [[ -z "$DIR" || -z "$serial" ]]; then
...
        BT_PEB=$REPLY
-       echo "Ok, $BT_MAC it is."
-       if [[ -z $BT_PEB ]]; then
-          echo Ok, no Pancreabble for you.
-          else
-          echo "Ok, $BT_PEB it is."
-       fi
+       echo "Ok, $BT_PEB it is."
+    else
+       echo echo Ok, no Pancreabble for you.
```
scottleibrand: Echo echo... 😉

bin/oref0-setup.sh
```diff
@@ -180,6 +184,13 @@ if [[ -z "$DIR" || -z "$serial" ]]; then
           echo "Ok, $BT_MAC it is."
        fi
     fi
+    if [[ ! -z $BT_PEB ]]; then
+       read -p "For Pancreabble enter Pebble mac id (i.e. AA:BB:CC:DD:EE:FF) hit enter to skip " -r
+       BT_PEB=$REPLY
+       echo "Ok, $BT_PEB it is."
+    else
+       echo echo Ok, no Pancreabble for you.
```
scottleibrand: Echo echo...😉

bin/oref0-bluetoothup.sh
Bender1061: note the hostname is not actually needed. but it does help when you have multiple rigs paired to your phone, so you can tell what rigs are paired. it also helps with the Pebble, as I found I can have multiple rigs paired with one pebble (I haven't tried to see what happens when both rigs are running at same time)

Bender1061: the name is actually changing the Alias name for the BlueZ, but in order for the phone or device to see that name you must do the power off and power on in the bluetoothctl. so it only helps with pairing, the phone does not really care what the name is, it only looks at the mac address after pairing. So your paired device name only helps you, not the phone.

bin/oref0-setup.sh
```diff
@@ -180,6 +184,13 @@ if [[ -z "$DIR" || -z "$serial" ]]; then
           echo "Ok, $BT_MAC it is."
        fi
     fi
+    if [[ ! -z $BT_PEB ]]; then
+       read -p "For Pancreabble enter Pebble mac id (i.e. AA:BB:CC:DD:EE:FF) hit enter to skip " -r
+       BT_PEB=$REPLY
+       echo "Ok, $BT_PEB it is."
+    else
+       echo Ok, no Pancreabble for you.
```
scottleibrand: Typically this would come after an interactive prompt for Pebble info. If you want it to be a hidden command-line option, perhaps this line isn't needed?

scottleibrand:
And that was supposed to be a comment, not a request changes. (Reviewing on mobile on a plane.)

bin/oref0-setup.sh
Bender1061: are you referring to the "no pancreabble for you part" or the whole thing to get the mac for the pebble?, cause yes the 'else' really is not needed.

scottleibrand: I'm just referring to the else clause. If you don't want to prompt about pancreabble unless someone passes the -p flag that is fine with me.

Bender1061: I'll get rid of the else. Probably could do that for the phone too.

scottleibrand:
Ready to merge?

Bender1061:
it's ready now

bin/oref0-bluetoothup.sh
```diff
@@ -0,0 +1,19 @@
...
+fi
+
+if ! ( hciconfig -a | grep -q "PSCAN" ) ; then
+   sudo killall bluetoothd
+   sudo /usr/local/bin/bluetoothd --experimental &
+fi
+
+if ( hciconfig -a | grep -q "DOWN" ) ; then
+   sudo hciconfig hci0 up
+   sudo /usr/local/bin/bluetoothd --experimental &
```
scottleibrand: Do we need to kill off bluetoothd first here? Or is it guaranteed not to be running if hciconfig -a says DOWN?

Bender1061: nope, cause at that point it's already dead.

Bender1061: from all the testing I did, the killall was the only thing that put the hciconfig down.

bin/oref0-setup.sh
```diff
@@ -586,6 +586,9 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
     crontab -l
     if [[ ! -z "$BT_PEB" ]]; then
        (crontab -l; crontab -l | grep -q "cd $directory && ( ps aux | grep -v grep | grep -q 'peb-urchin-status $BT_PEB && openaps urchin-loop'" || echo "* * * * * cd $directory && ( ps aux | grep -v grep | grep -q 'peb-urchin-status $BT_PEB && openaps urchin-loop' || peb-urchin-status $BT_PEB && openaps urchin-loop ) 2>&1 | tee -a /var/log/openaps/urchin-loop.log") | crontab -
+    fi
+    crontab -l
```
scottleibrand: There should only be one crontab -l at the end to show the end result.

bin/oref0-bluetoothup.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+if ( ifconfig wlan0 | grep -q "inet addr" ) && ( ifconfig bnep0 | grep -q "inet addr" ); then
+   sudo killall bluetoothd
```
scottleibrand: Do we really want to kill Bluetooth just because we got a wifi IP? Shouldn't we test Internet connectivity over it, and only disconnect the bt-pan (not kill Bluetooth) if wifi has access to the Internet?

Maybe this part should be a new PR, as it's not pancreabble related and might require some more discussion about desired behavior.

bin/oref0-bluetoothup.sh
scottleibrand: Did you try bt-pan -d? https://github.com/openaps/oref0/blob/dev/bin/bt-pan#L103

lib/oref0-setup/pancreoptions.json
```diff
@@ -0,0 +1,6 @@
+{
+	"urchin_loop_status": false,
+	"urchin_iob": true,
+	"urchin_Temp_Rate": false,
```
PieterGit: inconsistent up/lowercase. should be urch_temp_rate

lib/oref0-setup/pancreoptions.json
```diff
@@ -0,0 +1,6 @@
...
+	"Notify_Temp_Basal": false
```
PieterGit: inconsistent up/lowercase. should be notify_temp_basal

bin/oref0-bluetoothup.sh
Bender1061: I like that, it's much cleaner, though you do have to pass the bt_mac to it, but it's less disruptive to the pancreabble stuff.

Bender1061:
fixed my case issue. I tend to that that a lot. Thanks for catching that for me

Bender1061:
made a lot of changes pretty fast, I'm going to rerun and install this

Bender1061:
had to fix an issue with the cron, but it's working.

bin/oref0-bluetoothup.sh
```diff
@@ -0,0 +1,20 @@
...
+if ( ifconfig wlan0 | grep -q "inet addr" ) && ( ifconfig bnep0 | grep -q "inet addr" ); then
+   sudo bt-pan client $MAC -d
+fi
+
+if ! ( hciconfig -a | grep -q "PSCAN" ) ; then
+   sudo killall bluetoothd
+   sudo /usr/local/bin/bluetoothd --experimental &
+fi
+
+if ( hciconfig -a | grep -q "DOWN" ) ; then
```
scottleibrand: Now that this is the same action as for PSCAN, should we make it an || check and consolidate them?

Bender1061: the experimental part is the same, but one brings up the hci device while the other it's all ready up, just needs to kill the bluetooth and start it backup. The Down only happened occasionally, and I'm wondering if that's been part of the problem that people have while trying to pair.  as in doing things in the wrong order.

scottleibrand: Ah, right, ok.

mddub:
:tada: :tada: :tada: :tada: :tada:

jasoncalabrese:
Not sure if it's known issue or not, but openxshareble prevents pancreable from staying connected by connecting/disconnecting

Bender1061:
what exactly is it doing? I use MDT for cgm, so I don't have anything with openxsharble to test with. 
like what kind of error messages are you getting?

PR #332, opened 2017-01-13
bucketed_data already has <39 filtered out

MilosKozak:


scottleibrand:
Anyone want to make that change, or should we just leave the duplicate code rather than trying to sync it up?

PR #338, opened 2017-01-20
MDT sensor alarms: If we have an alarm description, use it.

cjo20:
Currently, MDT sensor alarms just show "SensorAlarm <pumpmodel>" in the notes field on nightscout. This change should use the "alarm_description" field for the note if present (e.g. "High Glucose" or "Cal Reminder").

Not yet ready for merge, needs more testing first (from me, if not from others).

### DIFF
--- a/bin/mm-format-ns-treatments.sh
+++ b/bin/mm-format-ns-treatments.sh
@@ -31,7 +31,8 @@ oref0-normalize-temps $HISTORY  \
   | json -e "this.enteredBy = 'openaps://medtronic/$model'" \
   | json -e "if (this.glucose && !this.glucoseType && this.glucose > 0) { this.glucoseType = this.enteredBy }" \
   | json -e "this.eventType = (this.eventType ?  this.eventType : 'Note')" \
-  | json -e "if (this.eventType == 'Note') { this.notes = this._type + ' $model ' + (this.notes ? this.notes : '')}" \
+  | json -e "if (this._type == 'AlarmSensor' && this.alarm_description) {this.notes = this.alarm_description}" \
+  | json -e "if (this.eventType == 'Note' && !this.alarm_description) { this.notes = this._type + ' $model ' + (this.notes ? this.notes : '')}" \
   | json > $OUTPUT
 
 


cjo20:
image
vs
image

cjo20:
Not seen any problems with this yet, other entries on Nightscout seem to still have the correct text.

bin/mm-format-ns-treatments.sh
```diff
@@ -31,7 +31,8 @@ oref0-normalize-temps $HISTORY  \
   | json -e "this.enteredBy = 'openaps://medtronic/$model'" \
   | json -e "if (this.glucose && !this.glucoseType && this.glucose > 0) { this.glucoseType = this.enteredBy }" \
   | json -e "this.eventType = (this.eventType ?  this.eventType : 'Note')" \
-  | json -e "if (this.eventType == 'Note') { this.notes = this._type + ' $model ' + (this.notes ? this.notes : '')}" \
+  | json -e "if (this._type == 'AlarmSensor' && this.alarm_description) {this.notes = this.alarm_description}" \
+  | json -e "if (this.eventType == 'Note' && !this.alarm_description) { this.notes = this._type + ' $model ' + (this.notes ? this.notes : '')}" \
```
scottleibrand: What happens if `this.eventType == 'Note' && this.alarm_description` or if `this._type == 'AlarmSensor' && !this.alarm_description`?  Is that possible?

cjo20: Most of those fields come from the original decocare object. The `decode` call in `decocare\history.py` always sets the `alarm_description` field for any `AlarmSensor` class object, using "Unknown Sensor Alarm" when it doesn't know how to decode it.

This is the only place `alarm_description` is used in the decocare code base, and it isn't referenced in `oref0/dev` apart from the 2 lines  I have added to this script. I believe this means that `alarm_description` should definitely be set if the `_type` is AlarmSensor and not set at any other time.

I don't currently see a way that `this._type == 'AlarmSensor' && !this.alarm_description` could happen with the code as it is right now.

By the same reasoning, other case, `this.eventType == 'Note' && this.alarm_description` will happen, but only when `_type == 'AlarmSensor'`, which will have been handled by the previous statement, which sets the 'notes' field correctly.

Sorry, that's probably as clear as mud. Let me know if you want me to have another go :P

scottleibrand: Nope, that makes perfect sense, thanks.

scottleibrand:
LMK when you're ready and I'll merge it.

cjo20:
Probably best to just go for it.

PR #344, opened 2017-01-24
Allow additional preferences to be customized in oref0-setup

moomoobloo:
In addition to max_iob, allow the following from preferences.json to be specified during setup:
- max_daily_safety_multiplier
- current_basal_safety_multiplier
- bolussnooze_dia_divisor
- min_5m_carbimpact

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -124,7 +124,7 @@ if ! ( git config -l | grep -q user.name ); then
     git config --global user.name $NAME
 fi
 if [[ -z "$DIR" || -z "$serial" ]]; then
-    echo "Usage: oref0-setup.sh <--dir=directory> <--serial=pump_serial_#> [--tty=/dev/ttySOMETHING] [--max_iob=0] [--max_daily_safety_multiplier=3] [--current_basal_safety_multiplier=4] [--bolussnooze_dia_divisor=4] [--min_5m_carbimpact=3] [--ns-host=https://mynightscout.azurewebsites.net] [--api-secret=myplaintextsecret] [--cgm=(G4|shareble|G5|MDT|xdrip)] [--bleserial=SM123456] [--blemac=FE:DC:BA:98:76:54] [--btmac=AB:CD:EF:01:23:45] [--enable='autosens meal dexusb'] [--radio_locale=(WW|US)]"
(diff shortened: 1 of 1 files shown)


bin/oref0-setup.sh
```diff
@@ -108,7 +124,7 @@ if ! ( git config -l | grep -q user.name ); then
     git config --global user.name $NAME
 fi
 if [[ -z "$DIR" || -z "$serial" ]]; then
-    echo "Usage: oref0-setup.sh <--dir=directory> <--serial=pump_serial_#> [--tty=/dev/ttySOMETHING] [--max_iob=0] [--ns-host=https://mynightscout.azurewebsites.net] [--api-secret=myplaintextsecret] [--cgm=(G4|shareble|G5|MDT|xdrip)] [--bleserial=SM123456] [--blemac=FE:DC:BA:98:76:54] [--btmac=AB:CD:EF:01:23:45] [--enable='autosens meal dexusb'] [--radio_locale=(WW|US)]"
+    echo "Usage: oref0-setup.sh <--dir=directory> <--serial=pump_serial_#> [--tty=/dev/ttySOMETHING] [--max_iob=0] [--max_daily_safety_multiplier=3] [--current_basal_safety_multiplier=4] [--bolussnooze_dia_divisor=2] [--min_5m_carbimpact=3] [--ns-host=https://mynightscout.azurewebsites.net] [--api-secret=myplaintextsecret] [--cgm=(G4|shareble|G5|MDT|xdrip)] [--bleserial=SM123456] [--blemac=FE:DC:BA:98:76:54] [--btmac=AB:CD:EF:01:23:45] [--enable='autosens meal dexusb'] [--radio_locale=(WW|US)]"
```
scottleibrand: I left these out of the interactive setup on purpose. It looks like you're only adding them as command line options, but I would prefer we don't add them to the usage prompt either, and make people look at the code to discover how to set them, since they're safety features that most people shouldn't be tweaking.

bin/oref0-setup.sh
```diff
@@ -278,10 +318,28 @@ if openaps vendor add --path . mmeowlink.vendors.mmeowlink 2>&1 | grep "No modul
 fi
 
 cd $directory || die "Can't cd $directory"
-if [[ "$max_iob" -eq 0 ]]; then
+if [[ "$max_iob" -eq 0 && -z "$max_daily_safety_multiplier" && -z "&current_basal_safety_multiplier" && -z "$bolussnooze_dia_divisor" && -z "$min_5m_carbimpact" ]]; then
```
scottleibrand: What happens here if you only specify some of them?

moomoobloo: It falls through to the else clause. The specified options are used, the rest come from the defaults.

bin/oref0-setup.sh
moomoobloo: Done, undocumented them.

moomoobloo:
I verified that the extra options only show up in "run this again later" if they're actually specified.

scottleibrand:
Thanks. Will merge and test, most likely tomorrow, if no one beats me to it.

PR #348, opened 2017-01-25
Missing dependency

beached:
Added dependency on package json.  It is called but wasn't explicitly required


beached:
Why not put it all there. I used apt for jq, but even though json was an npm packaged item I didn't know it.  The big problem with the json one is that if it is not there the name is so generic that it took me a while to find out what I was even looking for.  The search for "json -a" lead me to it as the man page came up.

scottleibrand:
Want to add it to this PR?

PR #350, opened 2017-01-26
Updated for HAPP insulin treatments parsing

taylorfowler:
It may also make sense to look for !(current.timestamp) && current.insulin since I believe HAPP may give you the ability to edit the "Entered by" string (need to confirm that though).


scottleibrand:
Merging this for now.  Should probably revisit a possibly more general solution at some point.

PR #355, opened 2017-01-28
Pancred update

Bender1061:
Few Updates to the pancrebble stuff in oref0.

put the urchin loop completely inside the peb-urchin-status.sh 
add the ability to turn off the urchin loop completely so you can run only notifications.

add of the urchin status for the current temp basal.


scottleibrand:
@Bender1061 hope you're recovering well.

Have you tested this thoroughly enough to confirm that this is better than what's in dev now?  If so, I can merge to dev before the release to master, and then we can do a patch release later if we find anything else that needs fixing...

Bender1061:
my surgeon got the flu, so we have set up the surgery for next week now. 

but yes I've been running this and am very happy with the changes. 

unless someone has any issues, I say lets put it in.

PR #356, opened 2017-01-28
Fix MDT mmtune

scottleibrand:
... to always run on failures, even if wait-for-long-silence fails


Bender1061:
I say do it, I've been running this a while and it's helped a bunch. but I'm wondering if we want to have this happening earlier.  This does work, but we might be able to make it happen earlier, but might require a rework on how the get-bg works.

tim2000s:
I think this removes a symptom but not the problem. I've been running with "openaps mmtune" running before each data extraction cycle and it means each run works. 

What I've observed is that it prior to this mmtune has run every four runs due to issues with errors linked to "pumphistory" or "temp refresh". 

I'm not clear what the root cause is because it happens with various manual commands, so I suspect something along the lines of a pump radio busy error. 

As an aside, it seems to make each run take about 6 'minutes on my Edison and compared to the battery life when pulling Dex data from Nightscout, it cuts the battery life by more than a third.

PR #358, opened 2017-01-28
Saving interactive setup output to a file that can be run later

danamlewis:
To run: `bash ~/myopenaps/oref0-runagain.sh`


Kdisimone:
Worked great.  Thanks!

PR #362, opened 2017-01-29
Take advantage of mmtune-speedup in 0.10.1

scottleibrand:
@cjo20 can you review this please when you get a chance?


cjo20:
Actually, it looks like the latest release of mmeowlink is 11.0, which will need a change in the `egrep` code

Bender1061:
the mmtune speed up is nice, and working for me, it does go much faster so :+1:

scottleibrand:
Fixed the version check; merging.

PR #365, opened 2017-01-29
Mdt data get-bg improvement 

Bender1061:
Change to get-bg so that it gives a message on what it's going and if it fails it will start and error in the pump-loop that will help start an mmtune earlier so less time is wasted waiting for later in the loop to fail out. 
it's a small improvement but improves loop time by a couple mins when an mmtune is needed.


Bender1061:
Ok, I'm a little confused by the all checks have failed? What does that really mean?

I'm actually looking forward to seeing what this does in combination of the faster mmtune adding from the other PR in right now.  That might really improve the average time of loop running with the MDT.

scottleibrand:
I closed and reopened the PR to re-run the unit tests, and they're passing now.

cjo20:
Redirecting the command output to /dev/null means we lose information about what caused the command to fail (if it does fail). This could hinder diagnosing intermittent issues.

scottleibrand:
I've had good success (in https://github.com/openaps/oref0/blob/smb/bin/oref0-pump-loop.sh so far) with just adding `2>&1 >/dev/null | tail -1` to the end of any `openaps report invoke` commands.  That redirects stdout to /dev/null and stderr to stdout, and then prints just the last line of stderr.

cjo20:
The issue with that is that it doesn't allow the return value from the command to be used to exit from the script early (which I guess is why @Bender1061 redirected it to `/dev/null` instead). The command I posted (or a variation of it) prints the last line of the output if the command fails, otherwise it continues.

scottleibrand:
Ah.  This is getting a bit OT for this PR now, but perhaps we should make a generalized function that runs the command, prints the last line of stderr, and returns its return code, and then use that each time we call an `openaps report` command.  I'll give that a shot in my pump-loop, unless you've already done something I can steal.  :-)

cjo20:
Not done anything like that yet, the snippet in my earlier comment was my first attempt.

tim2000s:
I've been running this PR for a couple of hours. It's

tim2000s:
I've been running this PR for four hours and it seems to run okay. It jumps to mmtune more quickly than the previous MDT config did. One thing I'd note is that this (and the other WW branches)  don't install the subg-rfspy git affecting the mmtune performance.

cjo20:
> One thing I'd note is that this (and the other WW branches) don't install the subg-rfspy git affecting the mmtune performance.
They were probably branched before the change that checked for the correct version was made.

scottleibrand:
@tim2000s if you want to test them along with all the other stuff recently merged to dev, you can create a branch from dev and merge the branch you want to test (or merge dev back into the test branch) to get both.

Bender1061:
I've been running this with the faster mmtune the last couple of days, and the performance is a fair amount faster.  but if we have a consensus on what kind of changes we would prefer, I'm game. Anything to help improve. I likely will not be able to get anything changed or tested myself for the next couple days. I'm going into surgery to remove my left foot in the morning (it's been a long time coming, it's the second foot I've taken off) So I'll be slightly drugged up tomorrow.

tim2000s:
I've got the faster mmtune and this merged into my Dev branch. Should be able to install all together later today to provide feedback.

tim2000s:
I've been running the combined setup for the last 90 mins with no issues. The loop is definitely recovering more quickly And kicking through to mmtune when it has issues.

tim2000s:
I'm happy to sign off both the mmtune enhancements and the data get improvement. I've run it all day today with no issues.

scottleibrand:
Ok, cool.  I'm fine merging this as-is, and then working on the output redirection stuff in dev after the release, possibly along with refactoring things out of aliases into more readable shell scripts with proper functions.  If no objections, I'll probably do that this evening.

PR #366, opened 2017-01-29
lowtempimpact based on 30 minutes

AdrianLxM:
@scottleibrand, @MilosKozak 

This is my proposition of how to handle running low temps. If `currenttemp.minutesrunning` is not defined (like in Oref0), it will be assumed that the overall length is 30 minutes.

This will reduce the changed calculation of `30-minutes_running` to `30-(30-currenttemp.duration)` = `currenttemp.duration` and not change anything for the default case.

In case the overall duration is longer (as in some cases in AndroidAPS), it will be calculated as if it was 30 minutes long.

If a low temp is actually running longer than 30 mintues, `adjEventualBG` will rise over `eventualBG` and the "keep it running" branch will not be entered. (As the `eventualBG < min_bg` situation is already handled above.)

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -377,7 +377,7 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
   
     var minutes_running;
     if (typeof currenttemp.duration == 'undefined' || currenttemp.duration == 0) {
-        minutes_running = 0;
+        minutes_running = 30;
     } else if (typeof currenttemp.minutesrunning !== 'undefined'){
         // If the time the current temp is running is not defined, use default request duration of 30 minutes.
         minutes_running = currenttemp.minutesrunning;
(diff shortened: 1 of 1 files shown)


lib/determine-basal/determine-basal.js
```diff
@@ -374,10 +374,20 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
             }
         }
     }
+  
+    var minutes_running;
+    if (typeof currenttemp.duration == 'undefined' || currenttemp.duration == 0) {
+        minutes_running = 0;
```
scottleibrand: This will cause incorrect behavior when duration hits 0: it will cause minutes_running to drop from 29 to 0, and cause lowtempimpact to be calculated incorrectly.  This should default to 30 instead I believe.

AdrianLxM: I believe in this case it does not matter as a duration of 0 or unedfined means that no temp is running at all.
But yes, better use the logical default value in case there are some corner cases or it it might get used in other places in the future.

scottleibrand: Thanks.  It matters because "temp not running" is represented (in OpenAPS) as rate 0, duration 0.  So the behavior you just fixed was causing it to calculate a non-zero lowtempimpact in that (common) situation.

PR #367, opened 2017-01-30
Support multiple bluetooth MACs

scottleibrand:
And if wifi is not working (for example an open wifi with a captive portal) but Bluetooth is, it will disable wifi (release the wlan0 IP) until Bluetooth goes away (most likely after you get home and kill the hotspot manually).  If it doesn't have to do that, it gracefully turns off Bluetooth when it gets a working wifi IP.

### DIFF
--- a/bin/oref0-bluetoothup.sh
+++ b/bin/oref0-bluetoothup.sh
@@ -1,10 +1,6 @@
 #!/bin/bash
 MAC=$1
 
-if ( ifconfig wlan0 | grep -q "inet addr" ) && ( ifconfig bnep0 | grep -q "inet addr" ); then
-   sudo bt-pan client $MAC -d
-fi
-
 if ! ( hciconfig -a | grep -q "PSCAN" ) ; then
    sudo killall bluetoothd
    sudo /usr/local/bin/bluetoothd --experimental &
--- a/bin/oref0-bluetoothup.sh
+++ b/bin/oref0-bluetoothup.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
-MAC=$1
 
 if ! ( hciconfig -a | grep -q "PSCAN" ) ; then
    sudo killall bluetoothd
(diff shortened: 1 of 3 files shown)


Bender1061:
I like that, bit more elegant, and makes sense putting that in the oref0-online. I'll have to test it later

alternateal:
This set up works well for me

tim2000s:
Humm. This didn't work for me!

scottleibrand:
Do you have regular Bluetooth working already? Looks from your message in Gitter that you didn't properly install Bluetooth.

Bender1061:
Worked for me today with no issue so :+1:

tim2000s:
I think I misunderstood the install instructions. From the conversations in Gitter I didn't think a btmac entry was required in the setup process. I reran it specifying a MAC address and it installed okay. bluetoothctl ran correctly.

jbwittmer:
I'm having a bit of problem with this.  When I disconnect our cable modem from the home network (simulating a "captive network") the rig recognizes this at switches over to hotspot via Bluetooth, but from there it never establishes a connection to the internet.  Here's the error message I'm seeing in the network log.  Any ideas?


Starting oref0-online.
At Sat Feb  4 07:18:02 CST 2017 my local IP is: 172.20.10.2

At Sat Feb  4 07:18:02 CST 2017 my public IP is:
Error, cycling networking
Stopping networking (via systemctl): networking.service.
Starting networking (via systemctl): networking.service.
and getting new wlan0 IP

At Sat Feb  4 07:18:34 CST 2017 my public IP is:
At Sat Feb  4 07:18:49 CST 2017 my public IP is:
Error, connecting BT to 64:A3:CB:38:7D:18, getting bnep0 IP and releasing wifi $

At Sat Feb  4 07:19:08 CST 2017 my public IP is: Finished oref0-online.

scottleibrand:
Does it work when walking away from wifi? Have you tried the latest dev with https://github.com/openaps/oref0/pull/376 ?

jbwittmer:
It does work when leaving wifi coverage.   Switches seamlessly, which is nice.  I'll try the latest dev and see if that fixes the problem.  Thanks

PR #369, opened 2017-01-31
install oref0 from npm when on master branch

scottleibrand:
use `npm run global-install` otherwise

Helps address https://github.com/openaps/oref0/pull/302


scottleibrand:
I'll go ahead and merge this, probably tonight, if no objections.

PR #372, opened 2017-02-02
Mdt trend setup

cjo20:
Add medtronic trend data to the default oref0-setup script. Takes in `monitor/cgm-mm-glucosedirty.json`, produces `monitor/cgm-mm-glucosetrend.json`. 

`cgm/cgm-glucose.json` then reads `glucosetrend.json` instead of `glucosedirty.json`.

Also fixes an issue where the _tell resets halfway through glucose data (caused by sensor restart?)

### DIFF
--- a/bin/oref0-mdt-trend.js
+++ b/bin/oref0-mdt-trend.js
@@ -1,9 +1,9 @@
 #!/usr/bin/env node
 
-function findRecord(arr, tell) {
+function findRecord(arr, tell, date) {
 	for (var i = 0; i < arr.length; ++i)
 	{
-		if (arr[i]._tell == tell)
+		if (arr[i]._tell == tell && arr[i].date == date)
 		{
 			return i;
 		}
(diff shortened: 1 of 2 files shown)


scottleibrand:
Does this need someone to test, or have you tested it sufficiently to be confident it's as good as what's in dev now?

cjo20:
I'd like someone to test it if possible given that I don't use oref0-setup in production. 

Having said that, I've tested it myself by running `openaps get-bg`, which showed the correct results, and `openaps upload-bg`, which uploaded the trend data to nightscout correctly.

I've also manually run `pump-loop`, which seems to have managed to set a reasonable rate on my pump (after having to run `monitor/iob.json manually`.

scottleibrand:
Actually, since this doesn't touch oref0-setup or anything else used by non-MDT users, I'm gonna go ahead and merge it based on @cjo20's testing.  If necessary we can fix it in next-dev and MDT CGM users can keep using dev like they've been doing for months.  :-)

Issue #129, opened 2016-06-04
iob should consider suspend/resume/square-waves


bewest:
IOB is impacted by square waves, and suspend/resumes.  It'd be nice to have them considered.

PR #375, opened 2017-02-02
mmhistorytools integration

scottleibrand:
This needs testing.


scottleibrand:
From @thebookins' original PR:

This PR
- adds `mmhistorytools` prepare step in the workflow, which normalizes a treatment history against a basal profile, resolving https://github.com/openaps/oref0/issues/95
- separates out the logic of pump-specific history parsing from iob maths
- considers square waves and pump suspends and resumes in iob calculation as per https://github.com/openaps/oref0/issues/129
- adds an optional flag `--prepared` to `bin/oref0-calculate-iob.js` and `bin/oref0-meal.js` signalling these processes to expect pump history prepared with `mmhistorytools`
- does not break any existing rigs
- modifies `bin/oref0-setup.sh` to use `mmhistorytools` workflow in all new setups.

To acheive this, `lib/iob/` and `lib/meal/` are copied to `lib/iob-prepared/` and `lib/meal-prepared/`, and any changes to support the new workflow are made in the new directories.

Likewise, `tests/iob.test.js` is copied to `tests/iob-prepared.test.js` and the new file is furnished with tests appropriate to the `mmhistorytools` workflow.

Additionally, the calculation for `iobContrib` and `activityContrib` is simplified in `lib/iob-prepared/`: rather than use numerical integration of temp basals (splitting a square delivery into 0.05U microboluses) the activity function `f0` is integrated once (`f1`) and again (`f2`) and the resulting functions are used to determine the iob and activity contributions from square waves and temp basals.

At this stage, the pump history used for autosense is not passed through `mmhistorytools`, but for consistency this might make sense later.

This PR leaves `lib/iob/` and `lib/meal/` as legacy code; if pulled there should be a plan to deprecate this code in time.

To properly deal with pump suspends this PR also depends on a PR to `openaps-mmhistorytools` (https://github.com/loudnate/openaps-mmhistorytools/pull/55).

I realise there are a lot of changes in here - if there is any way to make the edits easier to review / pull I'd appreciate the feedback - this is my first big PR 😄.

Issues:
- `lib/meal-prepared/history.js` ignores Boluses in units of U/hour (i.e. square waves). I'm not sure what the correct behaviour here should be when a square bolus is encountered.
- There are two zoned pump histories hanging around (`monitor/pumphistory-zoned.json` and `monitor/pumphistory-zoned-ns.json`). The latter is used wherever the MDT format of pump history is expected - in `merge-pumphistory-long` and for uploading to NS. It would be simper to use the history passed through `mmhistorytools` for everything if possible.

Bender1061:
so while running this I get 
```
 Starting pump-loop at Fri Feb 17 20:54:11 EST 2017:
Radio ok, Listening: .No pump comms detected from other rigs
MDT cgm data retrieved
MDT cgm data reformatted
Old pumphistory: Refresh
Error, retrying
```
when I try to figure out why it failing
when I run `openaps monitor-pump`
i get 
```
root@Hal9000:~/johnopenaps# openaps monitor-pump
pump://JSON/read_clock/monitor/clock.json
reporting monitor/clock.json
pump://JSON/read_temp_basal/monitor/temp_basal.json
reporting monitor/temp_basal.json
pump://JSON/iter_pump_hours/monitor/pumphistory.json
reporting monitor/pumphistory.json
history://JSON/prepare/monitor/pumphistory-prepared.json
monitor/pumphistory-prepared.json  raised  can't open 'settings/basal_profile.json': [Errno 2] No such file or directory: 'settings/basal_profile.json'
  File "/usr/local/bin/openaps-report", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/root/src/openaps/bin/openaps-report", line 82, in <module>
    app( )
  File "/root/src/openaps/openaps/cli/__init__.py", line 51, in __call__
    self.run(self.args)
  File "/root/src/openaps/bin/openaps-report", line 75, in run
    output = app(args, self)
  File "/root/src/openaps/openaps/cli/subcommand.py", line 52, in __call__
    return self.method.main(args, app)
  File "/root/src/openaps/openaps/reports/invoke.py", line 40, in main
    output = task.method(args, app)
  File "/root/src/openaps/openaps/uses/use.py", line 45, in __call__
    output = self.main(args, app)
  File "/usr/local/lib/python2.7/dist-packages/openapscontrib/mmhistorytools/__init__.py", line 454, in main
    args, kwargs = self.get_program(self.get_params(args))
  File "/usr/local/lib/python2.7/dist-packages/openapscontrib/mmhistorytools/__init__.py", line 445, in get_program
    basal_schedule=_opt_json_file(params.get('basal_profile')),
  File "/usr/local/lib/python2.7/dist-packages/openapscontrib/mmhistorytools/__init__.py", line 78, in _opt_json_file
    return json.load(argparse.FileType('r')(filename))
  File "/usr/lib/python2.7/argparse.py", line 1142, in __call__
    raise ArgumentTypeError(message % (string, e))
argparse.ArgumentTypeError: can't open 'settings/basal_profile.json': [Errno 2] No such file or directory: 'settings/basal_profile.json'
root@Hal9000:~/johnopenaps# 
```

thebookins:
@Bender1061 are you still having trouble with this branch? I tried it today with no problems. I wonder if it might resolve itself after a few runs - as `settings/basal_profile.json` should be available after running `get-settings`.

scottleibrand:
@thebookins @Bender1061 Do you think this is ready to merge to dev for wider testing?

Bender1061:
I'll try testing this again in the morning. I'll let you know.

Bender1061:
Ok, so I tried this again.

Let it run for about 25 mins and it would just fail out of Old pumphistory: Refresh
Error, retrying

I ran openaps get-settings in the aps directory and it suddenly started working.

now why would that be? I've never had to do that before. Or are we needing something from the settings that the history needs to run?

I did check and the settings had not been run, because I started out with a fresh clean APS folder.

thebookins:
@Bender1061 thanks for trying this. Yes, the history now needs the basal profile, because the "prepared" pump history is normalized against the basal profile. I haven't had a chance to look yet, but it may be as simple as changing the order in which the aliases are called.

@scottleibrand apart from the above
 - I think someone should review `oref0/lib/iob-prepared/calculate.js`, since I've used an analytical function to calculate the iob contribution from square waves. I'm pretty sure the maths is correct, but we should double-check to be safe.
 - If everyone who updates the dev branch also runs `oref0-setup.sh`, then we can consider removing the legacy code now. The current PR uses a command line switch `--prepared` to switch between the mmhistorytools implementation and the original. Might be simpler to get rid of this? Also happy to do this as a later PR.

I'll have a look tonight at the issue @Bender1061 raises. I assume I can just push a PR to this branch as normal?

scottleibrand:
If anyone wants to pick this up and work on integrating it with oref0 0.5.x, please speak up.  Otherwise I'm afraid we're going to have to close it out for lack of interest in doing all the hard integration work.

thebookins:
Happy to integrate into 0.5.x, as long as we have consensus that it's still the right thing to do (some of these features have been rolled in to oref0 since). Personally I think it's worth it so that suspends etc are properly accounted for.

scottleibrand:
I think that's actually the crux of the issue: it's gonna be a lot of work to ensure that we have parity with the existing method.  It might actually be easier to just track suspends within the oref0 framework: I'm not sure.

thebookins:
Agree. I do like the modularity of the mmhistorytools approach, plus that it keeps oref0 pump-agnostic. I'll have a think about how to approach it. Perhaps a series of more easily-digestible PRs.

scottleibrand:
Pump agnosticism is achieved in practice via AndroidAPS.  I don't know of any other kind of pump that we would want to use with the openaps toolkit other than older MDT ones, so that is a rather theoretical concern.

I do like your idea of chopping the problem up into bite-sized pieces.

scottleibrand:
It seems like this change is too major to undertake as-is.  If no objections, I'm going to close this PR and we can open new ones later for the more bite-size chunks.

PR #377, opened 2017-02-04
save crontab to ~/crontab.txt and prompt to crontab -r before recreating

scottleibrand:


bin/oref0-setup.sh
```diff
@@ -635,6 +635,14 @@ echo
 
 read -p "Schedule openaps in cron? y/[N] " -r
 if [[ $REPLY =~ ^[Yy]$ ]]; then
+
+    echo Saving existing crontab to $HOME/crontab.txt:
+    crontab -l | tee $HOME/crontab.txt
```
moomoobloo: Maybe call it crontab.old.txt?

PR #379, opened 2017-02-05
increase frequency of enlite upload to nightscout

Bender1061:
This increases the frequency of enlite bg uploads to nightscout.
by removing the upload-bg from the uploads alias
having the nightscout/glucose.json report called by the get-bg
having get-bg called from muliple points in the pump-loop
putting the upload-bg in the cron

also getting rid of the cgm device since everything is on the pump

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -657,6 +657,9 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
         (crontab -l; crontab -l | grep -q "cd $directory && ps aux | grep -v grep | grep -q 'openaps get-bg'" || echo "* * * * * cd $directory && ps aux | grep -v grep | grep -q 'openaps get-bg' || ( date; openaps get-bg ; cat cgm/glucose.json | json -a sgv dateString | head -1 ) | tee -a /var/log/openaps/cgm-loop.log") | crontab -
     fi
     (crontab -l; crontab -l | grep -q "cd $directory && ps aux | grep -v grep | grep -q 'openaps ns-loop'" || echo "* * * * * cd $directory && ps aux | grep -v grep | grep -q 'openaps ns-loop' || openaps ns-loop | tee -a /var/log/openaps/ns-loop.log") | crontab -
+    if [[ ${CGM,,} =~ "mdt" ]]; then
+       (crontab -l; crontab -l | grep -q "cd $directory && ps aux | grep -v grep | grep -q 'openaps upload-bg'" || echo "* * * * * cd $directory && ps aux | grep -v grep | grep -q 'openaps upload-bg' || openaps upload-bg | tee -a /var/log/openaps/upload-bg-loop.log") | crontab -
+    fi
     if [[ $ENABLE =~ autosens ]]; then
         (crontab -l; crontab -l | grep -q "cd $directory && ps aux | grep -v grep | grep -q 'openaps autosens'" || echo "* * * * * cd $directory && ps aux | grep -v grep | grep -q 'openaps autosens' || openaps autosens | tee -a /var/log/openaps/autosens-loop.log") | crontab -
     fi
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -540,11 +540,6 @@ fi
 if [[ ${CGM,,} =~ "mdt" ]]; then
     sudo pip install -U openapscontrib.glucosetools || die "Couldn't install glucosetools"
     openaps device remove cgm 2>/dev/null
-    if [[ -z "$ttyport" ]]; then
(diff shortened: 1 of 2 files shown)


cjo20:
What's wrong with the current frequency of uploads to nightscout?

You can get fresh BG data at most every 5 minutes. It currently kicks the ns-loop alias every minute, https://github.com/openaps/oref0/blob/dev/bin/oref0-setup.sh#L664 which calls `openaps upload` (https://github.com/openaps/oref0/blob/dev/lib/oref0-setup/mdt-cgm.json#L144), which in turn calls `openaps upload-bg` (https://github.com/openaps/oref0/blob/dev/lib/oref0-setup/mdt-cgm.json#L135)

Similarly, the loop attempts to kick off get-bg every minute: https://github.com/openaps/oref0/blob/dev/bin/oref0-setup.sh#L662

Neither of these operations should take more than a couple of minutes to run, so new data should appear well before the next reading is available.

This means that new data should be uploaded pretty much as soon as it is available. `get-bg` is a pretty expensive operation in terms of run time and pump comms (and therefore battery life). It also makes it harder for other rigs to find free airtime to communicate with the pump.

Bender1061:
whenever I watch the time on nightscout for the last cgm data, it would be usually be more than 10 mins old, and that's with everything working correctly. and then you don't get an update till it's 16+ mins old
you throw in a error or an mmtune in there and you can be sometimes 20+ mins old.
with this it get normally less than 10 (sometimes 4 or 5) and and it's usually updated before it gets to 10 mins old

while the cgm data normally updates fairly quickly for openaps to do with it as it needs, the uploads tend  to happen at the wrong time it reads the last cgm and then tries to upload it when the next one might be a min or two out of getting some newer data. 

as for battery life, yeah It might take a little more batter, but using the enlite is going to take more anyways. 

and as for the other rigs, if the one rig is working, what does it matter if the other rig can't communicate?

cjo20:
It sounds like you're having problems with communications succeeding, rather than cgm data uploading late. Although I don't use oref0-setup myself, from what I've seen it seems unlikely that it would regularly take 10-15 minutes to get through the loop in its current form. How long does it take to run `get-bg` and `ns-loop`? I'd be surprised if either took more than a couple of minutes.

By "more battery" I meant "more than should be necessary to run with the enlite".

As for other rigs, if you're moving from one room to the other, then your pump may be out of range of one rig, but the other rig won't successfully detect silence to start executing its own loop.

Bender1061:
I'm doing some tests to see if I can alleviate some of your concerns, and still increase the upload frequency.

cjo20:
A good first step would be to check how long each of those reports takes to run.

Bender1061:
haha, yup already ahead of you on that. Granted, I think the rig that I have my changes on must have had a older decocare or something, because yes, the new dev on everything is actually much better than, what the rig which must have had a slightly older version of all of dev. So I'm doing further testing, I still do notice it uploading faster, but I might not need to run the get-bg more than once per loop, which is the biggest issue right now, I still think it's a good idea to have the nightscout/glucose.json report being called with the reformatting. That way it's easy to check if the upload even needs to happen, which might reduce some of the traffic to nightscout. but that probably can happen without as much drastic changes, as I have made currently.

Bender1061:
So we don't really have a good way to log when a most current bg is uploaded to nightscout, well I can see it with my new cron but comparing that to the current dev there is not a way.

So I first tried to see if just changing the dev to just take in account the nightscout/glucose in the get-bg and changing upload-bg to only update to nightscout if there was a new glucose in nightscout/glucose.json. 
problem with this is how long it takes the ns-loop to run.

so I removed the upload-bg from the ns-loop and have it uploading as a seperate cron. 
this did improve and get us uploading faster than dev, but since I have the default pump-loop being used, so get-bg is only called once per loop, that does minimize the concern about battery and keeps the pump communications the same as before. (get-bg takes a few more seconds, due to the formating of nightscout/glucose.json)
the upload is not as fast as it can take about 5 mins from one get-bg to the other, and takes about another 3 to 4 mins to upload that data to nightscout.
What the original plan was to reduce that 5 mins to 2 or 3 mins. But battery life and pump-comms is a valid concern. So I'm going to put that pump-loop back to original, 

Do you have anyother concerns about this @cjo20 ?

If not I think we are at a point for people to test this. and see if anyone else notices an improvement.

scottleibrand:
Another trick might be to have pump-loop sleep for the amount of time until the next BG is expected to be available. There is an example in the Dexcom cgm-loop stuff used by openxshareble and g4-raw if you want to copy that...

Bender1061:
that is a good idea, I'll look into that later today.

Bender1061:
Will more testing, I'm not really seeing a big advantage to this, so I'm going to close it

PR #380, opened 2017-02-05
Changes necessary for running autotune on OS X 

elnjensen:
Haven't updated the docs yet - using this on OS X will require installing `coreutils` from homebrew, e.g. with `brew install coreutils`.


scottleibrand:
Merged via #384

PR #382, opened 2017-02-06
Adding explorer board to setup script prompt and auto port-itize it if so

danamlewis:


jmcrawford:
The syntax: "Explorer Board? [Y/n]" to me would imply that Y is the default, it is what would be chosen if you just hit enter.  But with no response, it assumes "n".  Do you intend that Y is chosen by default?

danamlewis:
Y is meant to suggest that it's the most frequent pick; but not necessarily the default.

scottleibrand:
Opened https://github.com/openaps/oref0/pull/389 to use the same syntax we use elsewhere in the script to indicate the default choice.

PR #388, opened 2017-02-06
Automatically creating "l" for log shortcut

danamlewis:
This attempts to create a tail-log file (oref0-tail-log.sh) and have it in package.json as an automatic shortcut for "l" once this version is run by a user.

(chmod +x happened, too)


Kdisimone:
This will be great!

danamlewis:
(add "wifi" shortcut to tail the network log, which will now include wifi name (thanks @Kdisimone for idea) plus IP)

Issue #373, opened 2017-02-02
Autotune: Automatically handle pumpprofile.json/autotune.json creation


jessiepusateri:
In the Phase C setup, the following steps:

- E. Create a pumpprofile.json that is the same as your profile.json. On the command line run: cp profile.json pumpprofile.json
- F. Create a third file from the command line by running: cp profile.json autotune.json

should be handled in the bash script. If the files are not present, it should prompt you "The following files are missing: pumpprofile.json, autotune.json. Would you like to them to be created from the profile.json file? (Y/N)".

Yes will automatically do the copying/creation for you.

danamlewis:
(Just for reference, if anyone wants to jump in on this, the additions will be similar to the type of thing I just added to oref0-setup.sh recently, see: https://github.com/openaps/oref0/pull/382/files - happy to help if someone wants to kickstart this)

danamlewis:
I believe this is resolved via https://github.com/openaps/oref0/pull/392, so closing this as an issue.

PR #392, opened 2017-02-08
Automatically copy pumpprofile.json & autotune.json from profile.json

jessiepusateri:
In reference to #373

### DIFF
--- a/bin/oref0-autotune.sh
+++ b/bin/oref0-autotune.sh
@@ -127,6 +127,9 @@ else
   exit 1
 fi
 
+# Create pumpprofile.json, autotune.json from profile.json
+cp -up profile.json pumpprofile.json
+cp -up profile.json autotune.json
 
 # Get profile for testing copied to home directory. "openaps" is my loop directory name.
 cd $directory && mkdir -p autotune
--- a/bin/oref0-autotune.sh
+++ b/bin/oref0-autotune.sh
@@ -127,13 +127,11 @@ else
   exit 1
 fi
 
-# Create pumpprofile.json, autotune.json from profile.json
-cp -up profile.json pumpprofile.json
-cp -up profile.json autotune.json
-
 # Get profile for testing copied to home directory. "openaps" is my loop directory name.
 cd $directory && mkdir -p autotune
 cp settings/pumpprofile.json autotune/profile.pump.json
+# This allows manual users to be able to run autotune by simple creating a settings/pumpprofile.json file.
+cp -up settings/pumpprofile.json settings/profile.json
 # If a previous valid settings/autotune.json exists, use that; otherwise start from settings/profile.json
 cp settings/autotune.json autotune/profile.json && cat autotune/profile.json | json | grep -q start || cp autotune/profile.pump.json autotune/profile.json
 cd autotune
--- a/bin/oref0-autotune.sh
+++ b/bin/oref0-autotune.sh
@@ -130,7 +130,7 @@ fi
 # Get profile for testing copied to home directory. "openaps" is my loop directory name.
 cd $directory && mkdir -p autotune
 cp settings/pumpprofile.json autotune/profile.pump.json
-# This allows manual users to be able to run autotune by simple creating a settings/pumpprofile.json file.
+# This allows manual users to be able to run autotune by simply creating a settings/pumpprofile.json file.
 cp -up settings/pumpprofile.json settings/profile.json
 # If a previous valid settings/autotune.json exists, use that; otherwise start from settings/profile.json
 cp settings/autotune.json autotune/profile.json && cat autotune/profile.json | json | grep -q start || cp autotune/profile.pump.json autotune/profile.json


scottleibrand:
It looks like we actually need to do things a bit differently to make this compatible with OpenAPS rigs running autotune alongside the rest of oref0.  

I think what would work best for the folks running things manually, while not breaking people running this in their closed loop would be to actually have people edit settings/pumpprofile.json rather than settings/profile.json, and then have the script do a `cp -up settings/pumpprofile.json settings/profile.json`.  It turns out that https://github.com/openaps/oref0/blob/dev/bin/oref0-autotune.sh#L133-L135 actually already copies settings/pumpprofile.json over to autotune/profile.json if it doesn't exist, so I think we're good there.

We also need to place the `cp -up settings/pumpprofile.json settings/profile.json` line a bit lower, so that it's running after https://github.com/openaps/oref0/blob/dev/bin/oref0-autotune.sh#L132

Would you like to take a shot at updating that, or would you prefer I do it?

-Scott

P.S. If you want all the gory details, here's what we have on one of our rigs:

```
pi@pi2-erf ~/myopenaps $ ls -la settings/profile.json settings/pumpprofile.json settings/autotune.json autotune/profile.json autotune/profile.pump.json  | sort -gk 8
-rw-r--r-- 1 pi pi 1718 Feb  7 00:05 autotune/profile.pump.json
-rw-r--r-- 1 pi pi 1965 Feb  7 00:08 autotune/profile.json
-rw-r--r-- 1 pi pi 1965 Feb  7 00:08 settings/autotune.json
-rw-r--r-- 1 pi pi 1718 Feb  7 18:55 settings/pumpprofile.json
-rw-r--r-- 1 pi pi 2617 Feb  7 19:09 settings/profile.json
```

The settings/pumpprofile.json and settings/profile.json are the newest files, because they are created (by oref0-get-profile) from the various other files in settings, and in the profile.json case, from the settings/autotune.json file.

```
pi@pi2-erf ~/myopenaps $ openaps report show settings/pumpprofile.json --cli
openaps use get-profile shell --model=settings/model.json settings/settings.json settings/bg_targets.json settings/insulin_sensitivities.json settings/basal_profile.json preferences.json settings/carb_ratios.json
pi@pi2-erf ~/myopenaps $ openaps report show settings/profile.json --cli
openaps use get-profile shell settings/settings.json settings/bg_targets.json settings/insulin_sensitivities.json settings/basal_profile.json preferences.json settings/carb_ratios.json settings/temptargets.json --model=settings/model.json --autotune settings/autotune.json
```

Those end up getting called, in order, by `openaps get-settings`:
```
pi@pi2-erf ~/myopenaps $ openaps alias show get-settings
get-settings report invoke settings/model.json settings/bg_targets_raw.json settings/bg_targets.json settings/insulin_sensitivities_raw.json settings/insulin_sensitivities.json settings/basal_profile.json settings/settings.json settings/carb_ratios.json settings/pumpprofile.json settings/profile.json
```

scottleibrand:
The other option, of course, would be to do as you suggested in https://github.com/openaps/oref0/issues/373 and only copy things if the files don't exist at all.  Since those files should always exist on OpenAPS rigs, that would only trigger for manual installations.

jessiepusateri:
fullsizerender

Okay, I'm a visual person so I drew out my understanding of the process. Please correct anything that's not quite right.

Step 1: Create `autotune/profile.pump.json`.
- Do this by copying `settings/pumpprofile.json` to `autotune/profile.pump.json`.

Step 2: Create `autotune/profile.json`.
- A. First try to use `settings/autotune.json`.
- B. If that's empty/missing, use `settings/profile.json`.

Step 3: Create `autotune/autotune.json`.
- Do this at the end of running autotune to save the new results.

If we add in Step 1.5, which creates `settings/profile.json` from `settings/pumpprofile.json`, I think this will work correctly for manual builds because Step 2B will be able to create `autotune/profile.json` from `settings/profile.json`. (Actually now that I'm writing this, it could be done as Step 0.5 or Step 1.5 as long as it is done before Step 2). I can go ahead and make those changes.

Which leaves me with two questions:
1. Why is `autotune/profile.pump.json` not named `autotune/profile.json`?
2. Is `settings/autotune.json` only present when autotune is run on the OpenAPS rigs?

scottleibrand:
1. They are two different things. Autotune needs to know your pump profile to know what to anchor on for the 20-30% safety caps. 

2. Yeah. It's done at https://github.com/openaps/oref0/blob/master/bin/oref0-setup.sh#L685

scottleibrand:
And I believe everything else is correct and would work.

jessiepusateri:
Ok, thanks for the info!

bin/oref0-autotune.sh
```diff
@@ -127,13 +127,11 @@ else
...
 fi
 
-# Create pumpprofile.json, autotune.json from profile.json
-cp -up profile.json pumpprofile.json
-cp -up profile.json autotune.json
-
 # Get profile for testing copied to home directory. "openaps" is my loop directory name.
 cd $directory && mkdir -p autotune
 cp settings/pumpprofile.json autotune/profile.pump.json
+# This allows manual users to be able to run autotune by simple creating a settings/pumpprofile.json file.
```
scottleibrand: "simply"

scottleibrand:
I think this is ready for testing by both manual and closed loop users.

PR #395, opened 2017-02-09
oref0-setup.sh new cron line for proper shutdown

ceben80:
Adding a new cronline for proper shutdown in case the edison battery is empty (< 3050mV; 2950mV is dead)

* running each 15 min (in case of failure one is able to power on and perform `crontab -r`
* using `openaps battery-status`
* parsing monitor/edison-battery.json using `awk` searching for the 2nd column `$2` for the voltage value
* e.g. content of monitor/edison-battery.json `{"batteryVoltage":4057, "battery":88}`
* in case the voltage is < 3050mV `sudo shutdown -h now` will be performed
* when monitor/edison-battery.json is empty or does not exist, nothing is done

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -694,6 +694,12 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
     if [[ ! -z "$BT_PEB" || ! -z "$BT_MAC" ]]; then
        (crontab -l; crontab -l | grep -q "oref0-bluetoothup" || echo '* * * * * ps aux | grep -v grep | grep -q "oref0-bluetoothup" || oref0-bluetoothup >> /var/log/openaps/network.log' ) | crontab -
     fi
+
+    # proper shutdown once the EdisonVoltage very low (< 3050mV; 2950 is dead)
+    if egrep -i "edison" /etc/passwd 2>/dev/null; then
+     (crontab -l; crontab -l | grep -q "cd $directory && openaps battery-status" || echo "*/15 * * * * cd $directory && openaps battery-status; awk '/batteryVoltage/ {if (\$2<=3050)system(\"sudo shutdown -h now\")}' $directory/monitor/edison-battery.json") | crontab -
+    fi
+    
     crontab -l | tee $HOME/crontab.txt
 fi
 
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -697,7 +697,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
 
     # proper shutdown once the EdisonVoltage very low (< 3050mV; 2950 is dead)
     if egrep -i "edison" /etc/passwd 2>/dev/null; then
-     (crontab -l; crontab -l | grep -q "cd $directory && openaps battery-status" || echo "*/15 * * * * cd $directory && openaps battery-status; awk '/batteryVoltage/ {if (\$2<=3050)system(\"sudo shutdown -h now\")}' $directory/monitor/edison-battery.json") | crontab -
+     (crontab -l; crontab -l | grep -q "cd $directory && openaps battery-status" || echo "*/15 * * * * cd $directory && openaps battery-status; cat $directory/monitor/edison-battery.json | json batteryVoltage | awk '{if (\$1<=3050)system(\"sudo shutdown -h now\")}'") | crontab -
     fi
     
     crontab -l | tee $HOME/crontab.txt


ceben80:
@scottleibrand my solution in the past relied on `monitor/edison-battery.json` but this report does not exist anymore. EdisonVoltage now is creating through `openaps battery-status` but the output is completely different, so the above in my PR won't work

output from `monitor/edison-battery.json`:
Battery Voltage: 3854mV (67%)

output from `openaps battery-status`:
{"batteryVoltage":3856, "battery":68}

You see, looks different, any idea as linux expert how to realize the "shutdown" cron line?

Thanks in advance!

ceben80:
@scottleibrand 

are you able to implement:

**before** shutdown (in case of nearly empty lipo) -> loop in "failsafe", cancel TBR, or what do you think?

scottleibrand:
oref0 is already designed for any temp basal to be the safest course of action if run to completion. I don't think the additional complexity would be worthwhile. Also, most batteries will die when the PWD is awake, so if we wanted to do anything extra, alarming might be more useful.

ceben80:
@scottleibrand fully agree regards with the cron line

ceben80:
@scottleibrand that's the new cron line relying on `openaps battery-status` 

```
*/15 * * * * cd /root/myopenaps && openaps battery-status && cat /root/myopenaps/monitor/edison-battery.json | json batteryVoltage | awk '{if ($1<=3050)system("sudo shutdown -h now")}'
```

I am going to test this and come back...

ceben80:
@scottleibrand 

the amended cron line based on the right `openaps openaps battery-status` json output is now changed, I have tried with several cut-off voltages (manually change the cronline):

4050, 3950, 3850, 3700

Work's fine for me regarding the `sudo shutdown -h now` once the voltage was touched.

Before merging to DEV please:

* double-check the syntax of adding the cronline
* after running oref0-setup.sh and inserting the cron line to the cron job, please test whether the cut-off voltage at 3050mV works fine or set higher by changing this value.

scottleibrand:
Thanks, will do. 

How many minutes of runtime does the Edison usually have left when it hits 3050? Say, with a 2000mAh battery?

ceben80:
Good question, I started with 3100 in my cron line but each minute (okay risky in case of failure), then I reduced to 3050 (near by 2950 the dead value). Maybe we should increase again to 3100 due to the fact the cron is only running each 15 min?!

jbwittmer:
I would love to have this functionality for our small rig with the 2000 mAh battery but when I added the above cron line, I did not get a shut down.  Tried setting to threshold to 4000 just to test and confirmed that the edison-battery.json had value less than that, but never shut down.

ceben80:
@jbwittmer I have a few questions:

that's your cronline?

```
*/15 * * * * cd /root/myopenaps && openaps battery-status; cat /root/myopenaps/monitor/edison-battery.json | json batteryVoltage | awk '{if ($1<=3050)system("sudo shutdown -h now")}'
```
does you have ` openaps battery-status' in your openaps.ini?

Please try this here (in your myopenaps directory):

`openaps battery-status; cat /root/myopenaps/monitor/edison-battery.json | json batteryVoltage`

jbwittmer:
No, what I had used was slightly different (and I was triggering it more
often with testing).

*/5 * * * * cd /root/myopenaps && openaps battery-status && cat
/root/myopenaps/monitor/edison-battery.json | json batteryVoltage | awk
'{if ($1<=4000)system("sudo shutdown -h now")}'



When I execute the following from myopenaps:


openaps battery-status; cat /root/myopenaps/monitor/edison-battery.json |
json batteryVoltage


I got the following return:

3934

jbwittmer:
OK I used the revised cronline edit ( semicolon after openaps
battery-status rather than && ) and set the threshold to 4000 for testing
and it shut down the system properly.  Looks like that fixed it.

Thanks.

ceben80:
@scottleibrand I think that's sounds good, questions are, does the above syntax work in oref0-online.sh and what should be the final cut-off voltage if it's running "only" each 15 min? 3050, 3100 or a little bit higher at 3200?

scottleibrand:
Can we get some additional input on the desired user experience here?  Do we want the rig to automatically shut down when battery is low and likely to die in the next 15-30 minutes?  Or do we want to wait until it's critically low, even if that means the rig might die before we shut it down manually?  I've never seen any problems from letting a rig's battery die completely, so I don't really have an opinion on the desired behavior.

ceben80:
@jbwittmer you did have negativ experiences with a hard edi shutdown?
I run this cron line for a long time every minute, ok, that's risky and one must really logon quickly when something goes wrong to disable the crontab :) I think we can reduce the battery check to 10 min. (or 5 min?) Sufficient time to interact in case of failure? And  raising the cut-off voltage to 3200 mV, what do you think @scottleibrand ?

tim2000s:
I've run both with and without this behaviour, and in all honesty, I've not hit any issues when the battery goes flat. Obviously a safe shutdown makes more sense, but I think you're likely to be okay doing it at 4 or 5% battery.

fjungman:
So, the Edison doesn't corrupt the flash like the RPi does when it loses power?  If it doesn't, if it isn't a risk, then I don't see the point of this complexity.  Let it run until it shuts off.  I assume the power management unit will prevent it running in brown-out conditions.  What is the upside, if not preventing flash corruption?

ceben80:
I know 2 looper with a broken file system after a hard shutdown, in both cases the edison was flashing new :(

fezulin:
I am using the cron line offered by ceben80. I changed the value from 3050 to 3100. It seems to me, that this is working well for my 2000mAh battery.

libxmike:
yes it makes sense and could/should be implemented with email notification when the system is going down, I posted while ago (end of 2016? ) this solution in intend-to-bolus.
With email notification you know that you not looping anymore.

jbwittmer:
Yes I did have a rig that became corrupted after a hard shutdown.  I have been using the cron line offered by ceben80.  I have used a voltage of 3300 with 15 minute intervals and have confirmed by reviewing syslog that it shuts down appropriately when reaching the threshhold.  I have found that voltage drops very rapidly when below 3300 and that the battery could deplete prior to the next 15 minute cron run when using lower values.  Like others I have previously experienced plenty of hard shut downs without system corruption, but it's a situation I prefer to avoid entirely if possible.  Because of this issue we have taken to using a 2500mAh battery routinely, sacrificing size (slightly) to better match my son's rig run-time requirements.

Bender1061:
I don't have an issue with this, while I've never seen any problems this will likely turn off your rig a few mins faster, but really not going to effect anyone negatively.

ceben80:
Only FYI as I remember Chris defined in the EdisonVoltage c code 2950mV as completely dead.

dramageek:
Though I've never experienced corruption, I'd also prefer to lose the rig a few minutes early and execute a clean shutdown.

I thought I saw somewhere reference to the wireless chip on the Edison shutting down before the Edison does? At that point, aren't most people offline anyway? Does anyone know if there's some flag that gets thrown when that chip goes offline that we could use to trigger a clean shutdown of the rest of the system?

cjo20:
Wifi and BT should shut down before the Edison dies according to Intel. Not aware of any flag. If someone's battery is low they might be able to check the status of `ifconfig` to see if there is any change we can grep for. 

Personally, id go with shutting down early; on occasions where my battery has been completely drained to the point where it shuts off, I've had problems getting my Edison to start again without leaving it to charge for a while with no power.

ceben80:
@cjo20 what cut off voltage would you recommend regarding to the cron job to be run every 15min (I would run it very 5 min btw)?

cjo20:
@ceben80  I haven't done enough logging to be able to suggest a value. If people have real-world values that work for them, then I'd go with one of those.

scottleibrand:
Sounds like there's a consensus for adding this in, but we may need to do some more real-world testing to adjust the voltage cutoff if needed.  Until we have more data, it seems like a low voltage cutoff is a good place to start, and we can raise it if we're seeing the rig die before it triggers.  So I'm going to merge this as-is to dev for testing there.

PR #396, opened 2017-02-10
Changes to oref0-online

ptinson:
Replace how we get the SSID, use iwgetid instead.
Replace how we the IP of the wlan and bnep interfaces to use the ip command.


scottleibrand:
Merging to dev for testing there.

PR #401, opened 2017-02-12
Rename G4-raw to G4-upload, and G4 to G4-local-only

scottleibrand:
Also removes ShareBLE from the interactive prompts per @kdisimone's #398


scottleibrand:
Need someone to run through each G4 CGM option on a new myopenaps directory and make sure they get set up as intended and work properly.

scottleibrand:
Merging this to dev for testing.

PR #402, opened 2017-02-13
Add Shortcuts for Log files to help with terminal

Bender1061:
add of log shortcuts to .bash_profile

shortcut is just name of file without the .log.

for example you only have to type in `pump-loop` instead of 
`tail -n 100 -F /var/log/openaps/pump-loop.log`

If someone has a much better idea for what the shortcut names should be, I am open to changing them. 

checks if Alias for that shortcut is there, if not it write it to the end of the file.

this is similar to @danamlewis #388 but adds all the log shortcuts.


Bender1061:
So I have tested this. 
I do have a problem with the reloading 
source .bash_profile
I've also tried 
. ~/.bash_profile 

did not work, but a reboot did make everything start working.

Bender1061:
issue with running pump-loop as the alias to run since that is also a command. Changed everything with a log at the end so it's now pump-looplog

Bender1061:
Another option that I can do, is put in a .sh in bin with all the shortcuts, and we can make it so you can run it manually. that way if you don't want it you can. that's currently what I've been doing to test other things right now. but I've been enjoying using the shortcuts, cause it's made life much easier

Bender1061:
made changes to have shortcuts added by a bash script

scottleibrand:
@danamlewis @kdisimone What do y'all think of this approach to installing aliases by default, compared to what we now have in the docs?

Kdisimone:
Would this solution prevent me from having customized ones such as edit-pref or edit-wifi?  Would we have some wag to account for if people wanted to add ones beyond just the logs?

Kdisimone:
That said...I love having things included in the setup script. Much easier to troubleshoot too if I can tell someone edit-wifi vs walking them through directories and commands.

scottleibrand:
This would just add things to your .bash_profile.  If you want to have additional custom aliases there as well, it won't override those.  That said, if you have any other shortcuts you think everyone should have, maybe this would be a good place to be able to get them added?

Kdisimone:
Yup, I'll add other commands as a pr after merge. Looks like a good method. Thanks

Issue #390, opened 2017-02-07
COB Decay Issues


JELCRAWFORD:
COB decay problems - see images and related treatment info:

![jan 4th - cob issues 1](https://cloud.githubusercontent.com/assets/24684595/22681474/a95c132e-ecca-11e6-8302-2061827b8dd8.PNG)
![image 1 - jan 4th](https://cloud.githubusercontent.com/assets/24684595/22681479/abf21c82-ecca-11e6-8b2d-059760860544.png)
![image 2 - jan 4th](https://cloud.githubusercontent.com/assets/24684595/22681483/b6446a5a-ecca-11e6-878d-d076b0634228.png)

2nd data set:
![jan 3rd cob issues](https://cloud.githubusercontent.com/assets/24684595/22681486/c59f66da-ecca-11e6-8714-fe391124d731.PNG)
![treatments jan 3](https://cloud.githubusercontent.com/assets/24684595/22681489/c81c8a78-ecca-11e6-9bcf-36d17aec1db0.png)

JELCRAWFORD:
We have ongoing discrepancies with COB display seeming to have much higher COB than should be onboard.  This does not seem to be related to double entries.

scottleibrand:
Had a troubleshooting session with @JELCRAWFORD and @jmcrawford, and it looks like they're seeing this due to mis-sorted treatments.  Opened https://github.com/openaps/oref0/pull/404 to address that, and set them up to run it to test and see if it fixes the issue.

JELCRAWFORD:
HI Scott - I think this is working fine - we have had only 1 incident that we saw an increase from zero  to 20 carbs around 11pm at night - we have not noticed it any other than that, and meal cob seems to be consistently accurate.

scottleibrand:
Thanks. Merged #404

PR #404, opened 2017-02-13
Sort treatments before calculating COB

scottleibrand:
Improperly sorted treatments appear to be causing, or at least contributing to, https://github.com/openaps/oref0/issues/390, in cases where carbs come from multiple sources (in the Crawford's case, both pump and NS).  This PR makes sure they're sorted properly before calculating COB.

### DIFF
--- a/lib/meal/total.js
+++ b/lib/meal/total.js
@@ -1,5 +1,4 @@
 var tz = require('moment-timezone');
-var _ = require('lodash');
 var calcMealCOB = require('oref0/lib/determine-basal/cob-autosens');
 
 function diaCarbs(opts, time) {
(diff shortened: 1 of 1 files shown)


scottleibrand:
@JELCRAWFORD and @jmcrawford, please let us know if this helps with your COB miscalculation.

tim2000s:
I've noticed that this is especially an issue if you lose connectivity between NS and the rig. When you come back on connection, it pulls all the data from NS into COBhistory, and about 10-15 mins later recovers to the correct level.

scottleibrand:
Per https://github.com/openaps/oref0/issues/390#issuecomment-282456688 this looks good to merge to dev.

PR #405, opened 2017-02-13
fix missing bluetooth options in oref0-runagain.sh

PieterGit:


dramageek:
I realize that I used them in my previous edit, but do we need the single quotes around the addresses or not? Readthedocs manual method doesn't have the quotes. Do they not matter?

scottleibrand:
I believe those single quotes prevent the $ variables from being evaluated in this script, and instead writes them out literally as they are into the oref0-runagain script to be evaluated when that is run.

Issue #397, opened 2017-02-11
autotune does not use api-secret when needed


beached:
I have a nightscout setup that is in default deny mode and needs an api-secret to view data.  I cannot see how or any references to it in the autotune code when calling curl to retrieve data.  I am not sure the approach, but seems this may be tangentially related to #299

gregtcm:
This showed up for me with this error:
```
src/oref0/lib/autotune-prep/categorize.js:21
        var glucoseData = opts.glucose.map(function prepGlucose (obj) {
                                       ^

TypeError: opts.glucose.map is not a function
    at categorizeBGDatums (/home/gnu/src/oref0/lib/autotune-prep/categorize.js:21:40)
```

PieterGit:
This is fixed with https://github.com/openaps/oref0/pull/563
Autotune now accepts: 
- no authentication (readable Nightscout)
- hashed API_SECRET. API_SECRET must be set as environment variable to the autotune script.
- token based authentication, API_SECRET starts with `token=hash`

lgruen:
Maybe my version of curl (7.54.0) on Mac OS is weird, but I had to change the `ns-get` script from
`CURL_AUTH='-H "api-secret: ${API_SECRET}"'`
to
`CURL_AUTH="-H api-secret:${API_SECRET}"`
for this to work. As mentioned in the previous comment, `API_SECRET` needs to be set to the _hashed_ API_SECRET, which can be generated using
`nightscout hash-api-secret $UNHASHED_API_SECRET`

mountrcg:
In oref0-autotune.sh I have tried to set environment variable API_SECRET using the hash or the token:
```
API_SECRET=32ea1df04----------------e84f2750
API_SECRET=token=autotune-5f--------a8
```
both don't work on a Nightscout site with `AUTH_DEFAULT_ROLES` set to `denied`. Any ideas how to get it to work on a Mac?

mountrcg:
found solution. had to specify it in osx environment variables
`export API_SECRET=token=autotune-5f--------a8`

Issue #299, opened 2016-12-29
Remove need for environement variables such as NIGHTSCOUT_HOST and API_SECRET and use ns device


PieterGit:
@scottleibrand suggested to remove the environment variables for nightscout and only use the ns device.

Stuff to do:
- integrate ns device in oref0-setup.sh (done)
- remove environment variables uses in oref0-setup.sh 
- remove direct calls to `curl` (e.g. meals-and-carbs) and use `ns` device
- make sure there is only one place for nightscout host and api secret. That would be ( `ns.ini`). Currently it's in the crontab and in the ~/.bash_profile 
 remove nightscout host and api scret in ~/.profile (work in progress)
- ...

vincepf:
I'm not sure if I understand this properly, but in the current setup script here:
```bash
# configure ns
if [[ ! -z "$NIGHTSCOUT_HOST" && ! -z "$API_SECRET" ]]; then
    echo "Removing any existing ns device: "
    killall -g openaps 2>/dev/null; openaps device remove ns 2>/dev/null
    echo "Running nightscout autoconfigure-device-crud $NIGHTSCOUT_HOST $API_SECRET"
    nightscout autoconfigure-device-crud $NIGHTSCOUT_HOST $API_SECRET || die "Could not run nightscout autoconfigure-device-crud"
fi
```
doesn't the autoconfigure-device-crud set up use nightscout.sh, which here:
```bash
autoconfigure-device-crud)
  NIGHTSCOUT_HOST=$1
  PLAIN_NS_SECRET=$2
  API_SECRET=$($self hash-api-secret $2)
  case $1 in
    help|-h|--help)
      setup_help
      exit 0
      ;;
  esac
  # openaps device add ns-get host
  test -z "$NIGHTSCOUT_HOST" && setup_help && exit 1;
  test -z "$API_SECRET" && setup_help && exit 1;
  openaps device add ns process --require "oper" nightscout ns "NIGHTSCOUT_HOST" "API_SECRET"
  openaps device show ns --json | json \
    | json -e "this.extra.args = this.extra.args.replace(' NIGHTSCOUT_HOST ', ' $NIGHTSCOUT_HOST ')" \
    | json -e "this.extra.args = this.extra.args.replace(' API_SECRET', ' $API_SECRET')" \
    | openaps import
  ;;
```

sets up the `ns` device itself as simply a process to use `nightscout.sh`?

scottleibrand:
[...]

PR #407, opened 2017-02-17
Updated to use API_SECRET_READ for the api key needed to read from

beached:
This patch will allow oref-autotune.sh to use the environment variable API_SECRET_READ.  This contains the unhashed API Secret needed to read from Nightscout when the default read permission is deny.


beached:
This is related to https://github.com/openaps/oref0/issues/397

scottleibrand:
Would it be better to use the `ns` device / `nightscout` script to use the secret already stored in ns.ini?

beached:
I am not sure, I have 1 reference to ns.ini in the exported-loop.json.  Maybe wrong branch?

scottleibrand:
More details for doing this in oref0 more generally are at https://github.com/openaps/oref0/issues/299.  I think we would want to do the same thing for autotune in order to inherit the ability of the `ns` device to use the ns.ini.  (We don't have to reference it directly.)

PR #408, opened 2017-02-19
Install missing jq dependency when --btpeb is set

moomoobloo:


Bender1061:
there was talk about making jq an install always. 
might be better to add it to 
https://github.com/openaps/docs/blob/master/scripts/quick-packages.sh
and 
https://github.com/openaps/docs/blob/master/scripts/quick-src.sh

moomoobloo:
@Bender1061 always installing jq sounds fine, but I don't see how adding it to those scripts in the docs will help. It should be installed as part of setup and AFAIK those scripts are not called from setup.

scottleibrand:
Those scripts are a prerequisite to oref0-setup, so that would be a good place to install jq. I'm not too worried about anyone who doesn't follow the OpenAPS install instructions, so as long as it's in one of the scripts we have people run, or added to one of the copy-paste lines, that's good enough for me. 

Scott

Issue #406, opened 2017-02-16
BT tethering and looping stops


Kdisimone:
I *think* the issue starts when the rig comes off her school wifi and onto her iPhone via BT tether.  Rig stops looping.  Won't restart looping unless a manual reboot is performed.  Tonight I manage to pull this off the rig after it had been stalled for about 4 or 5 hours while she was at track.  Rig was connected to her hotspot, on her body the whole time while at track.  The rig did not come off hotspot for our home wifi when she got home, and was not looping.  I was able to pull this off the pump log

![img_3473](https://cloud.githubusercontent.com/assets/12402160/23005879/81e6c650-f3b3-11e6-91b3-296e7963549d.PNG)

Then I rebooted the rig and just before the pump started rolling again...i copied this off...

```
Starting pump-loop at Wed Feb 15 15:52:15 PST 2017:
Traceback (most recent call last):
  File "/usr/local/bin/mmeowlink-any-pump-comms.py", line 15, in <module>
    app.run(None)
  File "/usr/local/lib/python2.7/dist-packages/decocare/helpers/cli.py", line 113, in run
    self.prelude(args)
  File "/usr/local/lib/python2.7/dist-packages/mmeowlink/cli/any_pump_comms_app.py", line 28, in prelude
    super(AnyPumpCommsApp, self).prelude(args)
  File "/usr/local/lib/python2.7/dist-packages/mmeowlink/cli/base_mmeowlink_app.py", line 26, in prelude
    self.link = link = LinkBuilder().build(args.radio_type, port)
  File "/usr/local/lib/python2.7/dist-packages/mmeowlink/link_builder.py", line 16, in build
    return SubgRfspyLink(port)
  File "/usr/local/lib/python2.7/dist-packages/mmeowlink/vendors/subg_rfspy_link.py", line 55, in __init__
    self.open()
  File "/usr/local/lib/python2.7/dist-packages/mmeowlink/vendors/serial_interface.py", line 23, in open
    raise AlreadyInUseException("%s already in use" % self.device)
mmeowlink.vendors.serial_interface.AlreadyInUseException: /dev/spidev5.1 already in use
Starting pump-loop at Wed Feb 15 16:20:16 PST 2017:
Traceback (most recent call last):
  File "/usr/local/bin/mmeowlink-any-pump-comms.py", line 15, in <module>
    app.run(None)
  File "/usr/local/lib/python2.7/dist-packages/decocare/helpers/cli.py", line 113, in run
    self.prelude(args)
[...]

PR #409, opened 2017-02-20
Spidev already in use reboot

Kdisimone:
I'm not entirely sure that this change is entered correctly...trying to resolve the "spidev already in use" error by adding a crontab line.  PLEASE take a look at this carefully as I'm a complete noob and basically was working off existing syntaxes to help me add the line to the script

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -686,6 +686,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
     fi
     if [[ "$ttyport" =~ "spi" ]]; then
         (crontab -l; crontab -l | grep -q "reset_spi_serial.py" || echo "@reboot reset_spi_serial.py") | crontab -
+	(crontab -l; crontab -l | grep -q "reboot spidev error" || echo "*/10 * * * * if (tail -n 50 /var/log/openaps/pump-loop.log | grep "/dev/spidev5.1 already in use"); then reboot; fi") | crontab -
     fi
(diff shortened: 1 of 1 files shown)


Bender1061:
are you sure you want to use 50 lines? I know you have it waiting for 10 mins, but matters how long your pump loop is. With the Enlite people, it can take about 7 mins to complete a loop, so I would be a bit more comfortable with 30 lines myself.

Kdisimone:
I'm totally open to suggestion....I'm definitely not an expert on what would work best here for the greatest good.

moomoobloo:
I'm not sure what the context for this is, but why would we want to reboot the device just because spidev is already in use?

Getting that error is normal and expected in many situations. For instance, any time a pump command is manually invoked, it's expected that the loop would produce the 'already in use' error, and I don't think it would be helpful to reboot in that situation.

Also, I have similar concerns to @Bender1061 that triggering reboot based on log output is not very robust. It seems to create the risk of infinite reboot loops or at least a series of several consecutive reboots if the error is not cleared from the logs in time. 

Is there a more specific condition that can be detected instead of the 'already in use' error? Is there a way to detect it independent of the log output?

Kdisimone:
It is related to this issue https://github.com/openaps/oref0/issues/406

Kdisimone:
I see now that @scottleibrand commented this cron line won't work well as the solution...I'd missed that prior to submitting this.  I will close this out...but the issue still needs resolution.  If one of you can suggest a better way and make a PR, that would be much appreciated.

Kdisimone:
I have been using a different solution currently, but do not know how to PR it in.  I do know however, that the BT tethering definitely hangs us up everyday that my daughter goes from the car ride (BT use) to school (school wifi).  It will not automatically hop off the BT and get on school wifi.  It gets stuck with this error and only a reboot will solve the issue.

PR #411, opened 2017-02-22
spidev5.1 already in use reboot

scottleibrand:
Sometimes, particularly after oref0-online restarts networking and/or switches back and forth between Bluetooth and wifi lots of times, we'll see the issue described at https://github.com/openaps/oref0/issues/406#issuecomment-281543469 and will need to reboot to fix.  This PR attempts to safely automate rebooting only in that situation.


scottleibrand:
This has been working well for us.  Merging it to dev for further testing there.

PR #413, opened 2017-03-02
Delete Future Data for better recovery of Time on System during No network power on for edison

Bender1061:
This runs a bash Script when the system reboots that checks the Date on the last CGM entry in monitor/glucose to see if that is in the future of the current time.
This is only an issue with the network was not available and the clock was not automatically set.
If the date of the rig is still in 1999 and it sees that the time is in the future. 
all the other checks in the Loop only look to see if the time is older than current system time.
so if you have a monitor/clock-zoned.json that time will not always be current when the oref0-set-system-clock is run, it will either not set the clock correctly, or the time that is there is from when the rig was last ran.
so if the time is in the future of the rig, it will delete the contents of the monitor folder, allowing the rig to fully grab all data from the pump fresh. 
I found that I had a ~correct time set within two loops of the  pump-loop and I was setting Temp basals within 3 loops.

### DIFF
--- a/bin/oref0-delete-future-entries.sh
+++ b/bin/oref0-delete-future-entries.sh
@@ -1,0 +1,22 @@
+#!/bin/bash 
+ 
+# Since the Edison does not have an internal clock 
+# after a reboot if you do not have the correct time your date will be older than  
+# the time on the CGM data in the monitor folder.  
+# Currently all the checks for time in the system is looking at data being older than current  
+# time 
+# This will delete all the data in the  
+# monitor directory. Allowing oref0 to gather all of the data again in pump-loop 
+
+NEWTIME=$(date -d `(jq .[1].display_time monitor/glucose.json; echo ) | sed 's/"//g'` +%s)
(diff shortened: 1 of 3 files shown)


bin/oref0-delete-future-entries.sh
```diff
@@ -0,0 +1,22 @@
...
+ 
+# Since the Edison does not have an internal clock 
+# after a reboot if you do not have the correct time your date will be older than  
+# the time on the CGM data in the monitor folder.  
+# Currently all the checks for time in the system is looking at data being older than current  
+# time 
+# This will delete all the data in the  
+# monitor directory. Allowing oref0 to gather all of the data again in pump-loop 
+
+NEWTIME=$(date -d `(jq .[1].display_time monitor/glucose.json; echo ) | sed 's/"//g'` +%s)
```
scottleibrand: I believe we need to install jq by default before this will work for everyone.

Bender1061: I'll go make that PR in docs

PR #417, opened 2017-03-09
add unannounced meal detection to autotune

scottleibrand:
Need some testing to determine if this is a reasonable approach, and whether `iob.iob > currentBasal` is a good IOB threshold for triggering UAM detection, or if it should be higher (maybe `2 * currentBasal`?)


scottleibrand:
I think currentBasal is the right threshold for now.  Merging to dev for testing.

PR #426, opened 2017-03-17
limit ns-meal-carbs to 6 hours

PieterGit:
Found out ns-meal-carbs asks 4 days of carb data from Nightscout each call...  That should be limited to ~ 6 hours (max theoretically carb's can live) similar to ns-temp-targets

as discussed on https://gitter.im/nightscout/intend-to-bolus?at=58cc08da1c040b8e0409465a and https://gitter.im/nightscout/intend-to-bolus?at=58cc194aa84f611959b6e668

### DIFF
--- a/lib/oref0-setup/alias.json
+++ b/lib/oref0-setup/alias.json
@@ -57,7 +57,7 @@
   },
   {
     "ns-meal-carbs": {
-      "command": "! bash -c \"curl -m 30 -s \\\"$NIGHTSCOUT_HOST/api/v1/treatments.json?find\\[carbs\\]\\[\\$exists\\]=true\\\" > monitor/carbhistory.json; oref0-meal monitor/pumphistory-zoned.json settings/profile.json monitor/clock-zoned.json monitor/carbhistory.json monitor/glucose.json settings/basal_profile.json > monitor/meal.json.new; grep -q COB monitor/meal.json.new && mv monitor/meal.json.new monitor/meal.json; exit 0\""
+      "command": "! bash -c \"curl -m 30 -s \\\"$NIGHTSCOUT_HOST/api/v1/treatments.json?find\\[created_at\\]\[\\$gte\]=$(date -d \\"6 hours ago\\" -Iminutes -u)&find\\[carbs\\]\\[\\$exists\\]=true\\\" > monitor/carbhistory.json; oref0-meal monitor/pumphistory-zoned.json settings/profile.json monitor/clock-zoned.json monitor/carbhistory.json monitor/glucose.json settings/basal_profile.json > monitor/meal.json.new; grep -q COB monitor/meal.json.new && mv monitor/meal.json.new monitor/meal.json; exit 0\""
     },
     "type": "alias",
     "name": "ns-meal-carbs"


scottleibrand:
We'll need to confirm whether this works for our Aussie and Kiwi friends if their treatments are in local time instead of UTC.  I think the "gte" comparison is on the raw string, not the actual date.

A better long-term fix would probably be to check the contents of the carbhistory.json when using it in determine-basal...

scottleibrand:
Based on @jmcrawford and @jelcrawford's testing, I'm thinking about merging this until we have a more complete fix.  Any objections?

thebookins:
Unless I'm mistaken, we would want to use `date -d \\"6 hours ago\\" -Iminutes` for this query (i.e. without the `-u` option for UTC). MongoDB uses a simple string compare on the date, so they both need to be in the same timezone. Treatments are stored in NS with `created_at` in local time, not UTC.

PieterGit:
I copied the `-u` from this commit: https://github.com/openaps/oref0/commit/1ac8cb83e4dfe9de61eaede15ddb3a081a4dd4ea

PieterGit:
@thebookins @scottleibrand 

Did some further debugging into the issue with RoboMongo. Here are the queries I used:
`db.getCollection('treatments').find({carbs: { '$exist' }, created_at : { '$gte': '2017-03-18'}})`
showed indeed that carbs use a `created_at` in localtime, e.g. `2017-03-17T07:48:59+01:00`

Temp targets use `created_at` in zulu/utc time, e.g. `2017-03-17T06:24:47.472Z`. When I used the query:
`db.getCollection('treatments').find({eventType: { '$regex' : 'Target'}, created_at : { '$gte': '2017-03-17'}})` 

I noticed that the carbs are created with `mm://openaps/mm-format-ns-treatments/Meal Bolus`, wheres the temp targets are set in NS.

So I think this is an oref0 bug and we should
change `bin/mm-format-ns-treatments.sh` or `bin/oref0-normalize-temps.js` so that `created_at` is converted to UTC before uploading it to NS.

PieterGit:
I think the `created_at` in `mm-format-ns-treatments.sh` must be converted to UTC time. I can get the date/time converted, but am unable to integrated in to the mm-format-ns-treatments.sh. Help is appreciated.

The conversion can be done with
` date --utc +%Y-%m-%dT%H:%M:%SZ --date "2017-03-17T07:48:59+01:00"`
where the date must be taken from `created_at` and stored in `created_at`

jasoncalabrese:
I've been wondering if NS `treatments` should start storing a `date` field like `entries`, we could set the field for new uploads if it's missing, only downside would be some index overhead

sulkaharo:
The amount of data in treatments is so small, a date index shouldn't be an issue. OTOH as discussed earlier, we should delete most of the indexes in the database that point to fields that are practically never used for querying and which contain a very large amount of entries. Current index set is:

```
api.indexedFields = [
    'created_at'
    , 'eventType'
    , 'insulin'
    , 'carbs'
    , 'glucose'
    , 'enteredBy'
    , 'boluscalc.foods._id'
    , 'notes'
  ];
```
Of which I'm not sure we actually should have anything but created_at (possibly a combined index of created_at with eventType). I'm near 100% certain the query optimizer cannot use the other indexes ever and/or we don't actually query by them and maintaining this amount of indexes is very heavy on the DB.

thebookins:
On the issue of `created_at` being in local time for treatments, I have found that my NS site reports do not show any OpenAPS treatments after 1 pm, while treatments uploaded by Loop are fine. I'm at UTC +11, which strongly hints that the date boundaries are getting messed up. See https://gitter.im/nightscout/public?at=58d0f8ae8fa1633954fe4384.

I suspect that converting `created_at` to be in UTC time would fix this issue too.

scottleibrand:
This should no longer be needed now that we're merging pumphistory.json and pumphistory-24h.json

PR #427, opened 2017-03-20
only consider carbs from last DIA hours as a workaround for undead carbs

scottleibrand:
Currently AMA considers carbs up to 1.5*DIA hours old, but the pumphistory-zoned.json file only has 5 hours of insulin history.  As a result, when boluses age off at the 5h mark, carbs associated with them are decayed more slowly (often at min_5m_carbimpact), resulting in COB "coming back from the dead".  The longer-term solution to this is probably to merge the pumphistory-zoned.json with the pumphistory-24h-zoned.json so we can properly decay carbs back to 1.5*DIA, but until we do that, it would be safer to exclude carbs from more than DIA hours ago (typically 3-4h).

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -218,7 +218,7 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     // CI = current carb impact on BG in mg/dL/5m
     ci = round((minDelta - bgi),1);
     if (meal_data.mealCOB * 3 > meal_data.carbs) {
-        // set ci to a minimum of 3mg/dL/5m (default) if at least 1/3 of carbs from the last 1.5*DIA hours are still unabsorbed
+        // set ci to a minimum of 3mg/dL/5m (default) if at least 1/3 of carbs from the last DIA hours are still unabsorbed
         ci = Math.max(profile.min_5m_carbimpact, ci);
     }
     aci = 10;
--- a/lib/meal/total.js
+++ b/lib/meal/total.js
@@ -37,7 +37,7 @@ function diaCarbs(opts, time) {
 
     treatments.forEach(function(treatment) {
         var now = time.getTime();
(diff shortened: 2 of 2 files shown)


tim2000s:
I think that's a reasonable work around although I wonder how that will affect AMA with "killer" meals like pizza. 

I'm not sure how long this issue has been around for but I don't recall seeing it up until a couple of releases back. 

What might help is a "decayed carbs" variable that ama can write to so that you retain progress on slow meals which may be lost in this model?

PieterGit:
@scottleibrand I don't understand the factor 3 in line 201 in `determine-basal.js` https://github.com/openaps/oref0/blob/b10b1943e19dc8caba71753447efd6e5ce00f9b4/lib/determine-basal/determine-basal.js#L220 `if (meal_data.mealCOB * 3 > meal_data.carbs)`.  The comment says `if at least 1/3 of carbs from the last DIA hours are still unabsorbed`, than I would expect: `meal_data.carbs/3 > meal_data.mealCOB`. Can you explain the if-statement condition?

I notice you use quite some assumptions on the values of the configurable parameters (e.g. DIA typically 3-4h) that aren't checked in the code. Shouldn't the range of the parameters be checked, because out-of-range might cause strange/unsafe conditions in the code.

scottleibrand:
What I have is mathematically equivalent to `if (meal_data.mealCOB > meal_data.carbs / 3)`, which translates to English as if at least 1/3 of carbs from the last DIA hours are still unabsorbed.  Carbs is the total, and mealCOB is the remainder (COB).

What would you propose we do if someone chooses a non-standard DIA?  Refuse to run?  Obviously we need to implement a long-term fix, like using the full pumphistory-24h-zoned.json merged with the most recent pumphistory-zoned.json, to fix this bug for non-standard DIA as well.  But we've been operating for months now with this bug affecting standard DIAs, and it hasn't been a major safety issue.

PR #429, opened 2017-03-25
Save battery

scottleibrand:
When the Edison battery level is below 60%, this makes the rig stop running autosens every 20m, and instead run it every 90m.  It also causes oref0-pump-loop to wait up to 5 minutes for a BG reading newer than the most recent temp basal reading before even trying to use the radio or talk to the pump.  In combination with a new subg_rfspy firmware that I'm testing with @mogar, this should allow us to power down the cc1110 radio most of the time when necessary.

For now, I'm leaving the low battery cutoff at 60%, as the charging voltage reads at about 62-65%.  After some testing, we might want to make this the default mode whenever the rig is not actually plugged in (when battery reads >65% as well as <60%).

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -29,6 +29,7 @@ smb_main() {
     until ( \
         prep
         echo && echo Starting supermicrobolus pump-loop at $(date) with $upto30s second wait_for_silence: \
+        && low_battery_wait \
         && wait_for_silence $upto30s \
         && preflight \
         && refresh_old_pumphistory \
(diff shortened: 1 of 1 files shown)


scottleibrand:
```
Settings refreshed
Edison battery low: 84%
Pumphistory-24 less than 90m old
Completed supermicrobolus pump-loop at Fri Mar 24 23:46:30 MST 2017:


Starting supermicrobolus pump-loop at Fri Mar 24 23:47:02 MST 2017 with 15 second wait_for_silence:
Edison battery low: 85%; waiting up to 5 minutes for new BG: .............glucose.json newer than temp_basal.json
Radio ok, Listening: ...........No pump comms detected from other rigs
Preflight OK, Profile less than 60m old
```

scottleibrand:
One thing I'll want to test is whether this causes retries to take too long due to low_battery_wait being called inside the until() loop.  If so, we might want to move it to run prior to that.

ceben80:
@scottleibrand typo 1st sentence above, you mean autosens insteed of autotune :)

scottleibrand:
Yep, thx.  Edited.

PieterGit:
Please make it so that the battery level is not required. From a quick review it looks like you implemented it that way. The kernel 4.x on Edison does not have support for batterylevel yet, see https://github.com/andy-shev/linux/issues/9

scottleibrand:
This code defaults to current behavior if it can't read the battery level. Is that what we want?

PR #435, opened 2017-04-02
Implement BlueZ min version checking.

applehat:
Now checks for at least version 5.37, and will install 5.44 if no new enough version found.

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -417,14 +417,18 @@ done
 echo Checking for BT Mac, BT Peb or Shareble
 if [[ ! -z "$BT_PEB" || ! -z "$BT_MAC" || ${CGM,,} =~ "shareble" ]]; then
     # Install Bluez for BT Tethering
-    echo Checking bluez installation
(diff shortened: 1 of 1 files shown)


PieterGit:
I thought I read somewhere that the ` --enable-experimental` can be skipped for newer bluez versions, but I can't find it anymore. The changelog lists that from 5.42+ `The major API change with this release is that the GATT D-Bus API is no longer marked as experimental`.  I think the GATT D-Bus is required for all the GATT Low Energy (LE) profiles, but I can't find if all the LE profiles are non experimental now.

scottleibrand:
@pietergit Any harm in leaving --enable-experimental for 5.44?

This just avoids changing anything for users with a working 5.37 install, so I'm inclined to merge and do anything else as another PR, unless you see any problems with it.

PieterGit:
i don't think there is any harm in leaving `--enable-experimental`. I think we would like to have all stable (non experimental) versions of third party products when there available. But I agree that can be handled as another PR.
BTW: standard raspberry raspbian has 5.23 (as far as I know). I noticed a lot of people are also upgrading that. But we might also need something to upgrade bluez on a rpi.

PR #436, opened 2017-04-04
set oref0-runagain.sh as executable after move.

applehat:
I've had to manually set it to +x every time, so this will fix that for anyone with the issue (and not cause problems for those without)


scottleibrand:
Or you could just follow the directions and add `bash ` in front. 😜

PR #437, opened 2017-04-05
use compressed curl calls

mgranberry:
This saves me significant metered data when I'm tethered.  It also seems to help when I'm out of the LTE coverage area.

### DIFF
--- a/bin/oref0-upload-profile.js
+++ b/bin/oref0-upload-profile.js
@@ -60,7 +60,7 @@ if (!module.parent) {
 
     if (apisecret.length != 40) {
         var shasum = crypto.createHash('sha1');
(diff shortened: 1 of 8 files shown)


bin/oref0-upload-profile.js
```diff
@@ -60,7 +60,7 @@ if (!module.parent) {
 
     if (apisecret.length != 40) {
         var shasum = crypto.createHash('sha1');
-        shasum.update(apisecret);
+        shasum.update(String(apisecret));
```
mgranberry: I accidentally pulled this in, but it fixes a bug that occurs when a user has a numeric Nightscout authentication code.

PR #438, opened 2017-04-05
wait on glucose data that is newer than last action

mgranberry:
This also moves the responsibility for backdating the
monitor/glucose.json to the various actions that
are creating the file.  I didn't find anything else in this
repository, but it may also depend on changes to the xdrip
or Medtronic CGM sources.

I don't feel strongly about moving the timestamp-updating out
of the script, but it takes several seconds to run and doesn't 
need to be done unless `glucose.json` is modified.

### DIFF
--- a/bin/mm-format-ns-pump-history.sh
+++ b/bin/mm-format-ns-pump-history.sh
@@ -24,11 +24,10 @@ OUTPUT=${2-/dev/fd/1}
 #TZ=${3-$(date +%z)}
 
 cat $HISTORY | \
-  json -e "this.medtronic = this._type;" | \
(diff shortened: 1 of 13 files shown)


scottleibrand:
I'd rather not move anything new into openaps aliases.  It's much more readable and manageable if it's in a proper shell script instead.

mgranberry:
I made the change.  I've been running the un-reverted version for a week without noticing anything wrong.

scottleibrand:
Thanks. It looked like it would work, but it requires people to re-run oref0-setup to upgrade, which makes things hard to troubleshoot.

PR #442, opened 2017-04-11
don't do redundant JSON parsing

mgranberry:
This cleans up a number of chained `json` calls to execute all operations in a single execution.

### DIFF
--- a/bin/mm-format-ns-pump-history.sh
+++ b/bin/mm-format-ns-pump-history.sh
@@ -24,11 +24,10 @@ OUTPUT=${2-/dev/fd/1}
 #TZ=${3-$(date +%z)}
 
 cat $HISTORY | \
-  json -e "this.medtronic = this._type;" | \
-  #json -e "this.dateString = this.timestamp + '$(TZ=TZ date +%z)'" | \
-  json -e "this.dateString = this.timestamp + '$(date +%z)'" | \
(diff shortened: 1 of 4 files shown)


scottleibrand:
Have you tested this?  I haven't yet, but don't see any indication from `man json` that `json -e "somecode;" -e "someothercode;"` will execute both `somecode` and `someothercode`.

scottleibrand:
If not, we might want something like:

```
json -e "this.medtronic = this._type; \
    this.dateString = this.timestamp + '$(date +%z)';  \
    this.type = 'medtronic';"
etc.
```

mgranberry:
It works for me.  From the `json` source: 
```js
        case '-e':     
        case '-E':  // DEPRECATED in v9
            parsed.exeSnippets.push(args.shift());
            break;
...
     for (i = 0; i < opts.exeSnippets.length; i++) {
            exeFuncs[i] = new Function(opts.exeSnippets[i]);
        }   
```

scottleibrand:
Yay for undocumented features.  :-)

Would putting all the json into a single multiline quoted string be more readable or better for any other reason?  If not, I'm fine with testing your approach.

mgranberry:
I don't think it makes a big difference either way.  I'd be happy to change it to multiline strings if you have a preference for them.

PR #446, opened 2017-04-23
Syntax Error line 788

rkresha:
Erroronous `fi` at line 559. Pulled `fi` from 559 and removed comment on
correct `fi` at line 788. Also manual pretty printer because of mixed number of tab spaces on structure and general housekeeping.


rkresha:
Hold on this... still working through some issues

PR #449, opened 2017-04-28
fix logic to do cgm-loop stuff for g4-upload, not g4-local

scottleibrand:


PieterGit:
@scottleibrand +1 on this one. I couldn't get my cgm loop uploading without a similar fix.  Can you merge this to dev and adj-target-when-high branch?

scottleibrand:
Done.

PR #450, opened 2017-04-29
autotune carb_ratio null check

scottleibrand:
If an autotune run fails in such a way as to write a null carb_ratio, this can prevent all future autotune runs from working, and entirely prevent looping.  This allows autotune to recover from such a situation by using the profile.pump.json if the profile.json has a null carb_ratio.  Additional null checks (especially when writing a new profile.json) might also be useful.


scottleibrand:

PR #455, opened 2017-05-02
Blend UAM and COB properly

scottleibrand:
When UAM is enabled and COB is predicting a large rise, the current algorithm in floating-carbs is a bit too conservative, only SMBing properly when both COB and UAM predict a rise, and less than it should if there's lot of IOB.

This tweak uses higher of minCOBPredBG and the existing two (minIOBPredBG, blendedMinPredBG) to allow for proper SMB'ing whenever COB predicts a rise, but still limiting the size of SMBs allowed by UAM.  It also adds another safety check to make sure that minPredBG isn't higher than avgPredBG.

Overall, this should preserve most of the beneficial effects of the "don't SMB too much for rescue carbs fix" while eliminating some of the side effects of that (particularly where it would SMB too little too late after zero-temping to avoid a post-meal low, resulting in a post-meal high).

We'll want to test this in a rescue carbs situation (with UAM enabled, and rescue carbs entered) to confirm whether it behaves appropriately there.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ all: test
 report:
 	# report results to community
 test:
-	./node_modules/.bin/mocha ${TESTS}
+	./node_modules/.bin/mocha -c ${TESTS}
 
 travis:
 	${ISTANBUL} cover ${MOCHA} --include-all-sources true --report lcovonly -- -R tap ${TESTS}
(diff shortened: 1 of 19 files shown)


scottleibrand:
We've been testing this successfully, and found it to be an improvement over the old behavior currently in dev.  If no objections, I'd like to go ahead and merge it to dev.

PR #458, opened 2017-05-07
Pushover support for carbsReq and insulinReq notifications

scottleibrand:
Most significant changes in this PR, in order of appearance in the diff:

- oref0-online updates for stability
- oref0-set-device-clocks if pump clock is >1m off
- oref0-set-system-clock if determine-basal complains it's set incorrectly
- support for oref1 pushover alerts of insulinReq and carbsReq
- allow non-SMB code paths to set temps up to 120 minutes when needed
- allow non-zero temps in SMB mode when lastCOBpredBG > target_bg
- don't SMB if basal is < 0.3 U/hr (if 0.1U SMB represents >20m of basal)

### DIFF
--- a/bin/oref0-online.sh
+++ b/bin/oref0-online.sh
@@ -5,7 +5,7 @@ ip -4 -o addr show dev wlan0 | awk '{split($4,a,"/");print a[1]}'
 ip -4 -o addr show dev bnep0 | awk '{split($4,a,"/");print a[1]}'
 echo
 echo -n "At $(date), my wifi network name is "
-printf '%s' $(iwgetid -r)
+iwgetid -r wlan0 | tr -d '\n'
 echo -n ", and my public IP is: "
 if curl --compressed -4 -s -m 15 icanhazip.com | awk -F , '{print $NF}' | egrep "^[12]*[0-9]*[0-9]\.[12]*[0-9]*[0-9]\.[12]*[0-9]*[0-9]\.[12]*[0-9]*[0-9]$"; then
     # if we are back on wifi (and have connectivity to icanhazip.com), shut down bluetooth
--- a/bin/oref0-online.sh
+++ b/bin/oref0-online.sh
@@ -1,5 +1,11 @@
 #!/bin/bash
 echo; echo Starting oref0-online.
+if iwgetid -r wlan0; then
+    if ! ip route | grep default | grep -q wlan0; then
+        echo Attempting to renew wlan0 IP
+        sudo dhclient wlan0
+    fi
+fi
 echo -n "At $(date) my local IP is: "
 ip -4 -o addr show dev wlan0 | awk '{split($4,a,"/");print a[1]}'
 ip -4 -o addr show dev bnep0 | awk '{split($4,a,"/");print a[1]}'
@@ -33,7 +39,7 @@ else
     #sudo dhclient wlan0
     echo
     echo -n "At $(date), my wifi network name is "
-    printf '%s' $(iwgetid -r)
+    iwgetid -r wlan0 | tr -d '\n'
     echo -n ", and my public IP is: "
     # loop over as many MACs as are provided as arguments
     if ! curl --compressed -4 -s -m 15 icanhazip.com | awk -F , '{print $NF}' | egrep "^[12]*[0-9]*[0-9]\.[12]*[0-9]*[0-9]\.[12]*[0-9]*[0-9]\.[12]*[0-9]*[0-9]$"; then
@@ -58,7 +64,7 @@ else
         echo
     fi
     echo -n "At $(date), my wifi network name is "
-    printf '%s' $(iwgetid -r)
+    iwgetid -r wlan0 | tr -d '\n'
     echo -n ", and my public IP is: "
     curl --compressed -4 -s -m 15 icanhazip.com
 fi
--- a/bin/oref0-online.sh
+++ b/bin/oref0-online.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 echo; echo Starting oref0-online.
 # if we are connected to wifi but don't have an IP, try to get one
-if iwgetid -r wlan0 | egrep -q "[A-Z]|[a-z]|[0-9]*"; then
+if iwgetid -r wlan0 | egrep -q "[A-Z]|[a-z]|[0-9]?"; then
     if ! ip route | grep default | grep -q wlan0; then
         echo Attempting to renew wlan0 IP
         sudo dhclient wlan0
--- a/bin/oref0-online.sh
+++ b/bin/oref0-online.sh
@@ -22,15 +22,15 @@ if curl --compressed -4 -s -m 15 icanhazip.com | egrep "^[12]*[0-9]*[0-9]\.[12]*
         sudo dhclient wlan0
     fi
 else
-    echo; echo "Error, cycling networking "
+    #echo; echo "Error, cycling networking "
     # simply restart networking completely for stability purposes
-    sudo /etc/init.d/networking stop
-    sleep 5
-    sudo /etc/init.d/networking start
-    echo "and getting new wlan0 IP"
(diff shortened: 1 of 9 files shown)


scottleibrand:
Anything else we need here before merging this to dev?

danamlewis:
Since we just made a bunch of other changes, let's get either @tim200s or @kdisimone to run & confirm this doesn't break anything for them? Also please wait for us to get some prelim testing instructions/notes up somewhere before talking about this  more broadly/asking others to test.

babybluesky1:
Just installed, and it's working well.  Put in on several rigs without issues.  One small typo that I haven't gone through to look for yet is in the reason field.  Missing a space between target BG and the time left on temp basal.  for example `IOBpredBG 69; Eventual BG 71 < 9025m left and 0 ~ req 0U/hr`  haven't configured any pushover/ifttt yet until i figure the body texts

Kdisimone:
Just installed, and it's working well. Put in on several rigs without issues. One small typo that I haven't gone through to look for yet is in the reason field. Missing a space between target BG and the time left on temp basal. for example IOBpredBG 69; Eventual BG 71 < 9025m left and 0 ~ req 0U/hr haven't configured any pushover/ifttt yet until i figure the body texts

tim2000s:
Can confirm that I've been running the pushover details for a week or so and those aspects work as expected (i.e. Very well)

jbwittmer:
I've been running an installation of this branch this evening with no problems so far.  Am I correct that to enable the pushover function I need to re-run the script and append -pu=  and -pt=  and insert my pushover user and token respectively?  Have I missed anything else I need to do to enable the pushover capability?

scottleibrand:
That should be it.

bin/oref0-online.sh
```diff
@@ -1,13 +1,20 @@
 #!/bin/bash
 echo; echo Starting oref0-online.
+# if we are connected to wifi but don't have an IP, try to get one
+if iwgetid -r wlan0 | egrep -q "[A-Z]|[a-z]|[0-9]?"; then
```
PieterGit: The `egrep`  can be simplified and should allow all characters that are allowed in SSID's
I think you mean `+` (more than one)  instead of `?` (zero or 1) instance
For example I think it's cleaner to use `iwgetid -r wlan0 | egrep -q "[A-Za-z0-9_]+"`
Note: http://routersecurity.org/SSID.php says "Special characters (spaces, periods, dashes, underscores etc) are allowed. Its probably best to avoid the pipe character (vertical line). And, any use of a special character may be asking for trouble."

bin/oref0-online.sh
```diff
@@ -1,13 +1,20 @@
...
 echo -n "At $(date) my local IP is: "
 ip -4 -o addr show dev wlan0 | awk '{split($4,a,"/");print a[1]}'
 ip -4 -o addr show dev bnep0 | awk '{split($4,a,"/");print a[1]}'
 echo
 echo -n "At $(date), my wifi network name is "
-printf '%s' $(iwgetid -r)
+iwgetid -r wlan0 | tr -d '\n'
 echo -n ", and my public IP is: "
-if curl --compressed -4 -s -m 15 icanhazip.com | egrep "^[12]*[0-9]*[0-9]\.[12]*[0-9]*[0-9]\.[12]*[0-9]*[0-9]\.[12]*[0-9]*[0-9]$"; then
+if curl --compressed -4 -s -m 15 icanhazip.com | awk -F , '{print $NF}' | egrep "^[12]*[0-9]*[0-9]\.[12]*[0-9]*[0-9]\.[12]*[0-9]*[0-9]\.[12]*[0-9]*[0-9]$"; then
```
PieterGit: What does the `awk` statement do? I've seen false positives on restarting the wifi on my home network.
Perhaps upping `-m 15` to `-m 30` will also make it more robust, but I don't know what the response time of icanhazip.com is.

bin/oref0-pushover.sh
```diff
@@ -0,0 +1,25 @@
...
+
+date
+if find monitor/ -mmin -$SNOOZE | grep -q pushover-sent; then
+    echo "Last pushover sent less than $SNOOZE minutes ago."
+elif ! find $FILE -mmin -$SNOOZE | grep -q $FILE; then
+    echo "$FILE more than $SNOOZE minutes old"
+elif ! cat $FILE | egrep "add'l|maxBolus"; then
+    echo "No additional carbs or bolus required."
+else
+    curl -s -F "token=$TOKEN" -F "user=$USER" -F "sound=$SOUND" -F "message=$(jq -c "{bg, tick, carbsReq, insulinReq, reason}|del(.[] | nulls)" $FILE) - $(hostname)" https://api.pushover.net/1/messages.json && touch monitor/pushover-sent
```
PieterGit: Can you explain what the jq command does here? 
I think the implementation is quite tight to the pushover implementation, and making it work for other messaging types will take quite some more work (e.g. sent a notification to pebble directly). I would like to have rigs that also work offline.

tim2000s:
I guess the question is what you're trying to achieve. You either have a script that generates an information file  (which is the basis of what jq is doing - I use it similarly  to create an HTML file) that you then implement a second mechanism for notification dependent on what you want to use, i.e. Pushover, Pancreabble, some other local push mechanism to a watch, etc and keep the two separate, which I think is what you're describing. 

Alternatively you can tie the data creation into the mechanism that you are using to create the notification so that the structure fits the data structure requirement of your mechanism. 

I think either works, and for now this should be left as is with a new feature request based off it to allow the first mechanism I described to be built. What do you think?

bin/oref0-pushover.sh
scottleibrand: In this case jq is just displaying the fields of interest for the pushover notification.  I agree with @tim2000s that the requirements of each notification method are different, so we don't necessarily need to reuse that part of the code.

bin/oref0-online.sh
scottleibrand: The units on the -m argument are in seconds.  If it takes more than 15 seconds to get a single-packet response from a webserver, then more than likely we'd have better luck giving up on that particular connection type and trying something new.

bin/oref0-online.sh
scottleibrand: I think the `?` is identical to `+` (or leaving it off completely) in this case, since we're just substring matching within the SSID.  That said, I think your regex is a bit cleaner, so we can use that.

jbwittmer:
Can confirm also running on Raspberry Pi3 with TI stick, autotune and microbolus and autosens enabled.  No problems.

scottleibrand:
Just pushed a potential fix for the oref0-online stuff.  If anyone's feeling adventurous, feel free to pull it and test.  I haven't put it on any of our rigs yet, as I'm about to get on a plane and won't be  with them for a bit.

tim2000s:
Installed and in test.

tim2000s:
New oref0-online installed this morning. Went through the same routine as yesterday. This time connectivity has been maintained throughout, so the fixes seem to work.

jbwittmer:
The install that I did on Raspberry Pi with Autotune ran last night, but it only pulled Nightscout data for 1 day.  Is that normal for a first run?  Sorry, I don't usually run Autotune as part of my loop so don't know what to expect for behavior.  Not surprisingly it made no recommendations for changes.

scottleibrand:
The oref0-online script seems to be working well now, so I think we're about ready to merge this.  @danamlewis also wanted to write up something for the docs to describe the pushover feature, what to expect, and discourage people from blindly bolusing off the recommendation.

danamlewis:
I added some notes about Pushover usage with SMB to the draft docs we've got going around oref1 (so not merged, but you can see them via https://github.com/openaps/docs/blob/ea422e7fc70bc810737d5225c495cfa524ecc767/docs/docs/walkthrough/phase-4/oref1-features.md#pushover-smb-and-openaps), so I am +1'ing on top of the other comments we've gotten on Gitter about this being ready for merging to dev.

PR #456, opened 2017-05-02
fix --ww_ti_usb_reset

PieterGit:
@ceben80 noticed that the --ww_ti_usb_reset had gone broken on the dev branch. 

@ceben80 can you test this on my PieterGit:notify branch, or do you need it to be on dev branch to give it a try

PR #470, opened 2017-05-16
Nightscout token based authentication and other improvements

PieterGit:
Add Nightscout token based authentication.  This makes it possible to set  yout Nightscout `AUTH_DEFAULT_ROLES` from `readable` to `denied` and still use openaps/oref0. 
See read proposed docs  at https://github.com/PieterGit/docs/blob/tokenbased/docs/docs/walkthrough/phase-1/nightscout-setup.md#switching-from-api_secret-to-token-based-authentication-for-your-rig

Other improvements: 
- fix --ww_ti_usb_reset, https://github.com/openaps/oref0/pull/456
- remove `--config` and `--noop` from ns-get.sh and ns-upload.sh
- remove oref0-find-ti.sh (not used anymore, implemented in oref0_init_pump_comms.py and oref0_subg_ww_radio_parameters.py
- remove pump-loop loop from oref0-pump-loop.sh. Cron is responsible of calling each iteration of the loop
- log Aborted and exit code if pump loop fails
- less mmtune in maybe_mmtune 
- convert ns-temptargets and ns-meal-carbs to use NS device instead of calling curl


PieterGit:
The syntax in the curl call in in the NS device was a bit different (e.g. no `-u`), so I thought it wasn't used actively. My mistake. I'll revert it. The API description in `nightscout.sh` is quite unreadable btw.

I did an install on an Explorerboard, and it installs fine. I'll work on getting the proxy working (the authentication token can already be fetch), and seeing if everything works.

scottleibrand:
I'd have to go look at exactly how, but I know we pull a full 24h from NS for autotune.

PieterGit:
No problem. Can you merge #456 to dev, so this pull reflect only will be token based authentication (and perhaps a bluez upgrade to 5.45 if I can't find any errors).

PieterGit:
Status update: I got it working. I don't need the nightscout proxy, so I'll remove that (it's a pity). I'll test with it on my test rig and propose to merge it to dev after the pushover branch has been merged to dev.

PieterGit:
@scottleibrand I think the requested changes tag can be removed because I followed your advice and rollbacked the changes you had comments on.

I have two questions concerning the pump loop (not really related to token based authentication, but it's on my testing branch now):

1) can you review my changes to the pump loop, https://github.com/PieterGit/oref0/commit/ac54d66c6efef52752673bc02734d73c0e10c46d

I had trouble with too much mmtunes with my WW pump, and had pump loops that stopped working (always some error, continuously maybe_mmtune). I changed:

- removed loop in main() and smb_main(). keeping the loop running is a
responsibility of cron. This makes it possible to stop the pump loop
with `/etc/init.d/cron stop` and will not let `killall` kill
oref0-pump-loops.
- write in the logs when a pump loop is aborted (because of an error of
one of the commands)
- descreased maybe_mmtune. A limit of 26214 will mmtune in ~ 44% time
which seems too much (don't understand why, but tested it with a shell script)

2) I don't understand https://github.com/openaps/oref0/blob/dev/bin/oref0-pump-loop.sh#L67-L72
Why is not `suspended=true` an error and a reason to `maybe_mmtune`

I'm planning to polish the stuff this weekend, and I think it's ready to merge to dev after that.

PieterGit:
i think this pull request is ready for broader testing... it installs fine (tested with manual entry of all the data and with single command line). The pump is out of the house right now, so I can't test any further now. @scottleibrand  can you do a (final) review. Others willing to give it a testdrive? E.g. @ceben80 and @betluis

scottleibrand:
With regard to https://github.com/openaps/oref0/blob/dev/bin/oref0-pump-loop.sh#L67-L72 - if we can successfully talk to the pump and find that it's suspended, that does *not* indicate a pump comms issue that would justify mmtune, but rather just means the user suspended their pump.  There's no point in running mmtune at that point: better to just wait for the user to resume the pump.

PieterGit:
Removed the maybe_mmtune if suspended is false.  Is that all the changes you requested? @scottleibrand I'll do some final testing on saturday, and if somebody else gets it running I think it's ready to merge to dev.

PieterGit:
Added some documentation, see
https://github.com/PieterGit/docs/blob/tokenbased/docs/docs/walkthrough/phase-1/nightscout-setup.md#switching-from-api_secret-to-token-based-authentication-for-your-rig

PieterGit:
Tested it on a rig, looks fine. Please help testing this so that it can be merged to dev.

PieterGit:
This branch is on par with `dev`. It runs fine on my main rig. If you wan't to test it without breaking your production rig you can use `AUTH_DEFAULT_ROLES=readable devicestatus-upload` first. That way only the test rig will use token authentication and all other rigs will still work. Once all the rigs are converted to this branch (or dev/master once it's integrate) you can set `AUTH_DEFAULT_ROLES=denied` to secure your Nightscout.

ceben80:
@PieterGit I have successfully tested `token based authentication` it's working for me, well documented in the above instructions how to implement

PieterGit:
@scottleibrand or @danamlewis 
Both @ceben80 and @betluis have tested this new code and the docs. They both run one rig with token based authentication.

Can you please merge this pull request to dev  and the docs pullrequest https://github.com/openaps/docs/pull/862 to master.

betluis:
I have tested this branch and actually running it in our current rig. So far everything ok

PieterGit:
Tested by three users for 11 days now. Works ok. @scottleibrand : can you review and merge to dev when you have more time available?

scottleibrand:
Yep, planning to do so when I'm back home from Austria.

bin/oref0-pump-loop.sh
```diff
@@ -3,76 +3,73 @@
...
+        echo "Aborted pump-loop with exitcode $? at $(date)" \
+        echo Waiting 45s and maybe mmtune \
+        sleep 45 && maybe_mmtune && sleep 5
+    fi
 }
 
 # main supermicrobolus loop
 smb_main() {
     prep
-    until ( \
```
scottleibrand: Can we easily pull this change (to remove the until() loop) out of this PR to consider it independently?

PieterGit: I've been running with this configuration for about a month now without problems. I can't see any reason why removing the until creates a problem. I can think of various reasons when the until creates problems. At least WW-pumps seem to be not happy of killing processes half way and I had situation were a battery reset and mmtune was needed to make sure pump communication was ok again.

All other loops run only once, and get initiated with a minute-based cron.

scottleibrand: How many rigs do you typically have running at once?

PieterGit: mostly once

scottleibrand: Ok.  The until() stuff is somewhat useful when coordinating multiple rigs.  It may not be necessary, but I want to test it before merging.  If you want to separate that change out as its own PR, we should be able to go ahead and merge the rest of this one right away.

PieterGit:
Synced to dev and readded the `until` loop at @scottleibrand 's request. Tested the new pump loop manual on a rig, and worked fine. Please merge to dev.

bin/oref0-pump-loop.sh
PieterGit: done, i (temporary) readdded the until loop

bin/oref0-pump-loop.sh
```diff
@@ -67,8 +67,6 @@ smb_main() {
         if grep -q '"suspended": true' monitor/status.json; then
             echo -n "Pump suspended; "
             smb_verify_status
-        else
-            echo Error, retrying && maybe_mmtune
```
scottleibrand: Hmm, I just noticed this.  Why are you completely disabling the ability to mmtune on failures with SMB?

scottleibrand:
Fixed that myself and merged it.

PieterGit:
thanks. had a busy weekend, so no time to debug what was wrong.

PR #472, opened 2017-05-18
Adding prompts to script

Kdisimone:
adding prompts to script for microbolus flag

adding prompts to script for pushover token and user key

editing oref0-runagain.sh to return pushover entries.

### DIFF
--- a/bin/oref0-online.sh
+++ b/bin/oref0-online.sh
@@ -5,7 +5,7 @@ ip -4 -o addr show dev wlan0 | awk '{split($4,a,"/");print a[1]}'
 ip -4 -o addr show dev bnep0 | awk '{split($4,a,"/");print a[1]}'
 echo
 echo -n "At $(date), my wifi network name is "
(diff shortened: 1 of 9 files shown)


Kdisimone:
PLEASE check @scottleibrand that i've done these correct.  I haven't yet tested them...but I will soon

Kdisimone:
tested...they all worked the way they should.

Kdisimone:
while I'm on this file...what do you if I change the script so that the --btmac is not needed to prompt the question a question "do you want BT tether y/n" and then enter BTMAC if it's a yes?  It would be cleaner in the docs if there was just one setup script line (vs one that has a little tag for the --btmac in order to prompt the question)  @danamlewis ?

Kdisimone:
Closing pull request because pushover got merged to dev, now my PR is a bit outdated.  I'll look at current dev and see if any of these are still useful with new dev version and redo a PR, if needed

PR #476, opened 2017-05-21
Fix low glucose suspend

scottleibrand:
One of the changes merged to dev from the pushover branch attempted to disable low glucose suspend mode when IOB was highly negative, but that change went too far and mostly disabled LGS.  This fixes LGS to only be disabled when IOB is already super negative and BG is rising faster than predicted (such as from a meal eaten after a long period of zero temping).  It also adds that information to the reason field when disabling LGS.


PieterGit:
Can you add test cases to prevent LGS from failing in the future?

scottleibrand:
I could, but with m the current unit test framework all unit tests are by their nature incomplete, and take a lot more effort to write and continually update every time we change anything than they save in end-to-end testing. If anyone want to volunteer the time to figure out a way to do more real-world situation-playback testing, that would be useful...

tepidjuice:
Would an diabetic simulator help with testing? That is a function that took insulin and food inputs and returned a glucose value? You could put openaps between the glucose output of the simulator and the insulin input.

Kdisimone:
We installed it last night. Stayed suspended with a significant amount of neg iob during night. Would momentarily start a temp basal with a climb, but as soon as BG leveled off or dropped, it suspended basals again (appropriately)

PR #478, opened 2017-05-23
Fix for MDT Divide by Zero Error

Bender1061:
Fixes all the jq lookups for the nonexistent time value for MDT monitor/glucose.json. Make it look at nightscout/glucose.json instead. 
Changed the order of when the nightscout/glucose.json report was called at the end of the MDT reformatting  instead of the beginning of the upload so that this time can be used. and still be accurate. 

this has the added benefit of fixing the super long upload times for CGM data that I was seeing as well.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -339,7 +339,11 @@ function refresh_old_profile {
 
 function refresh_smb_temp_and_enact {
     # set mtime of monitor/glucose.json to the time of its most recent glucose value
-    touch -d "$(date -R -d @$(jq .[0].date/1000 monitor/glucose.json))" monitor/glucose.json
+    if grep "MDT cgm" openaps.ini; then
+      touch -d "$(date -R -d @$(jq .[0].date/1000 nightscout/glucose.json))" monitor/glucose.json
+    else
+      touch -d "$(date -R -d @$(jq .[0].date/1000 monitor/glucose.json))" monitor/glucose.json
+    fi    
     if ( find monitor/ -newer monitor/pump_loop_completed | grep -q glucose.json ); then
         echo "glucose.json newer than pump_loop_completed. "
         smb_enact_temp
@@ -353,7 +357,11 @@ function refresh_smb_temp_and_enact {
 
 function refresh_temp_and_enact {
     # set mtime of monitor/glucose.json to the time of its most recent glucose value
-    touch -d "$(date -R -d @$(jq .[0].date/1000 monitor/glucose.json))" monitor/glucose.json
+    if grep "MDT cgm" openaps.ini; then
+      touch -d "$(date -R -d @$(jq .[0].date/1000 nightscout/glucose.json))" monitor/glucose.json
+    else
+      touch -d "$(date -R -d @$(jq .[0].date/1000 monitor/glucose.json))" monitor/glucose.json
+    fi
     # TODO: use pump_loop_completed logic as in refresh_smb_temp_and_enact
     if( (find monitor/ -newer monitor/temp_basal.json | grep -q glucose.json && echo -n "glucose.json newer than temp_basal.json. " ) \
         || (! find monitor/ -mmin -5 -size +5c | grep -q temp_basal && echo "temp_basal.json more than 5m old. ")); then
@@ -368,7 +376,11 @@ function refresh_temp_and_enact {
 
 function refresh_pumphistory_and_enact {
     # set mtime of monitor/glucose.json to the time of its most recent glucose value
-    touch -d "$(date -R -d @$(jq .[0].date/1000 monitor/glucose.json))" monitor/glucose.json
+    if grep "MDT cgm" openaps.ini; then
+      touch -d "$(date -R -d @$(jq .[0].date/1000 nightscout/glucose.json))" monitor/glucose.json
+    else
+      touch -d "$(date -R -d @$(jq .[0].date/1000 monitor/glucose.json))" monitor/glucose.json
+    fi
     if ((find monitor/ -newer monitor/pumphistory-zoned.json | grep -q glucose.json && echo -n "glucose.json newer than pumphistory. ") \
         || (find enact/ -newer monitor/pumphistory-zoned.json | grep -q enacted.json && echo -n "enacted.json newer than pumphistory. ") \
         || (! find monitor/ -mmin -5 | grep -q pumphistory-zoned && echo -n "pumphistory more than 5m old. ") ); then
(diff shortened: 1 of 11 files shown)


lib/oref0-setup/mdt-cgm.json
```diff
@@ -125,15 +125,15 @@
...
+      "command": "! bash -c \" (echo -n MDT cgm data retrieve && openaps monitor-cgm 2>/dev/null >/dev/null && grep -q glucose cgm/cgm-glucose.json && echo d) && cp -pu cgm/cgm-glucose.json cgm/glucose.json && cp -pu cgm/glucose.json monitor/glucose-unzoned.json && (echo -n MDT cgm data reformat && openaps report invoke monitor/glucose.json nightscout/glucose.json 2>/dev/null >/dev/null && echo ted)\""
      },
     "type": "alias",
     "name": "get-bg"
   },
   {
     "type": "alias",
     "upload-bg": {
-      "command": "report invoke nightscout/glucose.json nightscout/recent-missing-entries.json nightscout/uploaded-entries.json"
+      "command": "report invoke nightscout/recent-missing-entries.json nightscout/uploaded-entries.json"
```
scottleibrand: What does removing this do? Or, conversely, what did it do with it in?

lib/oref0-setup/mdt-cgm.json
```diff
@@ -125,15 +125,15 @@
   },
   {
      "get-bg": {
-      "command": "! bash -c \" (echo -n MDT cgm data retrieve && openaps monitor-cgm 2>/dev/null >/dev/null && grep -q glucose cgm/cgm-glucose.json && echo d) && cp -pu cgm/cgm-glucose.json cgm/glucose.json && cp -pu cgm/glucose.json monitor/glucose-unzoned.json && (echo -n MDT cgm data reformat && openaps report invoke monitor/glucose.json 2>/dev/null >/dev/null && echo ted)\""
+      "command": "! bash -c \" (echo -n MDT cgm data retrieve && openaps monitor-cgm 2>/dev/null >/dev/null && grep -q glucose cgm/cgm-glucose.json && echo d) && cp -pu cgm/cgm-glucose.json cgm/glucose.json && cp -pu cgm/glucose.json monitor/glucose-unzoned.json && (echo -n MDT cgm data reformat && openaps report invoke monitor/glucose.json nightscout/glucose.json 2>/dev/null >/dev/null && echo ted)\""
```
scottleibrand: What does moving this here do exactly?

bin/oref0-pump-loop.sh
```diff
@@ -339,7 +339,11 @@ function refresh_old_profile {
 
 function refresh_smb_temp_and_enact {
     # set mtime of monitor/glucose.json to the time of its most recent glucose value
-    touch -d "$(date -R -d @$(jq .[0].date/1000 monitor/glucose.json))" monitor/glucose.json
+    if grep "MDT cgm" openaps.ini; then
+      touch -d "$(date -R -d @$(jq .[0].date/1000 nightscout/glucose.json))" monitor/glucose.json
```
scottleibrand: I wonder if it'd be worthwhile making the monitor vs. nightscout directory a variable so we only have to do the if clause once.

lib/oref0-setup/mdt-cgm.json
Bender1061: Since we are using the nightscout/glucose.json instead of the monitor/glucose.json, this keeps those two files with the same CGM data. Otherwise nightscout/glucose.json is updaed by the ns-loop. So this keeps the two timestamps about the same with the same data.

lib/oref0-setup/mdt-cgm.json
Bender1061: above comment

bin/oref0-pump-loop.sh
Bender1061: that I can work on tonight, not that it makes a huge difference. since that type of memory really is not a problem. But I'll make a bool.

scottleibrand: I am not concerned about memory usage, but rather code duplication. So I'd be fine with it being a sting, as long as we can reduce the number of places we do the same if check (and that will need to be updated together later if we change that).

Bender1061:
This change will check for any type of cgm and place that in a new data cgmtype. So if in the future anyone needs to do anything special for any of the cgm types everything is already there for it.. 

I figured getting that data from the oref0-runagain.sh seams like a logical way of doing that.

lib/oref0-setup/mdt-cgm.json
lukas-ondriga: Why do you use nightscout/glucose.json? For me it is quite confusing and it makes the code complicated too - you need additional conditions for that. I would prefer to create monitor/glucose.json in the same way as the nightscout/glucose.json is created.

Bender1061: I agree all the conversion of the enlite data is confusing. but since the enlite is the only CGM that uploads to nightscout, we have to have the correct format. And it took some very smart people a long time to get that all to work. You are more than welcome to try to simplify it :-) 

to answer your question
the reason to use the nightscout/glucose.json because that is the only cgm data that have the date field that is being used by jq

scottleibrand: That is actually no longer true.  The g4-upload setting uploads Dexcom G4 data as well, but it is quite a bit different than the MDT setup because it's also uploading all the calibrations and other data to display "raw" sensor readings in NS.

👍 to simplifying and consolidating all the aliases and reports.  The best way to do that might be to re-do the aliases as a shell script, as we did with oref0-pump-loop...

Bender1061: I stand corrected, but I'm not tearing into building that Bash Script right now. 
But you're right that would make it easier for people. It's awfully nice while troubleshooting pump loop to find where it's messing up now!

ericfeibelman: Testing this one all day ... dev, microbolus (SMB) enabled), 722, MDT enlites. It seems to be working very well, without the jq and date errors in the log and with very consistent BG updates to NS and faster looping. Thanks!

Bender1061:
I was thinking about this the other day, we need to get someone that is not using MDT to test this and make sure it still works for them

Kdisimone:
Worked fine for me using G4-upload just now.

scottleibrand:
Just opened https://github.com/Bender1061/oref0/pull/2 to clean up the code a bit.  It shouldn't change functionality at all, so I'm fine merging this to dev once that is merged.  Or, if @Bender1061 isn't available to merge my PR to his branch, I can just merge straight from mdt-div-zero to dev.

Bender1061:
Looks good, I won't be able to test till tonight though

Bender1061:
Works good 👍 Lets merge it!

PR #481, opened 2017-05-25
Setup script

Kdisimone:
adding pushover prompt questions
Saving pushover flags in runagain script
hidden microbolus enable prompt (for future use)
adding bt question


Kdisimone:
I'd love to add the BT choice as a question (rather than an added flag) so that the docs only have one setup script line in them...I think that would be easier for new users especially.

danamlewis:
LGTM (I added what I wanted to do with de-advancing features, too.) @scottleibrand do you want to merge this as-is to dev, or also add in the maxIOB changes here?

danamlewis:
Targeting this to max-iob branch so I can test it in concert w/ changes there

PR #484, opened 2017-05-29
restore cp of newprofile

tynbendad:
restore "Copy tuned profile produced by autotune to profile.json for use with next day of data"

PR #485, opened 2017-05-29
restore "Copy tuned profile produced by autotune to profile.json for …

hilarykoch:
Copy tuned profile produced by autotune to profile.json for use with next day of data"

Per suggestion for dev branch from tynbendad.


danamlewis:
Confirmed via https://github.com/openaps/oref0/pull/484 that this works, so merging this to this branch, too.

PR #486, opened 2017-05-31
Re-working oref0-setup.sh

danamlewis:
Includes:

- Formatting and colors for readability. (Ya'll are welcome)

-  AMA, autosens, and autotune are now out of advanced features

- Advanced features sub-menu now holds Pushover setup prompts, and a reminder to manually enable oref1 features if you want them.

- maxIOB will now be set here, with a note about recommending 0 or 1 hour of basal to newbies; it defaults to zero if not set.


danamlewis:
Testing checklist:

- Make sure maxIOB gets properly saved to preferences and is properly used in the loop
- Check via cat and nano that bash run-again is formatted correctly
- Confirm that both via copy-paste of run-again contents and bash run-again that those both successfully re-run the script with the input variables as expected.

danamlewis:
@tomasboudr per your comment in the other thread - did you accidentally hit enter instead of providing a value for the maxIOB Q? If you think you hit a number and it didn't save, can you screenshot me privately what your output from the script looked like?

tomasboudr:
I tested this and my maxIOB number from the script was corrected passed to preferences.json

danamlewis:
@tomasboudr awesome! can you check on the cat runagain and bash run again to confirm those work? I goofed up formatting on those earlier, so I want to make extra-sure to validate those are back to normal now.

Kdisimone:
Confirmed paper trail, setup script and BT tethering all working great

PR #494, opened 2017-06-09
Smarter carbsReq

scottleibrand:
The current pushover alerts tend to false-alarm for carbsReq when BG is still high and there's plenty of time to prevent a low with SMB zero temps.  Rather than simply including 30m worth of zero temps in calculating carbsReq, this calculates how many minutes BG is expected to stay above min_bg (the low end of your target range) and includes that many minutes worth of zero temp in the carbsReq calculations.  So if you aren't projected to go low for 90 minutes, it'll include 90m worth of zero temps, and likely won't alarm prematurely for carbsReq.  If you're already below target, or projected to drop below target in just a few minutes, it'll be much more apt to pushover for carbsReq.

### DIFF
--- a/bin/oref0-determine-basal.js
+++ b/bin/oref0-determine-basal.js
@@ -108,7 +108,7 @@ if (!module.parent) {
     // console.error("Printing this so you know it's getting to the check for autotune.")
 
     //printing microbolus before attempting check
-    console.error("Microbolus var is currently set to: ",params['microbolus']);
+    //console.error("Microbolus var is currently set to: ",params['microbolus']);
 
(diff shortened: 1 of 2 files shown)


PieterGit:
Is this already ready to test? Looks nice? Why don't you use the UAMpredBG if it's available?

scottleibrand:
Yes, should be fine for testing. UAMpredBG often predicts BG headed "up and to the right" and then a few minutes later starts predicting a low if BG flattens or drops. So I figured it would be safer to only use COB and IOB predBGs for snoozing carbsReq alarms.

Issue #175, opened 2016-07-22
Auto resume when suspended for too long


jasoncalabrese:
A common way to more accurately track IOB when disconnected from the pump while swimming, showering, etc is to temp to 0 and then suspend to "lock in the temp"

One of the risks with doing this is you may forget to resume.  To lower that risk we could auto-resume the pump after 30mins or until any 0 temp expires.

Related to this is correctly calculating IOB when pump was suspended without needing to also set a 0 temp.

scottleibrand:
Per https://gitter.im/nightscout/intend-to-bolus?at=57ed497a34a8d5681ccae705, one way we might be able to auto-resume more safely would be to wait until all running temps have expired, it's been 30m+ (configurable), *and* a high-temp is required.  This would avoid the (slight) possibility that we resume basal and then are unable to set a low-temp if needed, and would more generally avoid overriding users' decisions to temp to zero and suspend to shut off delivery for whatever reason.

scottleibrand:
Does anyone feel that this is a feature we really need?  If so, any volunteers to help code (or at least test) it?

scottleibrand:
Closing for lack of interest.  Please reopen if that changes.

esteward:
i have encountered this issue once before and forgot to resume my pump. I would be willing to test it. you do not want me going anywhere near the code.

scottleibrand:
This (https://github.com/openaps/oref0/issues/496) has now been merged to dev: you can `git pull` the latest oref0 dev and add `"unsuspend_if_no_temp": true` to your preferences.json to enable it.  Once enabled, you should keep / make a habit of temping to zero and suspending your pump before your shower, and unsuspending when you reconnect the pump. If you forget to unsuspend it will do so automatically after the zero temp expires.

We should also add these instructions to the OpenAPS docs before we close this issue.

danamlewis:
https://github.com/openaps/docs/pull/907 <-- issue can be closed, this docs PR is created and waiting for the next dev to master release.

PR #496, opened 2017-06-19
unsuspend pump automatically if temp basal duration is zero

scottleibrand:
Per https://github.com/openaps/oref0/issues/175, sometimes people forget to resume / unsuspend their pump after reconnecting it.  This would cause oref1 to automatically resume the pump if the underlying zero temp has expired.  As long as the zero temp is still running, it would honor the suspend status and do nothing.  As written, this would not apply to users running the old `openaps pump-loop`, only people who have enabled microboluses in oref0-setup and are using oref0-pump-loop.sh.

Thoughts?

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -217,8 +217,12 @@ function smb_bolus {
 function unsuspend_if_no_temp {
     # If temp basal duration is zero, unsuspend pump
     if (cat monitor/temp_basal.json | json -c "this.duration == 0" | grep -q duration); then
-        echo Temp basal has ended: unsuspending pump
-        openaps use pump resume_pump
+        if (grep -iq '"unsuspend_if_no_temp": true' preferences.json); then
+            echo Temp basal has ended: unsuspending pump
+            openaps use pump resume_pump
+        else
+            echo unsuspend_if_no_temp not enabled in preferences.json: leaving pump suspended
+        fi
     else
         # If temp basal duration is > zero, do nothing
         echo Temp basal still running: leaving pump suspended
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -227,7 +227,7 @@ function unsuspend_if_no_temp {
     else
         # If temp basal duration is > zero, do nothing
         echo Temp basal still running: leaving pump suspended
-    fi)
+    fi
 }
 
 # calculate random sleep intervals, and get TTY port
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -66,6 +66,7 @@ smb_main() {
     ); do
         if grep -q '"suspended": true' monitor/status.json; then
             echo -n "Pump suspended; "
+            unsuspend_if_no_temp
             smb_verify_status
         else
             echo Error, retrying && maybe_mmtune
@@ -212,6 +213,16 @@ function smb_bolus {
     else
         echo "No bolus needed (yet)"
     fi
+}
+function unsuspend_if_no_temp {
+    # If temp basal duration is zero, unsuspend pump
+    if (cat monitor/temp_basal.json | json -c "this.duration == 0" | grep -q duration); then
+        echo Temp basal has ended: unsuspending pump
+        openaps use pump resume_pump
+    else
+        # If temp basal duration is > zero, do nothing
+        echo Temp basal still running: leaving pump suspended
+    fi)
 }
 
 # calculate random sleep intervals, and get TTY port
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -64,6 +64,7 @@ smb_main() {
             && touch monitor/pump_loop_completed -r monitor/pump_loop_enacted \
             && echo \
     ); do
+        smb_verify_status
         if grep -q '"suspended": true' monitor/status.json; then
             echo -n "Pump suspended; "
             unsuspend_if_no_temp
--- a/lib/profile/index.js
+++ b/lib/profile/index.js
@@ -21,6 +21,7 @@ function defaults ( ) {
     , maxCOB: 120
     , override_high_target_with_low: false // allows a much higher target used only for avoiding bolus wizard overcorrections
     , skip_neutral_temps: false // if true, don't set neutral temps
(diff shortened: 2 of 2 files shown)


mgranberryto:
I have mixed feelings about this, but I also have it set up to SMS me if I leave the pump suspended.  I think the better fix is a better mechanism to notify OpenAPS that the pump is disconnected.  I can't think of a situation where this would have done me harm and several where it would have caught a mistake earlier, but it still feels like it's a surprising behavior.  I know a semi-pro tennis player that chooses to go pump-free while training, only bolusing for a couple snacks throughout the day because his insulin needs are so low.  He leaves his pump suspended during those periods because he simply doesn't need the insulin during his training days.

I'll concede that there are fewer people like him than people like me who sometimes forget to unsuspend the pump after a shower or other reasons to disconnect, for me, this would be a welcomed change, but it may surprise people and insulin delivery will be affected.  I can imagine a situation where I didn't have a reservoir on me and had to inject insulin away from a pump and would want to suspend delivery until the other insulin had cleared my system.  This is one of those situations where I wish we had an alerting device as part of the base system.

scottleibrand:
Wouldn't someone in that scenario need to set a long temp basal to zero (for the duration the pump is expected to be suspended) so that IOB is correct when the pump is later unsuspended?

scottleibrand:
@danamlewis Any thoughts?

danamlewis:
@scottleibrand 👍 would this be in preferences to toggle?

scottleibrand:
Only if someone insists that we need Yet Another Preference for it.  I would prefer to make it the default behavior.

danamlewis:
@kdisimone thoughts on Yet Another Preference(TM) vs having it in the code so if people do want to disable, they can just comment out the code? (I'm ok with lengthening the preferences list)

PieterGit:
@scottleibrand : I think this behaviour warrants a new preference. I think the mental model of people is that oref0 changes the temp basals, oref1 does microboluses. Having pumps unsuspend is not expected behaviour. So I would make it default False and if the user chooses to change the preference it's aware of this pump behaviour.

betluis:
Many times my daughter has ended up high due to forgetting to re-activate the pump, specially at night. My personal opinion is that the system should reactivate the pump when 0 temping is not needed any more. The pump therapy is required to keep a person within range, if somebody purposely wants to stop insulin delivery regardless of sugar levels then probably the best way is to disconnect the pump from the body altogether. In practice every time my daughter suspends the pump she intends to do it temporarily but then she forgets to reactivate it...

mgranberryto:
This would be helpful for me, but I know to expect it.  It seems like suspending is an explicit "do not give insulin" command and overriding that seems strange.  I'm with Pieter on this one.  It's a feature that I want enabled, but not by default.  What if emergency care workers suspend the pump and take over insulin delivery without disconnecting?  The corner cases are problematic here.

tim2000s:
My view on this is that suspension is a willing interference by a human so automatic resume should not be a default setting. There can be good reasons for it. It's also not a feature that I think should be enabled. For example, the shower setting is to remove the pump, set basal to zero then suspend. With something like this enabled, it would cancel the zero temp. 

I think we probably need to hash out what the use cases and counter cases are before setting up a test version. 

It raises the question about whether we should build in a feature that purports to replace a forgetful adult but also overrides a conscious one.

scottleibrand:
@tim2000s no, it wouldn't cancel the zero temp in a zero-temp then suspend case.  If there is a zero temp still running, it will never unsuspend.  Only once the zero temp expires would it do so.  So if a conscious adult wants to suspend for 6 hours, then they set a 6 hour temp to zero and suspend.  If they only plan to be suspended for 30m, then they only set a 30m zero temp before suspending.

Can you explain what use cases (good reasons) there would be for suspending without an appropriate-length zero temp, or for wanting the pump to remain suspended indefinitely (with the rig still running)?

moomoobloo:
How about starting with on on-by-default Pushover alert? That should be
fairly uncontroversial.

danamlewis:
Not everyone has or wants Pushover. That's certainly one option, but I'd rather create the preference, default it to false, and allow those who want it to turn it to be true. Any information about edge cases/cautions about using it can be added to the docs for people to consider, like any other feature, before turning it on.

scottleibrand:
A pushover alert would also be a good idea, but we can't rely on those, because they require the user to purchase and configure an extra service.  It sounds like there are enough people who might want to turn off auto-unsuspend that it should be a preference, and enough people are uncomfortable with on-by-default that I think I'd be ok with off-by-default.  I suspect this will turn out like AMA, though, and everyone will go turn it on as they set up their rigs.

danamlewis:
Added to preferences with an explanation; defaults to false. LGTM for merging to dev @scottleibrand

danamlewis:
Testing, it worked:

```
Temp refreshed: monitor/temp_basal.json: {"duration":0,"rate":0,"temp":"absolute"}
Checking pump status (suspended/bolusing): {"status":"normal","bolusing":false,"suspended":true}
Pump suspended; Temp basal has ended: unsuspending pump
{
  "status": "resumed",
  "recieved": true,
  "enacted_at": "2017-06-28T15:27:00.974838"
}
```

scottleibrand:
With pump suspended and zero temp still running:
```
Temp refreshed: monitor/temp_basal.json: {"duration":8,"rate":0,"temp":"absolute"}
Checking pump status (suspended/bolusing): {"status":"normal","bolusing":false,"suspended":true}
Pump suspended; Temp basal still running: leaving pump suspended
Checking pump status (suspended/bolusing): {"status":"normal","bolusing":false,"suspended":true}
```

scottleibrand:
As soon as the temp basal ended, it unsuspended:
```
Temp refreshed: monitor/temp_basal.json: {"duration":0,"rate":0,"temp":"absolute"}
Checking pump status (suspended/bolusing): {"status":"normal","bolusing":false,"suspended":true}
Pump suspended; Temp basal has ended: unsuspending pump
{
  "status": "resumed",
  "recieved": true,
  "enacted_at": "2017-06-28T16:12:54.057081"
}Checking pump status (suspended/bolusing): {"status":"normal","bolusing":false,"suspended":false}
```
and then set a temp on the next pump-loop.

scottleibrand:
Oh, and I forgot to copy and paste it, but we also saw that oref0-pump-loop would not unsuspend without `"unsuspend_if_no_temp": true` set in preferences.

I think that's all the testing we need before merging to dev.  If no objections, I'll do so shortly.

Issue #500, opened 2017-06-23
"ValueError: Invalid GPIO pin specified" error after script install


jbwittmer:
Multiple users reporting error "ValueError: Invalid GPIO pin specified" in pump-loop log after new flash and install of openaps using both Jubilinux 0.1.0 and 0.1.1 on both master and dev branches.

Stavrosnco:
In case it helps, it seems possible this issue has arisen in the past week as I flashed an edison and had it running fine on the dev branch on 6/15 and then flashed on today and got this error.

jbwittmer:
I built two rigs on either 6/17 or 6/18 -  and they both worked - so something changed even more recently than 6/15.

Stavrosnco:
A temporary and ugly work around can be achieved by the following:
Navigate to a directory you can make a new directory in without getting it mixed up with your openAPS directory. I just put mine in /home/test
`
git clone https://github.com/intel-iot-devkit/mraa.git
cd mraa/
git reset --hard e0052acfa7ab441b5417a0fbd010ce77b285a4d8
mkdir build
cd build
cmake -DBUILDSWIGNODE=OFF ..
make
sudo make install
reboot
`

jbwittmer:
This temporary fix seems to have worked for me as well

jbwittmer:
https://github.com/openaps/oref0/pull/501

Above is my attempt at changing oref0-setup.sh to pull the  e0052acfa7ab441b5417a0fbd010ce77b285a4d8 branch instead of master branch.  Haven't been able to test yet.

PieterGit:
@KDisimone @jbwittmer : you probably can best use the latest mraa release, https://github.com/intel-iot-devkit/mraa/releases/tag/v1.7.0 instead of a specific commit.

Did one of you create an issue at https://github.com/intel-iot-devkit/mraa/ because one of their latests commits broke openaps?

jburke41:
This may be a very dumb questions--I am unsure how to make a new directory in order to fix this error

PieterGit:
@scottleibrand asked me to create a fix. I added another pull request. I tested the git commands, but don't have a spare Explorer rig available.

PieterGit:
I created an issue upstream: https://github.com/intel-iot-devkit/mraa/issues/771

[...]

PR #504, opened 2017-06-23
Hotfix for mraa

PieterGit:
For release 0.4.2 a github checkout of the master branch was used. Therefore a change in mraa could break openaps. For the master branch we
should always use a mraa release instead of a master branch checkout.

This fixes https://github.com/openaps/oref0/issues/500 and makes it possible again to install openaps.

Rigs that have already mraa installed were not affected with this issue and there is no immediate need to upgrade from 0.4.2 to 0.4.3.


Kdisimone:
I'll give the fix a test later tonight (about 8 hours from now I should have some tests done)

PieterGit:
@Kdisimone : great! i found a small output error in the `die` statement which I just fixed.

scottleibrand:
Per https://github.com/openaps/oref0/issues/500#issuecomment-310792861 this has been successfully tested, so I'll go ahead and merge.

PR #511, opened 2017-07-02
raise default min_5m_carbimpact from 5 to 8

scottleibrand:
This should help reduce the impact of carbs that don't show up in BG deviations, and minimize the impact of any zombie carbs that reappear.


scottleibrand:
Closing in favor of #515

PR #513, opened 2017-07-02
merge_pumphistory to fix zombie carbs

scottleibrand:
and use pumphistory-merged.json for COB calculations in meal.json

This should help ensure that the insulin dosing history used to calculate historical carb absorption is always complete, which hopefully will eliminate zombie reappearing carbs.


scottleibrand:
Also updated the COB calculations to consider carbs from up to 6 hours ago, not just from the last DIA hours.  This should help AMA or SMB more accurately dose for the tail of large slowly-absorbing meals.

scottleibrand:
Created a zombie-carbs branch off of dev *without* these fixes, so if anyone needs to compare the before and after they can use that.

PR #514, opened 2017-07-02
Add MDT to pump-loop.sh

Bender1061:
add the get CGM data from MDT into pump-loop.sh

removes the Alias calls for getting the cgm data from the pump

fixed the grep for MDT so that it does not show the Printout of the Alias every time the script checks for MDT 

I made it so that when it does get the data from the pump, if it fails due to a decocare error
"decocare.errors.DataTransferCorruptionError: Page size too short"
I did never realize this before while it was in the Alias that this happens a lot, normally when it does fail if you just try to run it again it fixes and works fine, So I've made it so it can fail 3 times. I've never seen it have to run more than twice. 

Also have the loop check the time stamp on monitor/cgm-mm-glucosedirty.json after getting data if it is not any newer than the data already received, there is no reason to do the whole reformatting of the data again.
I've also made a function to wait for new CGM Data, but I don't think this is really necessary as it takes a while for the pump-loop to complete anyways, so very seldom when it runs do you not have new cgm data.  But I really would like some feedback if I should just get rid of it or not. as of right now, I do not have it being called. 

Was also debating taking all of this and making it's own bash script, but I don't think that is the best idea since it really does need to be run in the certain order in the Pump-loop. but I'm open to suggestions.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -250,25 +250,26 @@ function if_mdt_get_bg {
     echo -n
     if grep "MDT cgm" openaps.ini 2>&1 >/dev/null; then
         echo \
-		&& echo -n MDT cgm data retrieve
-		#due to sometimes the pump is not in a state to give this command repeat till it completes
+		&& echo -n Attempting to retrieve MDT CGM data 
+		#due to sometimes the pump is not in a state to give this command repeat until it completes
 		#"decocare.errors.DataTransferCorruptionError: Page size too short"
 		n=0
 		until [ $n -ge 3 ]; do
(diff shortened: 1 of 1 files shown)


scottleibrand:
Thanks.  I agree that waiting for new CGM data isn't really useful with a MDT CGM, as you might as well do the full loop if you have to talk to the pump anyway.  I added a TODO to remove that if no one decides it's useful by the time we do our next oref0 release.

Have you tested this thoroughly enough to be ready to merge it to dev?

Bender1061:
I've been running it for a couple weeks, I just updated my branch to the latest changes changes today. So I don't think there is any conflicts, but I'll run it over night and let you know Monday.

Bender1061:
I'm good, Lets Merge it 👍

PR #515, opened 2017-07-03
Update default preferences

scottleibrand:
In oref1 testing so far, nearly all complaints have been about it not dosing enough insulin after meals, and almost everyone has ended up raising remainingCarbsCap and remainingCarbsFraction.  This would change the defaults there to be the same as the max limits: 100% of 90g.

Also incorporates #511

### DIFF
--- a/bin/oref0-get-profile.js
+++ b/bin/oref0-get-profile.js
@@ -22,7 +22,7 @@ function usage ( ) {
 }
 
 function exportDefaults () {
-	var defaults = generate.defaults();
+	var defaults = generate.displayedDefaults();
 	console.log(JSON.stringify(defaults, null, '\t'));
 }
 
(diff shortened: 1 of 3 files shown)


danamlewis:
This LGTM for now, would like to continue investigating if we can not-auto-add-to-preferences.json the ones no one is likely to change (like these now, and maybe some others)

scottleibrand:
Ok, now only the typically-adjusted settings show up in preferences.json by default.  If anyone wants to add something else, they can do so manually, or use --updatePreferences to get the full list of all preferences added to their preferences.json.

Also enabling rewind_resets_autosens and adv_target_adjustments by default.

danamlewis:
👍

danamlewis:
Docs for the new world preferences will look like http://danamlewis-openaps-new-docs.readthedocs.io/en/latest/docs/While%20You%20Wait%20For%20Gear/preferences-and-safety-settings.html#other-preferences (test page of docs for now but that will become dev/master soon)

PR #493, opened 2017-06-08
Setup script: Clarifying SMB manual enable, NS URL example

Kdisimone:
Emphasizing that the preferences set to true will not enable SMBs/UAM.

Adding more response to the commented-out part

Changing example NS site to heroku since that's what most will use now.

PR #516, opened 2017-07-03
Safety gate oref1 stuff in oref0-setup

scottleibrand:
Fixed syntax and logic from https://github.com/openaps/oref0/pull/493


scottleibrand:
A quick run-through of the interactive oref0-setup looks good.  Should probably test on a real rig that it actually enables SMB successfully.

Kdisimone:
Worked well for me.  Appropriately kicked out (or let in) depending on my actions.  Running loop fine with it.

PR #517, opened 2017-07-03
treat max_iob as string to support fp

tynbendad:
untested


scottleibrand:
Does this result in `"max_iob": 4` or `"max_iob": "4"` in preferences.json?

tynbendad:
I think it should be the same but I will test later today.

tynbendad:
i just tested on 2nd rig - it worked fine, no errors, and preferences.json looks good:
        "max_iob": 2.8,

PR #518, opened 2017-07-03
find last BG from before a meal, and interpolate any gaps

scottleibrand:
This should fix the issue where a CGM restart or other data gap overlaps a meal entry, thereby causing carb absorption for that time period to be missed/ignored.

I've simulated that scenario by deleting CGM values out of glucose.json, but we'll also need to test it with real receiver data where the receiver was off, calibrating, or out of range around the meal time.

### DIFF
--- a/bin/oref0-get-profile.js
+++ b/bin/oref0-get-profile.js
@@ -22,7 +22,7 @@ function usage ( ) {
 }
 
 function exportDefaults () {
-	var defaults = generate.defaults();
+	var defaults = generate.displayedDefaults();
 	console.log(JSON.stringify(defaults, null, '\t'));
 }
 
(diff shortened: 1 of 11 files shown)


scottleibrand:
TODO: once testing is completed, remove extra debug statements.

Kdisimone:
Worked GREAT!  Ate 23 g carbs right before sensor session stopped. Started up new session about three hours later and cob was zero. Perfect!  So nice!!

Kdisimone:
img_4471

img_4473

scottleibrand:
We'll also want to test the scenario where the carb entry occurs after the sensor data stops, as that is the part that I changed in the code.

scottleibrand:
We haven't seen any regressions with this, and it seems to help with COB calculations, so I'm going to merge this to dev for further testing there.  If anyone testing this still wants to compare this to the old behavior, you can check out the `zombie-carbs` branch.

PR #519, opened 2017-07-03
log autosens ratio in pump-loop.log; better rounding

scottleibrand:


scottleibrand:
Minor cosmetic change; ready to merge if no objections.

PR #521, opened 2017-07-04
switch back to latest stable mraa release for oref0 release to master

scottleibrand:


scottleibrand:
@PieterGit I'm thinking we release oref0 0.5.0 with the stable mraa 1.7.0 we've been using for ages, and then check after each oref0 release to see if they've released a new stable mraa, and increment the version number at that point. Does this approach make sense to you?

PieterGit:
Makes perfectly sense for me. There was one RPI issue that was fixed for MRAA in the last months, but I don't think it affects us. See https://github.com/intel-iot-devkit/mraa/issues/771

After 0.5.0 is out and stable for about two weeks I think it's good to switch back to their master branch and keep the early integration on oref0 dev with MRAA to avoid surprise for stable users.

PR #512, opened 2017-07-02
per #419, ignore bolus wizard carb entries without a bolus

scottleibrand:
This needs to be tested by someone with a pump that is capable of writing carb-only bolus wizard records when you cancel out of the bolus wizard or enter a zero bolus.

Test procedure:

 - Enter a small number of carbs via the bolus wizard.  Cancel out or enter a zero bolus.  Make sure that the carbs show up in NS and in meal.json.
 - Check out this branch: `cd ~/src/oref0 && git fetch && git checkout ignore-carb-corrections && git pull`
 - Enter another small number of carbs the exact same way as before.  Make sure the carbs do *not* show up in NS or in meal.json.
 - Enter a small number of carbs with a minimal bolus (0.1U or whatever your pump supports).  Make sure those show up normally.
 - Test out other normal scenarios and make sure they all still work.

PR #526, opened 2017-07-06
upload carb-only bolus wizard entries to NS again

### LINKED ISSUE #524: Carbs and boluses not showing in NS treatments.
Installed the interpolate-lastBG-for-cob branch yesterday and while the initial purpose was fulfilled, it appears to cause intermittent problems with uploading treatment data from the pump, both from the bolus wizard as well as easy button boluses. 

Some meals have been recorded well, but most have not. A couple bolus wizard entries (carbs and bolus) were uploaded as notes only in the treatments. 

Most easy bolus button boluses have not shown in the treatments. 

Openaps does appear to be reading the values however, as iob and cob appear to be tracking correctly. 

Perhaps revert that branch until it can be further tested?

scottleibrand:
To fix #524, this reverts the portion of #512 that disabled the upload of carb-only bolus wizard entries, which looks to be independent of the changes to oref0-meal that ignore such entries for purposes of calculating COB.

We'll need to test that a carb-only bolus wizard entry when there's no COB does not end up contributing to COB via getting uploaded to NS and then downloaded again into carbhistory.json.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -247,7 +247,7 @@ function if_mdt_get_bg {
     echo -n
     if grep "MDT cgm" openaps.ini 2>&1 >/dev/null; then
         echo \
-	&& echo Attempting to retrieve MDT CGM data from pump 
+        && echo Attempting to retrieve MDT CGM data from pump
 		#due to sometimes the pump is not in a state to give this command repeat until it completes
 		#"decocare.errors.DataTransferCorruptionError: Page size too short"
 		n=0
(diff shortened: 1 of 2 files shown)


scottleibrand:
@Kdisimone feel free to test this if you like.  Otherwise we'll probably test it this weekend or next week.

scottleibrand:
Actually, looks like I won't be able to test this with our normal pump, as it doesn't log carb-only records.  @kdisimone would you prefer to test this here before I merge it to dev, or just test it in dev afterward?

scottleibrand:
Gonna merge this to dev to simplify testing it there.

Issue #527, opened 2017-07-06
Unable to initialize libusb: -99 errors with Dexcom


PieterGit:
Dexcom G4 connected to USB OTG to a new Explorerboard and powered with a Zendure powerbank.
When I charge the Zendure the power on the USB UART is very shortly unavailable.
It doesn't reboot because I have a LIPO attached to the Explorerboard.
But it will upset the USB-subsystem, causing the Dexcom to stop being able to get CGM data and to stop the pumploop after 10 minutes.

```
root@edison# lsusb
unable to initialize libusb: -99
```
Workaround I have is to add a crontab entry
```
*/15 * * * * /usr/bin/lsusb || shutdown -r now "lsusb error"
```

I think it might be related to the combination of Dexcom G4  and Linux Kernel.
```
root@edison# openaps use cgm GetFirmwareHeader
{
  "DexBootVersion": "6",
  "RFVersion": "1.0.0.28",
  "ApiVersion": "2.3.0.0",
  "ProductName": "Dexcom G4 Receiver",
  "SoftwareNumber": "SW10324",
  "TestApiVersion": "2.4.0.0",
  "SchemaVersion": "1",
  "PortVersion": "4.6.4.60",
  "FirmwareVersion": "2.0.1.216",
  "ProductId": "G4Receiver"
}
root@edison# uname -a
Linux lou78 3.10.98-poky-edison #5 SMP PREEMPT Tue Apr 4 22:42:44 EDT 2017 i686 GNU/Linux
```

Questions:
1. Does anybody else also experience this behaviour?
2. What kernel (`uname -r`) and Dexcom version do they use (`openaps use cgm GetFirmwareHeader`)?
3. Should create a PR for this? I see three types of implementation in oref0-setup.sh
- Default for `CGM=G4-upload` 
- Optional: As a question/command line item
- Integrate it with the cgm-loop after 10 minutes of no CGM

PieterGit:
@jbwittmer: since you have the same type of receiver? Can you tell me if you also experience this problem?

scottleibrand:
Another workaround is to get a power bank that supports pass through charging, like the Tecknet one. 

Would still be worthwhile trying to reboot if both NS and USB are unavailable. Wouldn't want to always reboot on lsusb -99, though, as that would mean constant reboots when the Dexcom receiver is unplugged.

PieterGit:
[...]

PR #528, opened 2017-07-07
Reboot rig if Unable to initialize libusb: -99 errors with Dexcom connected to USB OTG and uptime > 30m

PieterGit:
fixes https://github.com/openaps/oref0/issues/527


PieterGit:
You mean like this? I will test this on a new rig.

scottleibrand:
Yeah, I think that'll do it, thx.

PieterGit:
It works, but you need to repower the USB before the rig is rebooted. 
Otherwise it won't reboot anymore (because it has been rebooted once).
There is no trigger to know when the power on the UART is back up.
So I think I'm stuck with two options:
1.  create an option for it, and let the user decide if reboot after 16 minutes, and then each half an hour is ok
2.  find some way to find out at which moment the USB UART gets 5V power. Do you know any?
3. upgrade to kernel 4.12 in which most USB-kernel issues for OTG are probably gone. 

If nobody knows a way to do 2, I think short-term solution is 1. (which is fine for me, because we always use USB OTG and primarily one rig).

scottleibrand:
I don't understand the problem.  If you have incoming CGM data from either OTG or NS (your CGM data is more recent than the last reboot), then when CGM data gets to be 16 minutes old, it will reboot again.  What is the scenario where that's not good enough?

PieterGit:
- only use CGM from USB OTG
- battery on USB UART is empty
- will cause USB subsystem to cause lsusb -99
- rig will reboot once
- no effect, because battery is still empty
- user changes battery or powers batter
- rig won't reboot because it has rebooted once and uptime < cmg last update

scottleibrand:
Ok.  How are you preventing it from getting CGM data from NS?  Is this rig the only uploader?  Or is it just never online?

PieterGit:
The rig is the only uploader, so it will download CGM data from NS if it has uploaded it itselves. It's online when it's on wifi. Remember European Dexcom don't have bluetooth / Share, so only USB OTG.

scottleibrand:
In that case maybe some form of "I can hit the NS API, but it doesn't have any new CGM data" check would justify more frequent reboots?  That way, if you have two uploader rigs, or another uploader method like Share, it wouldn't reboot all the time, but if you don't have that it would fall back to rebooting periodically until it gets CGM data again...

If it's not obvious what I'm trying to accomplish here: I'd like to minimize the number of things for people to (mis)configure and instead have it behave sensibly in as many conditions as possible.

PieterGit:
I'll try to implement something like that later this weekend. Can I also reboot frequently (say every 30 minutes) if I cannot reach the NS API (timeout)? (e.g. if I'm offline and have a loose cable or empty USB UART battery).

scottleibrand:
It feels like there's some risk of doing that by default, but maybe it's not very high if we're only talking about a clean reboot every 30m...

PieterGit:
I'm thinking on rewritinga and creating a preference `reboot_on_cgm_usb_error` (default:  false) and a `reboot_on_cgm_usb_error_threshold` (default: 16 minutes). 
And require both uptime > reboot_on_cgm_usb_error_thresholdand cgm last updated > reboot_on_cgm_usb_error_threshold

scottleibrand:
I'm good with that.  Probably don't need to even have those preferences show up by default in preferences.json: anyone who wants them can add them in manually.

tynbendad:
just ran into this - glad someone else is thinking about this.. i "solved" this by adding a script to check if the lsusb output is different & do a pushover to me if it changes, so at least i can know when it failed ;).

>Probably don't need to even have those preferences show up by default in preferences.json: anyone who wants them can add them in manually.

this worries me - how will we find out about the options? are there other options like this & is there a central list of them?

scottleibrand:
@tynbendad openaps.rtfd.io/en/latest/docs/While%20You%20Wait%20For%20Gear/preferences-and-safety-settings.html

scottleibrand:
Sorry I forgot about this one.  The code looks good to me.  I'm not sure why you're requiring and upgrade of dependencies, though.

PieterGit:
i will drop the upgrade of dependencies for this pr, and won't reboot if fresh cgm's are available from Nightscout

scottleibrand:
Sounds good.  LMK when you're ready for me to take another look.

tynbendad:
Does this only work when enabled, or have a way to not affect non-receiver rigs?  Fwiw I rarely (maybe never?) see an lsusb error with receiver on 2017 explorer board or 2016 with trace cut.

PieterGit:
I will make it configurable with an option (default disabled). I can reproduce lsusb errors quite easily with both the 2017 and 2016 trace cut Explorer board. Today I added some extra logging for fresh rigs. I'll work on this PR the next days to test and improve it.

PieterGit:
@tynbendad @scottleibrand @sulkaharo : I'm thinking of closing this PR and not merging it.

Reasons:
- we should try to fix the root cause, probably kernel related
- during testing and developing my rig often rebooted, I can't use a `shutdown +5`  because the pump loop will then cancel that shutdown because there are no radio errors.  There is no coordinating between the pump and cgm loop that both might want to reboot the rig. Creating such a mechanism is too complex and error-prone.
- the added code to the cgm-loop is quite ugly and potentially error prone, we shouldn't let a workaround make the cgm-loop code less readable

I'm now thinking of three alternatives:
1.  documenting the best known workaround for those who are running in this problem:  
`*/15 * * * * /usr/bin/lsusb || shutdown -r now "lsusb error"`
2.  automating 1 in oref0-setup.sh (with a question and oref0 parameter to enable it in cron)
3. waiting to a jubilinux version with a 4.x kernel that probably doesn't have this problem

I propose to do 2. 
Thoughts?

scottleibrand:
Yeah, I think we should revisit this when we refactor the cgm-loop stuff.  I am fine with just documenting the workaround in the meantime.

scottleibrand:
Any reason to leave this open?

PieterGit:
Closing this PR. Adding this will be to much work. The workaround `*/15 * * * * /usr/bin/lsusb || shutdown -r now "lsusb error"` is much more robust for those who use rigs with Dexcom attached by USB.

PR #529, opened 2017-07-07
Fix argument order in ns-meal-carbs

jmcrawford:
Fix for error "Argument order has changed: please update your oref0-meal device and meal.json report to place carbhistory.json after basalprofile.json"


scottleibrand:
Looks like you applied this to an old version of dev that didn't have the --compressed stuff (and had some other diffs in files you're not modifying).  Also, I think the version of this that actually gets used is in lib/oref0-setup/alias.json rather than lib/templates/refresh-loops.json.

scottleibrand:
Applied via #532

PR #530, opened 2017-07-09
only pull 2h of pumphistory now that we merge it with pumphistory-24h

scottleibrand:
This should allow the pumphistory refresh to complete more quickly and more reliably without missing any data (since pumphistory-24h can never be more than 2h old).


scottleibrand:
Tested it on two rigs and confirmed that our pumphistory-merged.json is complete.

PR #534, opened 2017-07-10
upgrade to mmeowlink v0.11.1 for clearer pump comms messaging

scottleibrand:
In order to reduce confusion around people thinking that "no pump comms detected" is an error (rather than a good thing), I'd like to do a mmeowlink release and then upgrade everyone to mmeowlink v0.11.1  when they next run oref0-setup.

https://github.com/oskarpearson/mmeowlink/pull/63 needs to be merged (and released for `pip`) before this PR should be merged.


applehat:
@scottleibrand it's in pip now. Just as a heads up.

https://pypi.python.org/pypi/mmeowlink/0.11.1

Issue #531, opened 2017-07-09
Autotune as standalone fails to calculate ISF and makes wrong basal recommendations


drnoname82:
I am trying to use autotune as standalone. I enter carbs and boluses only through Nightscout Careportal directly. However, autotune fails to calculate an ISF because it fails to properly read boluses from Nightscout.

Boluses are included in ns-treatments.json, but the format of the bolus records does not seem to match what **oref0-autotune-prep** expects. This leads to BGI=0 at all times and thus a) no ISF recommendations b) wrong (too low) basal recommendations in the presence of boluses.

This is what a bolus entry in **ns-treatments.json** created by openAPS looks like:
```{"_id":"5678","bolus":{"_type":"Bolus","timestamp":"2017-07-06T18:31:24.000Z","programmed":1.2,"duration":0,"amount":1.2}}```
This is what a bolus entry created through Nightscout Careportal looks like:
```{"_id":"9999","enteredBy":"","eventType":"Meal Bolus","reason":"","carbs":48,"insulin":4.4,"duration":0,"created_at":"2017-07-06T19:29:00.000Z"}```

The error can be reproduced by running
```oref0-autotune-prep ns-treatments-openapsvscareportal.json profile.json ns-entries.2017-07-06.json > autotune.json```
using the files I have attached. Only the bolus entry created by openAPS will have an impact on BGI.

It appears to me that the parsing of treatments imported from Nightscout needs fixing to work with bolus entries created solely through Nightscout Careportal. This is where my understanding ends. I fail to understand how/where oref0-autotune-prep "recognizes" bolus entries from the ns-treatments file.

[Autotune error.zip](https://github.com/openaps/oref0/files/1134122/Autotune.error.zip)

scottleibrand:
This is handled at https://github.com/openaps/oref0/blob/dev/lib/iob/history.js#L97-L103

You probably just need to add another `} else if () {` clause to match on `"eventType":"Meal Bolus"` and use `current.insulin` instead of `current.amount` in that case.

drnoname82:
Thank you for pointing me in the right direction. I just created a pull request. Could you or someone else please have a thorough look at it. The entries created by Nightscout do not have a "timestamp", only a "created_at" time/date.

scottleibrand:
#547 has been merged, so closing this issue.

PR #535, opened 2017-07-10
Add import for Carbs&Insulin entered through Nightscout Care Portal

drnoname82:
This is an attempt to fix Issue #531.


scottleibrand:
Fixed some syntax errors with this branch in https://github.com/openaps/oref0/pull/547, so I'll close this and we can merge it via that PR instead.  Thanks!

PR #540, opened 2017-07-11
various oref0-pump-loop fixes affecting COB calculation

scottleibrand:
This avoids a corner case where both pumphistory files are old and the pumphistory-merged.json gets created before pumphistory-24h.json is refreshed, and then the rig miscalculates COB for a few minutes until pumphistory is refreshed again.


scottleibrand:
Also forgot to update ns-meal-carbs to use pumphistory-merged.json, so just fixed that.

PR #547, opened 2017-07-15
Add import for Carbs&Insulin entered through Nightscout Care Portal

scottleibrand:
This fixes some syntax errors (and inconsistent indentation) in https://github.com/openaps/oref0/pull/535 (which was an attempt to fix Issue #531) to avoid breaking IOB or COB calculations for any non-Nightscout-entered bolus records, and therefore get the failing unit tests to pass again.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -106,12 +106,8 @@ function smb_check_everything {
         && smb_verify_status \
         || ( echo Retrying SMB checks
             wait_for_silence 10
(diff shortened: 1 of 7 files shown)


scottleibrand:
@drnoname82: could you test this with your Nightscout-entered data and verify that it works?  You'll need to `git checkout drnoname82-dev` to get this branch.

drnoname82:
I tested all types of entries thoroughly, works like a charm. Thank you for helping me with this issue!

PR #549, opened 2017-07-16
Stop using ~/myopenaps/.git/

scottleibrand:
Since day 1, the `openaps` toolkit has used `git` to provide full auditability of every file whose contents are controlled by `openaps`. This has definitely been useful, especially in the early years when we were troubleshooting lots of bugs in the openaps toolkit and with peoples' custom aliases and reports. Since we've standardized everyone's openaps setups with oref0-setup, though, we've found that the need for json-file auditability has greatly diminished, to the point where I think it's been almost a year since we relied on the openaps git history for troubleshooting an issue. Instead, we either run openaps commands manually to reproduce an issue, or look at the pump-loop.log or other log output to see what was happening if there was an issue that is no longer occurring. This approach is serving us well enough that I don't think git is really needed any longer for most OpenAPS users.

In addition, we routinely (multiple times a week) have someone complaining on Gitter or (usually) Facebook that their rig has stopped working due to some sort of git error. We've automated fixes for most of these (up to and including automatically deleting the ~/myopenaps/.git repo), but would prefer to eliminate the problem entirely for average users by having oref0-setup configure their myopenaps directory with --nogit. Anyone setting up openaps manually can still instantiate their aps directory without the flag, of course, and will still get all the benefits of git auditability.

This branch has been tested on a single rig.  Test procedure was:
 - clone and checkout https://github.com/openaps/openaps/tree/nogit, and install using `python setup.py develop`
 - clone and checkout https://github.com/openaps/oref0/tree/nogit, and install using `npm run global-install`
 - remove ~/myopenaps and ~/myopenaps-cgm-loop
 - re-run oref0-setup via oref0-runagain.sh

We still need to test other upgrade paths, ensure that a version of `openaps` with --nogit support is installed, etc., and then do more extensive testing to make sure nothing else breaks.

Requires https://github.com/openaps/openaps/pull/130

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "oref0",
-  "version": "0.5.2",
+  "version": "0.6.0-dev",
   "description": "openaps oref0 reference implementation of the reference design",
   "scripts": {
     "test": "make test",


sulkaharo:
So, for the setup script, would the changes be to 1) disable git by default, including the git fix scripts and 2) ask user if she wants to run git in the advanced features menu and if yes, install the fix scripts to cron? I'm assuming it's going to continue to be beneficial to support git for development testing.

scottleibrand:
I would prefer to remove the git complexity entirely, unless someone can come up with a use case where we need it in oref0.  I don't share the assumption that it's useful for development any longer, so I'd like to hear if any developers actually make use of it before deciding to keep it as an option.

PieterGit:
@scottleibrand : I think git can be usefull, because it makes it reproducible of what exactly happened at some period of time (ability to audit). With git you can exactly check what inputs the oref0/openaps used at any time. On rigs with eMMC I didn't have a lot of git errors. I think having an option to be able to audit every filechange in git is still usefull. 


I agree that it's good to to also be able to disable git, because it tends to lead to nonfunctioning rigs (e.g. SD-card errors on RPI3).  It would be better that if the git commit fails, the rig just keeps working.

I can think of three reasons to keep git option available:
- Auditability is quite a common requirement for security and safety critical systems. Git currently has that role in openaps, and although `/var/log/openaps` logs are far more readable and useful for the average users, they can't fully reproduce the state the oref0 algorithm made calculations.
- some openaps users save the git history. I think @jasoncalabrese does with his indy repos.
- audit-ability in case something really bad happens (major bug causing too much insulin in a very specific situation). I think the `/var/log/openaps` logs won't guarentee that you will find the root cause
- the git repository might be useful for simulation of new openaps/oref0 algorithms, but maybe also using openaps-data-commons might be a better alternative for those use cases.

I noticed we don't have any statement on auditability or logging in the openaps reference design.

scottleibrand:
I don't see anything in there indicating that git *has been* or *is* useful, just that it *might be*.  Have there been any specific situations where you have actually used the git repo in ~/myopenaps?

I know @jasoncalabrese uses git, but he doesn't use oref0-setup, so he wouldn't be affected by this change.

If we want to design for auditability, I think we should consider from first principles what would be the best solution, given where things are today.  I suspect that being able to enable fully verbose logging of more / all data files would be more robust than using `git`, for example.  Also, if this is going to happen, it would need to be driven by someone who actually wants to do this work.  In my opinion, the logging we already have is mostly sufficient, and is easy to extend if we find something else we want to log.

sulkaharo:
Well, even if we remove git altogether from the setup script, if we don't remove the issue removal scripts, this would mean that someone who wants to use git would have to 1) remove the `git = false`line from their openaps.ini and 2) schedule any wanted git fix scripts in cron. I assume this is doable to the few people who want to keep using git?

scottleibrand:
This branch now automatically installs the necessary branch of openaps.  To test:
Install the new branch:
```
git fetch; git checkout nogit; git pull
npm run global-install
```
Move your myopenaps directory to a backup location, and run oref0-runagain from there:
```
mv ~/myopenaps ~/myopenaps.bak
~/myopenaps.bak/oref0-runagain.sh
```
Verify that git is not in use in ~/myopenaps:
```
cd ~/myopenaps
git status
```
If you want, copy your preferences.json and autotune directory back over:
```
cp ~/myopenaps.bak/preferences.json ~/myopenaps/
cp -r -up ~/myopenaps.bak/autotune ~/myopenaps/
```

scottleibrand:
Merging to new 0.6.0-dev branch

PR #550, opened 2017-07-17
Don't use xdrip/glucose.json when old or empty

scottleibrand:
This moves the xdrip logic out of cron into its own shell script, and adds logic (identical to what we already have in get-ns-bg) to only copy the xdrip/glucose.json file over if it actually contains recent data.


scottleibrand:
This has not yet been tested, as I don't have xdrip.  Will need someone who does to do the following (preferably on a dev rig):

First, kill BT tethering and confirm with `tail -f ~/myopenaps/monitor/glucose.json` that you're seeing the glucose.json file get truncated (and not re-populated until the next ns-loop run).

```
cd ~/src/oref0
git pull
git checkout fix-xdrip
npm run global-install
~/myopenaps/oref0-runagain.sh
```
answer Yes to deleting the old crontab

Check if everything still works.  If so, kill BT tethering again and re-run `tail -f ~/myopenaps/monitor/glucose.json` that you're no longer seeing the glucose.json file get truncated.

scottleibrand:
@kcrcgm tested this, and it fixed her issue.  Does anyone else want to test it before I merge it to dev?

PR #557, opened 2017-07-20
Update version number to 0.5.2.

applehat:
Reflect latest release verison


applehat:
Being overly critical, but I was confused before checking the source when doing the update (saw a bunch of oref0 0.5.1 references in the NPM lines)

scottleibrand:
Just merged master back to dev to fix this. Good catch, thx.

PR #559, opened 2017-07-21
Add module needed for EdisonVoltage to work on jubilinux 0.2.0

scottleibrand:


scottleibrand:
This should fix https://github.com/openaps/oref0/issues/553

scottleibrand:
I also just confirmed that adding this module and rebooting does not negatively affect older versions of jubilinux: I can still run a full loop, including EdisonVoltage, without error.

scottleibrand:
@skubesch tested successfully with jubilinux 0.2.0 (https://github.com/openaps/oref0/issues/553#issuecomment-317111381), so I think this is good to merge.

PR #560, opened 2017-07-22
fix &current_basal_safety_multiplier typo

scottleibrand:
Found this while looking for what might be causing the `SyntaxError: /root/myopenaps/preferences_from_args.json: Unexpected token .` error some people are reporting on FB. 

This might be contributing to it, or it might be unrelated, not sure.  Will need someone who can reproduce the original problem to test this as a possible fix.

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -547,12 +547,12 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
 #fi
 
     cd $directory || die "Can't cd $directory"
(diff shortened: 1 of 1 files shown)


scottleibrand:
To test this fix, you'll need to:
```
cd ~/src/oref0/
git fetch
git checkout preferences_from_args-fix
git pull
npm run global-install
```
and then re-run oref0-setup.

scottleibrand:
Any objections to just merging this to test in dev?

PR #561, opened 2017-07-22
enable SMB/UAM for a full 6 hours after any carb entry

scottleibrand:
With shorter DIAs and large meals with undercounted carbs, COB often decays to 0 before carb absorption finishes, and UAM cuts off after DIA hours.  This preference, if enabled, would allow UAM to remain active and allow oref1 to SMB if needed for 6 hours after any carb entry, unless disabled by a high temp target.


scottleibrand:
Tested successfully on our rigs.  Ready to merge if no objections.

PR #565, opened 2017-07-23
use sudo to install it decocare with easy_install. This is required for installing it on rpi. 

PieterGit:
small fix. make decocare work on a rpi.
tested ok.
please merge.

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -523,7 +523,8 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
     fi
 
     # install decocare with setuptools since 0.0.31 (with the 6.4U/h fix) isn't published properly to pypi
-    easy_install -U decocare
+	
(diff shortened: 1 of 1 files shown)


PieterGit:
I think we do. 


I can think of three reasons:
1. Best practice: Fail the build when there is something unexpected wrong.
2. Keep the number of configurations of oref0 software to a minimum. Each 0.6.0 oref0 will always have decocare 0.0.31 and that version is used by all users. Keeping the number of combinations limited makes the troubleshooting easier (remember jubilinux 0.1.1 and jubilinux 0.2.0 bugs)
3.  It's very unlikely that install fails. I can think of two main reasons that it can fail: 
- out of diskspace. If an installation fails on lack of diskspace the user should know, because he/she will have a half installed decocare
- bug in the installer of decocare or easy_install. Bugs must be found an eliminated

Issue #552, opened 2017-07-18
ImportError: No module named spi_serial with 0.5.x and Non Explorer board


PieterGit:
@LilDucky reports and error with Carelink USB, RPi2 and MDT CGM
https://gitter.im/nightscout/intend-to-bolus?at=596d4fc81c8697534a342b07

```
MDT CGM configured; not waiting
ImportError: No module named spi_serial
```

I think this error will exist for all non explorer board (pump device not `/dev/spidev5.1`) users with 0.5.x.
Note that the error is harmless, and the pump init and communication should just work fine. But we should avoid false error messages if we can.

It's caused by this pull request of https://github.com/openaps/oref0/pull/498 of @Kdisimone 

I noticed it, but didn't come by to fix it properly. I think the follow options can solve this bug:

Option 1. short term: don't add `reset_spi_serial.py 2>/dev/null;` to `lib/oref0-setup/alias.json` for users that don't have pump device `/dev/spidev5.1`

Option 2. fix improve oref0_init_pump_comms.py so that it will always invoke the reset_spi_serial.py.
The code is there, but somehow   @Kdisimone  and @scottleibrand had problems that it didn't executed the reset_spi_serial in some/all cases. The `oref0_init_pump_comms.py` has the code and logic to invoke the reset_spi_serial, here: https://github.com/openaps/oref0/blob/dev/bin/oref0_init_pump_comms.py#L23-L38

Perhaps not checking for the import, but checking for the pump device is better and will fix so that the extra `reset_spi_serial.py 2>/dev/null;` can be removed.

Option 3. (that's where I was heading for). Rewrite `oref0_init_pump_comms.py` to a `openaps use pump init` command that will call reset_spi_serial.py logic and init WW-pump (if applicable). That code should also be called by openaps if you use `openaps mmtune`. I tried to get that in, but my Python / openaps skills did not succeed yet.

One possible problem with option 2 is that there is some Python2 / Python3 thing. I wrote `oref0_init_pump_comms.py` in Python 3 because Python 3 has far better options of killing processes after a timeout. Killing a hanging pump WW init can also be handled by the `killall -g openaps ` in the crontab, but that takes much longer.

Option 2 and 3 also get rid of the double reset_spi_serial.py call, that has no purpose.
I prefer option 3 is that is, because it's the cleanest python and most openaps way 
But it will require a new openaps release and will force users to upgrade their openaps version if they want to use oref0 and  `oref0-setup.sh`.

Thoughts?

[...]

PR #569, opened 2017-07-25
only reset_spi_serial.py if [[  == /dev/spidev5.1 ]]

scottleibrand:
This should help avoid scary warning messages for folks not using Explorer boards, per https://github.com/openaps/oref0/issues/552#issuecomment-316219411


scottleibrand:
Not yet tested: we need to verify that mmtune still works properly for both Explorer boards and non-SPI mmeowlink setups.

scottleibrand:
Merging to dev to test there.

PR #575, opened 2017-07-30
oref0-pump-loop reliability and debugging improvements

scottleibrand:
Add some additional output to show what oref0-pump-loop is doing, and improve retry logic in a few places.


scottleibrand:
This is based off of nogit (since that's what my rigs were running), so I'll plan to merge it to there shortly if we don't see any issues with it.

PR #576, opened 2017-07-31
Fix argument definition

dramageek:
It makes more sense for this variable to be the autotune directory, but the code is written now for it to be the openaps directory. If I knew how, I'd update lines 140-142 instead, but that's beyond my skills :)


scottleibrand:
Looks like you figured out how to update 140-142?

dramageek:
Yes, but it sounds like @PieterGit prefers this fix. Needs to be re-targeted to dev though?

scottleibrand:
Just updated the target to dev. Do we want this in addition to, or instead of, the one I just merged.

Issue #443, opened 2017-04-12
Some closely-timed boluses go missing from nightscout uploads


mgranberry:
I have noticed that some boluses "go missing" while running my pump in SMB (on a 715 if that matters).  Take this snippet of `pumphistory-zoned.json` for example:

```json
[
  {
    "_type": "Bolus", 
    "_description": "Bolus 2017-04-12T12:49:49 head[4], body[0] op[0x01]", 
    "timestamp": "2017-04-12T12:49:49-05:00", 
    "_body": "", 
    "programmed": 3.0, 
    "_head": "011e1e00", 
    "amount": 3.0, 
    "duration": 0, 
    "type": "normal", 
    "_date": "71314c0c11"
  }, 
  {
    "_type": "Bolus", 
    "_description": "Bolus 2017-04-12T12:47:53 head[4], body[0] op[0x01]", 
    "timestamp": "2017-04-12T12:47:53-05:00", 
    "_body": "", 
    "programmed": 0.2, 
    "_head": "01020200", 
    "amount": 0.2, 
    "duration": 0, 
    "type": "normal", 
    "_date": "752f4c4c11"
  }
]
```

There are two distinct boluses here, both within a couple minutes of each other.  They are not associated with a Bolus Wizard event or any other metadata, but the second and more significant bolus is being treated as ancillary to the microbolus.  I have tracked the offending code down to `oref0/lib/bolus.js`, where the state from the first bolus gets overridden by the second bolus in an attempt to combine multiple records and only a single bolus event gets generated, but I don't know enough about the Medtronic record format to know how to properly resolve this.  I am thinking that a bolus event should be generated whenever state would be overridden.  Will that cause unforeseen problems?

scottleibrand:
Flagging this a blocking bug that needs to be fixed before releasing dev to master.

cjo20:
What exactly is that code meant to be doing? It looks like it's finding the first bolus from any records grouped closer together than 4 minutes and ignoring the rest.

scottleibrand:
It was intended for deduplication of boluses entered into NS and also executed on the pump.  We have since switched to a "we don't support that" stance, so if we implement stuff like https://github.com/openaps/oref0/issues/419 to clean up the remaining stragglers there, we can probably remove this deduplication check.

scottleibrand:
[...]

PR #579, opened 2017-08-01
Fix handling of closely-timed boluses

mgranberry:
**Why is this needed?**
Boluses, especially with SMB, may arrive spaced very closely together.  The current bolus pre-processing will combine these with bolus wizard records to generate synthetic records containing portions of the various records.  Instead of adding additional boluses that occur with a bolus wizard matching window (5 minutes), it simply uses the first that occurred after a bolus wizard record.  That was incorrect behavior.
**What is this doing?**
The bolus-wizard matching window is being shortened to 2 minutes, which is long enough to avoid some nightscout deduplication behavior that can lead to missed records.  Within the 2-minute window, any boluses are added together so the totals match reality, even if the event sequence doesn't exactly match pump history.  The total amount of insulin delivered will be correct.
**Fixes: [#443](https://github.com/openaps/oref0/issues/443)**

### DIFF
--- a/lib/bolus.js
+++ b/lib/bolus.js
@@ -9,7 +9,7 @@ function reduce (treatments) {
   function in_previous (ev) {
     var found = false;
     previous.forEach(function (elem) {
-      if (elem.timestamp == ev.timestamp && ev._type == elem._type) {
+      if (elem.timestamp === ev.timestamp && ev._type === elem._type) {
         found = true;
       }
     });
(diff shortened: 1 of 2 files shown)


scottleibrand:
If no objections, I'd like to merge this to dev for testing for a 0.5.3 release before we start merging new features (like exponential curves) for an eventual 0.6.0 release.

PR #582, opened 2017-08-04
simply calculate biobContrib using 2t instead of t

scottleibrand:
Rather than figure out how to modify the IOB formula, we can simply set t = t * profile.bolussnooze_dia_divisor and then use the same formula again with the new value of t.  This will cause us to find the point twice as far along the IOB curve, achieving the same things as we currently do for the bilinear curves, of making bolus snooze IOB decay twice as fast as "real" IOB.

Needs testing.


sulkaharo:
LGTM, I'm willing to merge right away

scottleibrand:
Any idea what's up with the test failures? https://travis-ci.org/openaps/oref0/jobs/260860258

sulkaharo:
Silly coveralls, coverage decreases because the amount of code is reduced. :)

PR #586, opened 2017-08-08
Calling out the number one beginner error

danamlewis:
Number one error for beginners that blocks looping is when people don't have the pump set to absolute u/hr (in % basal type instead), which prevents the rig from being able to set temp basals.

(Note: needs testing)

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -32,7 +32,7 @@ smb_main() {
     prep
     if ! ( \
         prep
-	# checking to see if the log reports out that it's on % basal type, which blocks remote temps being set
+        # checking to see if the log reports out that it is on % basal type, which blocks remote temps being set
         if grep -q '"Temp":"percent"' monitor/temp_basal.json; then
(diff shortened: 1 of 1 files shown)


sulkaharo:
Should this also just stop the loop from executing, so it's immediately obvious something is not right?

danamlewis:
I'd rather they get the full logs still, as there's often other things from the logs they need to fix, too, vs. exiting if they don't have the temp mode set right.

scottleibrand:
If no objections, I'd like to merge this to dev for testing for a 0.5.3 release before we start merging new features (like exponential curves) for an eventual 0.6.0 release.

scottleibrand:
Merged this to 0.6.0-dev. If everything looks good there I'll merge to dev as well for eventual 0.5.3 release.

scottleibrand:
Temp refreshed: monitor/temp_basal.json: {"duration":0,"rate":0,"temp":"percent"}
SMB pump-loop failed. Pssst! Your pump is set to % basal type. The pump won’t accept temporary basal rates in this mode. Change it to absolute u/hr, and temporary basal rates will then be able to be set.
Unsuccessful supermicrobolus pump-loop at Sun 13 Aug 21:39:13 MST 2017

PR #589, opened 2017-08-11
if currentDeviation is null or maxDeviation is 0, set mealCOB to 0

scottleibrand:
This should help eliminate one source of zombie carbs by forcing mealCOB to 0 until we have enough data to properly calculate deviations.

### DIFF
--- a/lib/meal/total.js
+++ b/lib/meal/total.js
@@ -67,6 +67,17 @@ function diaCarbs(opts, time) {
     // set a hard upper limit on COB to mitigate impact of erroneous or malicious carb entry
     mealCOB = Math.min( profile.maxCOB, mealCOB );
 
+    // if currentDeviation is null or maxDeviation is 0, set mealCOB to 0 for zombie-carb safety
+    if (typeof(c.currentDeviation) === 'undefined' || c.currentDeviation === null) {
+        console.error("Warning: setting mealCOB to 0 because currentDeviation is null/undefined");
+        mealCOB = 0;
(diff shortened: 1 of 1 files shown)


scottleibrand:
@tynbendad you might want to test this.

tynbendad:
awesome, will do!  i think i just need to pull from bad-deviation branch, will try that tonight before he heads to 2ndary rig.

scottleibrand:
Yeah, if you're already on dev, that should be sufficient.  Otherwise, you'll want to do `npm run global-install` as well after pulling that branch.

scottleibrand:
Merged this to 0.6.0-dev.  If everything looks good there I'll merge to dev as well for eventual 0.5.3 release.

tynbendad:
it did not fix the problem:
{"carbs":98,"boluses":10.7,"mealCOB":98,"currentDeviation":null,"maxDeviation":0,"minDeviationSlope":0}
mealCOB is still 98...  as is COB used for microbolus,etc.

root@benaps2:~# grep -n "currentDeviation === null" ~/src/oref0/lib/meal/total.js
71:    if (typeof(c.currentDeviation) === 'undefined' || c.currentDeviation === null) {

scottleibrand:
Let's jump on Gitter for some real-time troubleshooting if you can.

tynbendad:
summary from gitter: after running  npm run global-install from src/oref0 the fix appears to be working correctly.

PR #568, opened 2017-07-25
Exponential curves

sulkaharo:
Implements #544 - needs testing help, do not merge until confirmed to work.

**NOTE: consider testing this extremely experimental!** This code branch implements the new exponential curve model for insulin activity. This is a fairly big change compared to the old bilinear model, so if you do participate in the testing of this feature, assume you will need to monitor your BG closely and might need to adjust your profile to fit.

The code now supports three new configuration variables in preferences.json:

* curve, with legal values of blank, "bilinear", "rapid-acting" and "ultra-rapid"
  * If left blank or set to "bilinear", OpenAPS uses the old insulin activity model
  * If set to "ultra-rapid", the new curve model analyzed from publicly available data for Fiasp from Novo Nordisk is used, set to reach activity peak at 55 minutes
  * If set to "rapid-acting", the new curve model analyzed from publicly available data for Novorapid (Novolog) from Novo Nordisk is used. This curve should match well to Novorapid, Novolog, Humalog and Apidra and is set to peak at 75 minutes
* insulinPeakTime, with legal values from 35 to 120. The value defines the minutes from bolus to peak. Note if you don't set it, the above mentioned peak defaults are used.
* useCustomPeakTime, which must be set to "true" for the customized insulinPeakTime value to take effect

In order to use the rapid-acting or ultra-rapid curves, your pump must be set to 5 hour DIA, or longer. 6 hours is recommended as the starting point.

The underlying mathematical model is developed by @dm61.

Below are the activity and IOB graphs at various DIA and peak values

act new

iob new

**Testing**

Both Fiasp and Novorapid curves are modeled after 5 hour DIA (compared to 3 hour default DIA for the old model). Due to the curve shape, the insulin activity contribution at the 5 hour mark is extremely low but should model the actual insulin tail more effectively than the old model, so it'd be extremely beneficial to get testing help from users who currently have their DIA set to a lower value (3 or 4 hours), who would set their DIA to 5 hours with the new curve model and report back if OpenAPS correctly keeps the basal rates in balance.

For Fiasp users, please start testing with DIA = 6, curve = "ultra-rapid" and do not set insulinPeakTime. If it looks like Fiasp is peaking for you faster or slower than OpenAPS expects (similar to your DIA being shorter or longer in the old parlance), please adjust insulinPeakTime by a small amount (such as 5 minutes) from the default of 55.

**Note on Fiasp**

[...]

PR #592, opened 2017-08-12
Merge 0.6.0-dev to dev for eventual 0.6.0 release

scottleibrand:


tim2000s:
image001
image001
Stats and AGP for the first seven days of using this branch. Given the profound lack of bolusing and the tuning period, these results are remarkable. Range Used is consistent with clinical studies.

tim2000s:
I've noticed that the upload to Nightscout shows all insulin given by oref1 as SMB in the OpenAPS pill as Basal IOB. Would it make sense to separate out Basal IOB from meal specific Bolus IOB in the pill?

scottleibrand:
How exactly are you thinking we'd separate them out?

The only real function of the bolus vs. basal IOB is to decide whether to bolus snooze. If you recategorized SMBs as boluses, that would trigger bolus snooze, which would sometimes prevent oref1 from zero temping after microbolusing.

tim2000s:
It was more an information point. Most people are used to a bolus/background model, and this changes the thinking paradigm yet again. It may simply be that we need to document what's going on carefully.

tim2000s:
I've just updated to include the latest commits. I've been noticing some slightly frustrating patterns, but I'll wait to comment until I've run with this for a couple of days.

scottleibrand:
This is ready to merge to dev whenever @danamlewis is satisfied that we have things well documented enough for general testing.

tim2000s:
I think we need a lot of warnings in there. Over the past few days I've been seeing a lot of post prandial rises of 110+ mg/dl. It's almost always the same phenomenon and pretty much always comes with food with a higher fat content.  

When a meal is announced, the 50% load gets done quite quickly, then it pauses whilst there appears to be a period of no increase in glucose level. I then see a later absorption, for which there is not enough IOB, and the kick up starts. This is often at the point where the Fiasp peak has passed by about 30-60 mins, and as the earlier dose has not started to bring glucose levels back down, the new rise has a higher starting point. 

I've also found that when I stack carbs (with about an hour between them) it often struggles. I'm using SMBBasalMins=90 at the moment.

scottleibrand:
Warnings on what? We won't be encouraging people in the documentation to try no-bolus just because they're using 0.6.0. If people want to try reducing or eliminating boluses in certain situations, they certainly can. They'll undoubtedly still get better outcomes with partial manual boluses, but that is a tradeoff against usability.

tim2000s:
I think it's worth being aware of the kinds of questions that will get asked in looped and how those should be handled.

tim2000s:
Adding notes for 0.6.0 dev update relating to Exponential curves:

**Exponential Curves**
The code now supports three new configuration variables in preferences.json to allow exponential curves that better reflect the effects of insulin:

- curve, with legal values of blank, "bilinear", "rapid-acting" and "ultra-rapid"

1. if left blank or set to "bilinear", OpenAPS uses the old insulin activity model

2. If set to "ultra-rapid", the new curve model analyzed from publicly available data for Fiasp from Novo Nordisk is used, set to reach activity peak at 55 minutes
3. If set to "rapid-acting", the new curve model analyzed from publicly available data for Novorapid (Novolog) from Novo Nordisk is used. This curve should match well to Novorapid, Novolog, Humalog and Apidra and is set to peak at 75 minutes

- insulinPeakTime, with legal values from 35 to 120. The value defines the minutes from bolus to peak. Note if you don't set it, the above mentioned peak defaults are used.
- useCustomPeakTime, which must be set to "true" for the customized insulinPeakTime value to take effect

In order to use the rapid-acting or ultra-rapid curves, your pump must be set to 5 hour DIA, or longer. 6 hours is recommended as the starting point.

The details of the curve calculations and discussions can be found PR #568

scottleibrand:
If your pump DIA is set to less than 5h, and you turn on exponential curves, oref0 will use 5h as their DIA without complaining.

tim2000s:
**Enhanced SMB (eSMB)**

The default behaviour for SMB is that the max bolus that can be delivered is no greater than 30 mins of basal insulin. An additional preferences value, "maxSMBBasalMinutes", has been added to allow SMB to deliver a different amount of insulin as an SMB. This gives the ability to make SMB more aggressive if you choose. As with standard SMB, it is triggered by temp targets, carb entries and a pump bolus.

Increasing "maxSMBBasalMinutes" will allow the SMB functionality to deliver more insulin, earlier in the SMB process. 

"maxSMBBasalMinutes" can be added into `myopenaps/preferences.json`. It is recommended that the value is set to start at 30, in line with the default, and if you choose to increase this value, do so in no more than 15 minute increments, keeping a close eye on the effects of the changes. 

It is not recommended to set this value higher than 90 mins, as this may affect the ability for the algorithm to safely zero temp. It is also recommended that pushover is used when setting the value to be greater than default, so that alerts are generated for any predicted lows or highs.

tim2000s:
I'm running a clean install from flashing the Edison through installing everything, including totally fresh Bluetooth. Will let you know how it goes.

tim2000s:
Clean install worked very well. From nothing to 0.2.0 + OpenAPS fully looping in about 45 mins.

danamlewis:
TY @tim2000s for the starter notes and for the confirmation of clean installs. merging to dev for further testing.

PR #593, opened 2017-08-12
speed up COB calculation

scottleibrand:
We have been doing a bunch of unnecessary future IOB calculations in the COB calculation subroutine when all we really need at that point is the current IOB.  This passes a flag to get_iob() to tell it not to bother with all that when we're just going to throw away the results anyway.

My `time openaps report invoke monitor/meal.json` run (on an Edison with lots of treatments) went from 1m59s to 1m17s with this change.  Everything still seems to be working normally after a few completed loops.

### DIFF
--- a/lib/determine-basal/cob-autosens.js
+++ b/lib/determine-basal/cob-autosens.js
@@ -148,7 +148,7 @@ function detectSensitivityandCarbAbsorption(inputs) {
         iob_inputs.clock=bgTime;
         iob_inputs.profile.current_basal = basal.basalLookup(basalprofile, bgTime);
         //console.log(JSON.stringify(iob_inputs.profile));
-        var iob = get_iob(iob_inputs)[0];
+        var iob = get_iob(iob_inputs, true)[0];
         //console.log(JSON.stringify(iob));
 
         var bgi = Math.round(( -iob.activity * sens * 5 )*100)/100;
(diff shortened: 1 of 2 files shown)


scottleibrand:
This now incorporates https://github.com/openaps/oref0/pull/603 which speeds up COB calculation by another 5x

jsmurray1:
I've implemented the fixes identified here and tested on a Pi Zero W running oref0-0.5.2 I ran 10 pump loops before and 10 pump loops after.  

Before this fix:
 *3 out of 10 loops errored out 
*average loop completion time: 5:51
*Worst Time: 10:35
*Best Time: 0:37 (erroneously bad data? no BG data was listed in pump-loop.log)

After this fix:
*2 out of 10 loops errored out
*average loop completion time: 3:33
*Worst Time: 08:13
*Best Time: 0:41 (erroneously bad data? no BG data was listed in pump-loop.log)

I wonder if the really short times are due to overlapping loops? I did a ps -aux at one point and had three /bash pump-loop.sh processes running.

[02-pump-loop.log.after593.txt](https://github.com/openaps/oref0/files/1221089/02-pump-loop.log.after593.txt)
[01-pump-loop.log.before593.txt](https://github.com/openaps/oref0/files/1221088/01-pump-loop.log.before593.txt)

scottleibrand:
You shouldn't have any `openaps pump-loop` running in your cron, only `oref0-pump-loop`.  If you have the former, you'll need to re-run oref0-setup and say yes to deleting your crontab and re-creating it.  The cron job for oref0-pump-loop should prevent it from running any duplicates.

PR #598, opened 2017-08-12
import openaps install scripts from docs

scottleibrand:


scottleibrand:
This will need a corresponding update to the docs.

danamlewis:
I see no reason not to go ahead and merge this?

PR #601, opened 2017-08-13
don't SMB unless bg > threshold

scottleibrand:
https://github.com/openaps/oref0/blob/smb-lgs/lib/determine-basal/determine-basal.js#L564-L568 skips low glucose suspend if IOB is already super negative and BG is rising faster than predicted.  However, while it's useful to be able to avoid extended zero temping in that situation, it's best to avoid SMBs until BG has risen above the LGS threshold.


ericfeibelman:
working great so far for a full 24 hours. I like this new logic for SMB ...

` // min_bg of 90 -> threshold of 65, 100 -> 70 110 -> 75, and 130 -> 85
    var threshold = min_bg - 0.5*(min_bg-40);`

PR #607, opened 2017-08-14
make carbsReqThreshold configurable

scottleibrand:


scottleibrand:
To test:
```
cd ~/src/oref0
git fetch; git checkout carbsReqThreshold
npm run global-install
```
And then add carbsReqThreshold to your preferences.json, with the integer value of how many carbs you want to alert on.

Issue #590, opened 2017-08-12
preferences.json contains all defaults


thebookins:
[This code](https://github.com/openaps/oref0/blob/dev/lib/profile/index.js#L41-L60) is designed to expose only some of the configurable settings to the user through `preferences.json`. However, on a new setup from `dev` all default settings are in `preferences.json` (e.g. new or rarely touched settings like `enableSMB_after_carbs` and `remainingCarbsCap`). Having all of these settings in preferences could be confusing for users unaware of their effects. According to @scottleibrand the issue is at https://github.com/openaps/oref0/blob/dev/bin/oref0-setup.sh#L575.

Bender1061:
my workaround for this issue is I use WinSCP, I move the Preferences.json to my computer first, then after the Putty shows the preferences being made during the install, I then just copy it back to the rig. Right now we would need to made a script to check to make sure all the preferences are in the preferences.json, as it has grown by a lot since we first started. so we can't just see if the file is there and not overwrite.

tim2000s:
Easiest thing to do is include a line in the setup script that asks if we want to save existing preferences and if so copy it to a safe place then copy per intent parts back.

Bender1061:
That would probably work, but what about the people that are currently running the same stuff from before oref1?

scottleibrand:
Anyone care to test https://github.com/openaps/oref0/pull/609 and see if it gives the expected behavior?

PieterGit:
I think I'm -1 on this change, because it violates "Explicit is better than implicit", https://www.python.org/dev/peps/pep-0020/

What's the point of hiding defaults to users?

scottleibrand:
Because users are always asking "what should I set this to" and wanting to change things they shouldn't be changing.  This is a way to allow for things that we'd otherwise hard-code to be configurable, without making users thing they *should* configure them.

scottleibrand:
(Most of the rest of Zen of Python is trumped by "practicality beats purity" IMO.)

PieterGit:
The example of `enableSMB_after_carbs` is a feature that I think you can ask users to enable or disable.
The SMB-docs suggest: "You may want to experiment with turning only one of them on at a time so you can closely observe the behavior", http://openaps.readthedocs.io/en/master/docs/Customize-Iterate/oref1.html?highlight=turning#how-to-turn-on-smb-uam

If we want immutable preferences, we should not call them prefences but constants and implement them as something else, just like a Java `static const` (don't know the Javascript name for it).

[...]

PR #609, opened 2017-08-14
only updatePreferences with displayedDefaults

scottleibrand:
Per https://github.com/openaps/oref0/issues/590, preferences.json is still displaying all defaults in some cases.  This would force it to only add in displayedDefaults.  If someone wants to change a non-displayed default, they'll have to add it to preferences.json manually.


scottleibrand:
@thebookins any opinion here?

Issue #600, opened 2017-08-12
Some openAPS treatments appearing as future dated in Nightscout reports for Australia timezone


mel2095:
The issue is not affecting all openAPS generated entries - just ones that are created from the afternoon period onwards.

thebookins:
I looked into this a while back - at the time I came to the conclusion that it was because OpenAPS uploads treatments in local time (i.e. +10 in Sydney). Might be a change required to https://github.com/openaps/oref0/blob/dev/bin/nightscout.sh to upload using a timestamp converted to UTC?

mel2095:
Did you see it on all entries - I just have it happening with treatments from early afternoon and evening?

thebookins:
Only with treatments after 2 pm. During daylight savings it's treatments after 1 pm. Which is consistent with the time difference of +10/+11. Is that what you see?

At the time I we were switching over from Loop to OpenAPS. Loop treatments displayed fine, which suggested that it was the OpenAPS upload procedure that was at fault - not NS itself.

mel2095:
Yes that sounds the same as what I see - it's been like it since I setup in Dec. We haven't tried loop to compare it with.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

PieterGit:
@thebookins :the fix must not be in the `nightscout.sh`, but the `created_at` in `mm-format-ns-treatments.sh` must be converted to UTC time. See
https://github.com/openaps/oref0/pull/426#issuecomment-287545072

thebookins:
Thanks @PieterGit , I knew we'd been over this before. I see that your PR wasn't pulled; do you think that `created_at` still needs to be converted to UTC?

PieterGit:
Yes, we need the JSON command to convert `created_at` to UTC, e.g.
 `new Date(created_at_timestamp).toISOString()`

PieterGit:
@thebookins I just created a pull request https://github.com/openaps/oref0/pull/612
that fixes the issue... 

BUT BE CAREFULL: I'm seeing duplicate carbs because the mongodb will have both the `created_at ` in UTC format and the `created_at` in local format (+0200 for  me, +1100 for you). I think some/all of the carbs are not seen as duplicates.

@scottleibrand do you know a proper way to add the necessary unduplication code, so that carbs/treatments that are in TIMESTAMP+TZ format are ignored by oref0.

scottleibrand:
[...]

PR #612, opened 2017-08-14
Fix for Some openAPS treatments appearing as future dated in Nightscout reports for Australia timezone 

PieterGit:
Fix for https://github.com/openaps/oref0/issues/600

Nightscout stores treatments in UTC format. Nightscout find uses string compare for $gte, so:
- all treatments `created_at` fields must be set in UTC format
- all treatments must be queried in UTC format

I noticed also that `settings/temptargets.json` is queried in local time and not using a UTC, which I believe is a bug, because Nightscouts stores the `created_at` for temp targets in UTC format and Mongo `$gte` compare compares strings (and doesn't use the timezone).

This also fixes problems with false positive future treatments in Nightscout.

Installation
```
cd ~/src
mv oref0 oref0.openaps
git clone https://github.com/PieterGit/oref0.git
cd oref0
git checkout 201708_treatments
npm run global-install
rerun oref0-setup.sh or ~/myopenaps/oref0-runagain.sh
```

Please test on a seperate spare rig. 

The branch is based on the dev / 0.5.3 branch.

Pleas test: `openaps ns-meal-carbs`, and temp targets and Nightscout pump events.
Please check carb entries.

### DIFF
--- a/bin/mm-format-ns-treatments.sh
+++ b/bin/mm-format-ns-treatments.sh
@@ -27,7 +27,7 @@ model=$(json -f $MODEL)
 
 oref0-normalize-temps $HISTORY  \
   | json -e "this.medtronic = 'mm://openaps/$self/' + (this._type || this.eventType);" \
-    -e "this.created_at = this.created_at ? this.created_at : this.timestamp" \
+    -e "this.created_at = new Date(Date.parse(this.created_at ? this.created_at : this.timestamp)).toISOString()"\
     -e "this.enteredBy = 'openaps://medtronic/$model'" \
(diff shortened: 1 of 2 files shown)


PieterGit:
WARNING! 
Somehow I see a too high carb count, probably caused because the mongodb can have duplicate entries (one for the created_at with timezone, and one for the created_at in UTC format). These entries don't seem to get counted as duplicates. This needs to be looked into.

PieterGit:
Don't use this PR. It can double carbs during upgrade of the rig. 
A solution in Nightscout or the temporary fix of https://github.com/openaps/oref0/pull/614 is preferred. Closing this PR.

PR #614, opened 2017-08-15
Fix for missing temp targets and carb treatments for timezones east of UTC

thebookins:
This PR is a temporary fix to address an issue with temp targets and carb histories extracted from NS. There is an inconsistency with the `created_at` field where entries made through Care Portal have `created_at` in UTC, while entries uploaded from oref0 have `created_at` in local time. The simple string compare used in the query is not timezone aware. (See https://github.com/openaps/oref0/issues/600 for discussion on this.)

 - `bin/nightscout.sh` has been modified to use UTC time in the query string for both `temp_targets` and `carb_history`
 - `lib/oref0-setup/report.json` has been modified to use a lookback of 18 hours for `monitor/carb_history.json` (so that users at UTC -12 will still get 6 hours of data for entries with `created_at` in local time)

Note that I'm assuming temp targets are all entered through care portal and don't need the 18 hours treatment. This is a temporary fix; the more robust solution is to ensure that `created_at` is always in UTC.

Needs testing for various timezones.

### DIFF
--- a/lib/oref0-setup/report.json
+++ b/lib/oref0-setup/report.json
@@ -18,7 +18,7 @@
       "use": "shell",
       "reporter": "JSON",
       "device": "ns",
-      "remainder": "-6hours",
+      "remainder": "-18hours",
       "json_default": "True"
     },
     "name": "monitor/carbhistory.json"
--- a/bin/nightscout.sh
+++ b/bin/nightscout.sh
@@ -292,11 +292,11 @@ ns)
     ;;
     temp_targets)
     expr=${1--24hours}
-    exec ns-get host $NIGHTSCOUT_HOST treatments.json "find[created_at][\$gte]=$(date -d $expr -Iminutes)&find[eventType]=Temporary+Target"
+    exec ns-get host $NIGHTSCOUT_HOST treatments.json "find[created_at][\$gte]=$(date -d $expr -Iminutes -u)&find[eventType]=Temporary+Target"
     ;;
     carb_history)
     expr=${1--24hours}
-    exec ns-get host $NIGHTSCOUT_HOST treatments.json "find[created_at][\$gte]=$(date -d $expr -Iminutes)&find[carbs][\$exists]=true"
+    exec ns-get host $NIGHTSCOUT_HOST treatments.json "find[created_at][\$gte]=$(date -d $expr -Iminutes -u)&find[carbs][\$exists]=true"
     ;;
     *)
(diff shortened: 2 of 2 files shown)


scottleibrand:
Do we need both of these changes?  It seems that the addition of `-u` to `date` would shift you back 11 hours to UTC, so you shouldn't then need to also go back 18 hours more.  Have you tested it with just the `-u` addition to see if that is sufficient?

thebookins:
`-u` will be fine for me at +10 or +11. It won't be fine for anyone west of UTC, hence the 18 hour lookback for the hypothetical case of UTC -12. The `-u` will mean that carbs entered through NS will be fine with just a 6 hour lookback for anyone anywhere, it's just that there are pump-derived carbs as well with local timestamp.

Having said that, we can assume that all pump-derived carbs are known locally on the rig, and are then de-duped, so perhaps we don't care if some records with local timestamp are missed?

scottleibrand:
Note: this fix requires re-running oref0-setup.  If you just update your repo, NS carbs will be ignored.

scottleibrand:
This seems to be working on UTC-0700.  I'm ready to merge it to dev if no objections.

lib/oref0-setup/report.json
```diff
@@ -18,7 +18,7 @@
       "use": "shell",
       "reporter": "JSON",
       "device": "ns",
-      "remainder": "-6hours",
+      "remainder": "-18hours",
```
PieterGit: This can be set back to `-4hours`, once a Nightscout with this PR is used https://github.com/nightscout/cgm-remote-monitor/pull/2763 . In fact it depends what we expect of the maximum temp target time.

I think it's fine to stick to `-18hours` for the 0.5.3 release, but maybe we should urge people in the docs to upgrade their Nightscout once the next cgm-remote-monitor version (with my PR included) is released.

PieterGit:
Please see my review comment https://github.com/openaps/oref0/pull/614/files#r135363639

lib/oref0-setup/report.json
scottleibrand: I'm fine with encouraging people to run the latest NS, but is there any real downside to downloading extra carbhistory or temptargets?  As far as I can tell oref0 correctly ignores any carb treatments older than 6h, so there doesn't seem to be any functionality change or performance hit from downloading extra.  I'm not sure whether the longer download would allow for temp targets longer than 6 hours, but I'm OK with that either way...

Issue #619, opened 2017-08-18
Allow larger mealtime SMBs


scottleibrand:
In order to prevent post-meal lows, it would be good to be able to reduce the size of mealtime boluses when autosens is detecting sensitivity (autosens ratio < 1).  With #618 oref0's mealtime dosing will be adjusted for that, but it might also be good to make it easier for PWDs to avoid having to do such calculations themselves, and instead allow them to do a smaller initial meal bolus and let oref1 handle the difference as appropriate for each situation.

Once #618 is merged, we might also want to:
 - Reduce default remainingCI time from 4h down to 3h, effectively reducing the "default carb absorption time" accordingly, and allowing oref1 to cover more of a meal without always first assuming the meal with absorb slowly and cause a low around 3h afterward.  For example, 90g of COB would be assumed to absorb at 30g/hr over 3h instead of 22.5g/hr over 4h.  In the event of a large slow (pizza-type) meal, the timeframe over which the remaining COB are predicted to absorb would continue to be extended out to 3h from the current time until actual carb absorption ramps up enough to predict absorption of the remaining carbs before then.
 - Allow oref1 to be configured with a maxBolus value, and use that (rather than 30m worth of basal) when IOB is too low to cover more than a fraction (perhaps 2/3 or 3/4) of COB.  This would allow oref1 to more quickly SMB when a meal is under- (or not) bolused for.  For basals up to 1.5 U/h, a reasonable maxBolus value might be 1 U.
 - Consider allowing oref1 to SMB for 1/2 of insulinReq (vs. 1/3 normally) when IOB is too low to cover more than half of COB.  This would allow it to more quickly get insulin activity going when carbs are entered but no bolus is given.  For example, if IOB is 0, COB is 60g, and insulinReq is 3U, then oref1 could dose up to 1.5 U (or maxBolus, whichever is lower), and continue to dose half of the remaining insulinReq every 5m until IOB got up to 3U, at which time it would slow down to 1/3 of insulinReq every 5m.
 - In both of the latter cases, we could consider the SMB-delivered meal boluses as regular boluses that count towards bolus IOB rather than counting them toward basal IOB as we do with smaller SMBs.  Large SMBs could occur right after carb entry (until IOB catches up to COB), so the SMB time period would not be extended much.  This treatment might trigger bolus snooze in some cases, but less so than a (larger) manual bolus would.

[...]

PR #620, opened 2017-08-19
model remainingCarbs as a /\ shaped bilinear curve

scottleibrand:
Per https://github.com/openaps/oref0/issues/619#issuecomment-323494069, current oref0 code models any carb absorption that hasn't yet begun as a constant value over the next 4 hours, such that, for example, a 120g meal would have a remainingCI equivalent to a constant 30g/hr for the next 4 hours.

This changes remainingCI from a constant to a bilinear curve that increases from 0 to a peak absorption rate (in g/h) of remainingCarbs/2, and then symmetrically decreases back to 0 at the 4h mark.

So if you entered 120g of carbs and had zero absorption so far, this change would cause oref0 to model absorption starting at 0, increasing to 30 g/h after 60m (1h), 60 g/h after 120m (2h), then back to 30 g/h after 180m (3h), and 0 g/h after 240m (4h). The average carb absorption over hours 1 and 4 would therefore be 15 g/h, and 45 g/h for hours 2 and 3.

As actual carb absorption is observed, the extrapolation of the current rate of carb absorption, linearly decreasing just fast enough to use up all the carbs by the time absorption hits zero, would gradually take over, and the future predicted carb absorption dominated by the /\ shaped remainingCI bilinear curve would transition to a \ shaped linear one.

This change would have a few main benefits. Firstly, it would eliminate the sudden jumps and drop-offs we see with a constant remainingCI. Today, if a 120g meal is entered w/o observed carb impact, we assume that carb absorption will instantly ramp up from 0 to 30 g/h, and then instantly drop off from 30g/h to zero after 4h. That is obviously unrealistic. Secondly, it would better match predicted CI to what we actually would expect to observe from a typical meal, resulting in more accurate cobPredBGs. And thirdly, it would continue to allow us to "push out" the predicted remainingCI past 4 hours when current CI is not rising as fast as predicted, while allowing CI predictions to be "pulled in" when carb absorption is actually observed.

One other effect of this change is that it predicts a slower rise from newly-entered carbs than the previous constant remainingCI.  Since insulinReq is currently based on the insulin required to get the minimum predicted BG (from 60m out to ~4h into the future) down to target, this means that this change actually decreases the insulinReq immediately after meal carb entry.  To counteract this, I've changed the start time for calculating minCOBpredBG from 60m to 90m.

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -328,7 +328,7 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     //5m data points = g * (1U/10g) * (40mg/dL/1U) / (mg/dL/5m)
     // duration (in 5m data points) = COB (g) * CSF (mg/dL/g) / ci (mg/dL/5m)
     // limit cid to 4 hours: the reset goes to remainingCI
-    cid = Math.min(4*60/5,Math.max(0, meal_data.mealCOB * csf / ci ));
+    cid = Math.min(4*60/5/2,Math.max(0, meal_data.mealCOB * csf / ci ));
     acid = Math.max(0, meal_data.mealCOB * csf / aci );
     // duration (hours) = duration (5m) * 5 / 60 * 2 (to account for linear decay)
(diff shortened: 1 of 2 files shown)


scottleibrand:
This is working well in real world testing. I think it's ready to merge to 0.6.0-dev if no objections.

PR #621, opened 2017-08-19
microBolus 1/2 the insulinReq if IOB doesn't yet cover COB

scottleibrand:
Per #619, this would allow oref1 to SMB for 1/2 of insulinReq (vs. 1/3 normally) when IOB is too low to cover COB. This allows it to more quickly get insulin activity going when carbs are entered but no bolus is given. For example, if IOB is 0, COB is 60g, and insulinReq is 3U, then oref1 could dose up to 1.5 U (or maxBolus, whichever is lower), and continue to dose half of the remaining insulinReq every 5m until IOB got up to 6U, at which time it would slow down to 1/3 of insulinReq every 5m for any additional insulin required (i.e. for a carb undercount detected via UAM).

### DIFF
--- a/bin/oref0-autotune-export-to-xlsx.py
+++ b/bin/oref0-autotune-export-to-xlsx.py
@@ -137,9 +137,9 @@ def excel_init_workbook(workbook):
 # sort filenames. First on date and then on run number
 # put settings/profile.js
 def sortedFilenames():
-    filelist=glob.glob("../settings/profile.json")
-    filelist=filelist+glob.glob("../settings/pumpprofile.json")
-    profiles=glob.glob("profile*.json")
+    filelist=glob.glob("settings/profile.json")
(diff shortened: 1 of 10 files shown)


scottleibrand:
This PR now also includes a configurable value for maxSMBBasalMinutes, which is used to set a (presumably higher) maxBolus as long as IOB does not yet cover COB.  Once there is enough IOB on board to cover COB (usually as UAM takes over) it goes back to using 30m of basal as the maxBolus to avoid overreacting to noisy CGM data or a short sharp rise.

scottleibrand:
After testing, I've updated the PR to allow insulinReq/2 boluses (of up to maxSMBBasalMinutes' size) until IOB is enough to cover COB.

scottleibrand:
Testing on this has been very successful, so I'm ready to merge it to 0.6.0-dev if no objections.

PR #622, opened 2017-08-20
Move Edison refs to Explorer IF/Then section

jsmurray1:
An attmept to reduce problems on Pi based rigs by moving Edison references to the Explorer If/Then section of the setup.  Also edited openaps-install.sh to add a user check for user edison before attempting to adduser edison to sudo and dialout groups.  Please take a close look at this as this is my first attempt at modifying code and using the Github workflow.


jsmurray1:
Ok, I'll take another stab at fixing it later this week.  Obviously, I didn't look that hard at the code and don't understand what I'm doing.

scottleibrand:
Ready to merge this to dev as soon as someone tests it (or is ready to do so in dev).

PR #628, opened 2017-08-31
fix monitor-xdrip

thebookins:
Local (offline) looping hasn't been working on my rig since the latest update from dev. Turns out the script is called from crontab with extension `.sh` (`monitor-xdrip.sh`), which cannot be found. This commit removes the filename extension in consistency with other bash scripts called from crontab.

Note: this fix requires re-running `oref0-setup.sh`.

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # This script sets up an openaps environment by defining the required devices,
-# reports, and aliases, and optionally enabling it in cron, 
+# reports, and aliases, and optionally enabling it in cron,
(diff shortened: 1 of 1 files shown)


dramageek:
I'd noticed an error in mine, but figured it was being caused by something else. I ended up copying `/src/ore0/bin/monitor-xdrip.sh` into my `myopenaps` folder. This seems more like the proper way to fix it!

Tested your the fix, and confirmed that it works. @scottleibrand can we still get this into 0.5.3?

scottleibrand:
Yes, this kind of bug fix is very appropriate for 0.5.3.  Good catch: thanks for the fix and to both of you for testing it promptly.

PR #629, opened 2017-09-02
COB-aware autosens

scottleibrand:
This updates autosens to use the COB and UAM categorization logic from autotune to exclude meal-related deviations from autosens.  The autosens output has been updated to display in oldest to newest order (not backwards, like it previously did) and to display COB-related meal periods in parentheses following the COB amount, and UAM-related meal periods as u().

Example output:

`+++-60g(>>>>>>>>>>>>>>>>>>>>>>>>>>>>>)u(>>>)--------30g(>>>>>>>>>>>>>>>>>>>>>)------15g(>>>>>>>>>>>>>>>)-----------++---+60g(>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>17>>>>>>>>>>)----+++-++-+----+-+-+-++++++--++-+---++-----++++++-+++++--+--+---+++++++++-+-++>-++-++-++-+--++-----+---+++--+--+--++>-++-----+++-- 54% of non-meal deviations <= 0 (target 45%-50%)
Excess insulin sensitivity detected: ISF adjusted from 51.843 to 56`

This will need some testing to determine whether the 45%-50% target is still good, or if that needs adjustment.  I also split out the cob-autosens.json into two files, cob.js and autosens.js, which both need cleaning up to remove code and conditionals that are no longer needed in both.


scottleibrand:
This also still preserves the 6 mg/dL/5m cutoff for detecting unannounced and unbolused carb rises: we'll want to evaluate whether that threshold should be raised now to reduce false-positives.

scottleibrand:
Is anyone interested in building any unit tests to cover autosens.js?  I'm pretty happy with the real-world testing I've done so far, and inclined to ignore the coverage % decrease unless someone volunteers to write some tests.

scottleibrand:
This has been running well for several days and is ready to merge if no objections.

sulkaharo:
Should the 6mg/dL threshold be configurable? My experience is the kid hormonal rises easily go over this.

Another thing to think about for later is could we somehow support marking down rescue carbs to increase the accuracy of the analysis, without affecting the dosing.

scottleibrand:
Possibly, or it may be safe to remove it. I am testing the latter in a new branch, and can make a new PR if it works well. 

What Dana does for marking rescue carbs is use an IFTTT trigger what both enters carbs and sets a high temporary target to prevent any SMBing (either from the COB, or equally likely from UAM after the COB is expired).

PR #631, opened 2017-09-03
mmtune on failure if pump_loop_completed > 15m old

scottleibrand:
If a rig hasn't mmtuned or completed a pump-loop recently and fails to loop, it's probably best to do an mmtune.  Setting "recently" to 15m for now.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -348,7 +348,7 @@ function mmtune {
 }
 
 function maybe_mmtune {
-    if ( find monitor/ -mmin -15 -size | egrep -q "mmtune|pump_loop_completed" ); then
+    if ( find monitor/ -mmin -15 | egrep -q "mmtune|pump_loop_completed" ); then
         # mmtune ~ 25% of the time
         [[ $(( ( RANDOM % 100 ) )) > 75 ]] \
         && echo "Waiting for 30s silence before mmtuning" \
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -348,16 +348,16 @@ function mmtune {
 }
 
 function maybe_mmtune {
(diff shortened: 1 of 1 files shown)


bin/oref0-pump-loop.sh
```diff
@@ -348,11 +348,17 @@ function mmtune {
...
-    && wait_for_silence 30 \
-    && mmtune
+    if ( find monitor/ -mmin -15 -size +5c | egrep -q "mmtune|pump_loop_completed" ); then
+        # mmtune ~ 25% of the time
+        [[ $(( ( RANDOM % 100 ) )) > 75 ]] \
+        && echo "Waiting for 30s silence before mmtuning" \
+        && wait_for_silence 30 \
+        && mmtune
+    else
+        echo "mmtune.json and pump_loop_completed more than 15m old; waiting for 30s silence before mmtuning"
```
PieterGit: according to the the find command this should be `mmtune.json or pump_loop_completed more than 15m old`

PieterGit:
will test it on a 0.6.0-dev rig

bin/oref0-pump-loop.sh
scottleibrand: if find mmtune or loop_completed <15m old, randomize. 
else, both mmtune and loop_completed are >15m old, so mmtune.

scottleibrand:
In order to cut down on how many times an out-of-range rig can attempt to loop before mmtuning and going into 76-second wait-for-silence, I switched it to always mmtune if the last pump_loop_completed was more than 15m ago, regardless of how recent mmtune.json is.

scottleibrand:
This is working well on all of our rigs, and reduces interference from the ones that are just outside their usable range.  I think it's ready to merge to dev for final testing before 0.5.3 release, if no one sees any issues with it.

scottleibrand:
This doesn't seem to break anything, so merging it to dev for @mhaeberli to test.

PR #634, opened 2017-09-04
Allow nonzero low temps for SMB

scottleibrand:
Rather than oscillating back and forth between a 30m zero temp and a neutral temp, this allows SMB to set an intermediate low temp when less than 30m of zero temping is required.


scottleibrand:
This has been running well for 24h and is good to merge if no objections.

PR #578, opened 2017-07-31
Autotune tokenauth

scottleibrand:
Follow-up to https://github.com/openaps/oref0/pull/563

Please post test results here so we can get this merged to dev.

PR #636, opened 2017-09-04
URL encoding for autotune filters

scottleibrand:
The change in #578 to use ns-get instead of directly using curl broke the time based filters that autotune uses to limit its data download, because ns-get uses curl's `-g` flag.  This caused autotune in 0.5.3 dev to run extremely slowly (due to all the extra treatments) and to erroneously autotune with multiple days' worth of data each day.


scottleibrand:
Turns out removing -g does break things, so I URL-encoded the `[` and `]`'s as %5B and %5D in the autotune URLs instead.

scottleibrand:
This has been tested from cron and is ready to merge to dev if no objections.

PR #640, opened 2017-09-06
carbs: should only include carbs actually used in calculating COB

scottleibrand:
Since the original mealAssist days, the `carbs` value in meal.json, also known as `mealCarbs` in some files, has contained the total number of carbs seen in the last 6 hours or so.  When using this value in determine-basal to blend COB and UAM, it would be better if `carbs` value only contained those carbs used in calculating the current COB value, and ignored carbs from previous meals that had fully expired before the current meal (and weren't stacked on top of it).

This change removes from `carbs` any carb values that don't end up getting used in COB calculation.

### DIFF
--- a/lib/meal/total.js
+++ b/lib/meal/total.js
@@ -35,6 +35,7 @@ function recentCarbs(opts, time) {
         return bDate.getTime() - aDate.getTime();
     });
 
+    var carbsToRemove = 0;
     treatments.forEach(function(treatment) {
         var now = time.getTime();
         // consider carbs from up to 6 hours ago in calculating COB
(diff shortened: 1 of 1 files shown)


sulkaharo:
👍 I was wondering about that, should have asked. :)

PieterGit:
👍  I also wondered about the variable names. Shouldn't carbs named `carbs_on_board` to make the meaning more clear?

scottleibrand:
@PieterGit No, in meal.json, `carbs` is the total carbs entered (now for the current meal, instead of the last 6h).  `mealCOB` is the carbs still on board.

sulkaharo:
So, total_carbs and carbs_on_board then? ;)

PieterGit:
If I understand @scottleibrand comments correctly I would propose `carbs_last_meal`, `carbs_last_6h` and `carbs_on_board`. Dropping `mealCOB` if it means `carbs_on_board` would be a very good idea for 0.6.x. That would also harmonize pump loop output and Nightscout. Showing it as COBin the reason field and Nightscout UI is fine by me.

scottleibrand:
If you're going to start renaming variables, you're going to very quickly end up with having to do so everywhere in the system, and dealing with dependencies on input and output json files as well. I'd say let's not do that ad hoc as part of something like this. If you want to make a comprehensive proposal for which variables need to be renamed, we can discuss and agree and then go make all the changes together, perhaps as their own patch release to avoid making the other diffs completely unreadable.

PR #644, opened 2017-09-07
Autotune: calculate CR directly

scottleibrand:
Currently autotune categorizes glucose deviations data into buckets based on what is impacting things the most: basals, carbs, or insulin.  It uses the mealtime data to calculate CSF, the IOB-without-COB data to calculate ISF, and then the remainder to adjust basals.  This works fairly well, but due to the fuzzy line between carb-dominated and insulin-dominated deviations after a meal, it tends to overestimate ISF.  CR is calculated as ISF/CSF, so that tends to be overestimated as well.

As an alternative, this change calculates CR directly.  To do so, we use the BG and IOB at the time of carb entry, the BG and IOB after COB==0 and IOB is back down to low level, the total amount of carbs given, and the total amount of insulin given over the meal period.  After adjusting for the BG and IOB changes, that allows a direct calculation of carb ratio as total carbs / total insulin at mealtime.

Additionally, this change adds some more error checking around bad carb ratios (falling back to the pump profile carb ratio if needed), and avoids setting new ISF, CSF, or CRs unless the new ones are valid.  These should help with the occasional situation where someone's autotune gets messed up and their rig stops looping just after midnight.

### DIFF
--- a/bin/oref0-autotune-prep.js
+++ b/bin/oref0-autotune-prep.js
@@ -22,7 +22,7 @@
 
 var generate = require('oref0/lib/autotune-prep');
 function usage ( ) {
-        console.error('usage: ', process.argv.slice(0, 2), '<pumphistory.json> <profile.json> <glucose.json> [carbhistory.json] [autotune/glucose.json]');
+        console.error('usage: ', process.argv.slice(0, 2), '<pumphistory.json> <profile.json> <glucose.json> [pumpprofile.json] [carbhistory.json] [autotune/glucose.json]');
 }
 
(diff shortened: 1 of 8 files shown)


scottleibrand:
Additionally, some hours of the day rarely have data to tune basals due to meals.
When we don't have any data to tune a particular hour's basals, we should adjust it slightly toward the average of the surrounding periods that do have data.  This smooths out the big fluctuations we otherwise see in mealtime basals and ensures that they are tuned based on the best data we have from non-meal times before and afterward.

PR #647, opened 2017-09-07
remove  variable

thebookins:
`monitor-xdrip.sh` fails sometimes because `$directory` is null. It looks to be set in the autotune script [here](https://github.com/openaps/oref0/blob/92919bde860f438556e916e5c6a2c355d5006194/bin/oref0-autotune.sh#L79) so will be undefined if autotune hasn't run.

All of the other `cp` commands in the script work fine without it; as should this one.


scottleibrand:
Have you tested this with `$directory` removed?  What were your results before and after?

thebookins:
I have tested this with `$directory` removed. Before this change glucose data wasn't making it from `xdrip` to `monitor`. Afterwards it was. @efidoman first identified the issue may want to weigh in.

thebookins:
Perhaps a better solution is to define `$directory` in crontab?

PieterGit:
The common way to do this is to start all scripts from the openaps dir and use relative dirs from there.
So afaict the PR looks fine (i don't use xdrip).

efidoman:
Agreed ... the directories in the script should use relative path. This change is good. I've tested it and it works now. I tested it by disabling wifi temporarily on my rig, connecting via the USB cable and looking at the files, the xdrip-loop.log, and the openaps log. It is offline looping like it should now with this change.

PR #649, opened 2017-09-10
Disable IPv6 for rig stability purposes

efidoman:


PieterGit:
Shouldn't this only be done for Linux kernel 3.x on an Edison board? I think there is no need to disable IPv6 on e.g. a raspberry Pi. This change won't apply it if the sysctl's are already set in `/etc/sysctl.conf`, e.g. `net.ipv6.conf.all.disable_ipv6=0`.

scottleibrand:
I've seen IPv6 local networking problems affect all of my rigs. Until we add IPv6 networking checks to oref0-online, I think it's most prudent to disable it.

sulkaharo:
Could this and the other changes made by the scripts be documented outside the script? As is the setup scripts make an increasing amount changes to the OS setup that I bet any users familiar with Unix will want to understand. I haven’t personally now run the setup script for months because I haven’t had the time to read it with thought and didn’t want to run anything that touched the networking on the rigs that already work...

scottleibrand:
Feel free PR to somewhere unobtrusive in the docs if you want, but I'm not in any hurry to change this on my working rigs either.

PR #655, opened 2017-09-20
Less aggressive UAM

scottleibrand:
In 0.5.3, UAM assumes that current deviations will only decrease to zero over 3 hours if the currentDeviation exceeds the maxDeviation over the last ~45m.  This means that the UAMpredBG purple prediction line goes way too far up-and-to-the-right when carb absorption is peaking.  As an alternative to make UAM safer / less aggressive in that situation, this change causes UAM to assume that current deviations will ramp down over time proportional to how quickly they've ramped up from the recent (~45m) minimum deviations.  So if deviations are rising rapidly due to fast carbs, UAM will not project that to continue indefinitely, but typically only for 1-2h.  For meals whose carb absorption ramps up more slowly, the minDeviation will be higher, and a slower ramp down will be projected as well.  Once deviations actually peak, the existing code to project the rate of decrease from peak takes over.  As a result, UAM predictions are now much more reasonable, while still serving as a useful guideline to safely dosing additional insulin when meal carbs are underestimated and BG rises more than expected.

### DIFF
--- a/lib/determine-basal/cob.js
+++ b/lib/determine-basal/cob.js
@@ -141,7 +141,7 @@ function detectCarbAbsorption(inputs) {
         if (i==0) {
             currentDeviation = Math.round((avgDelta-bgi)*1000)/1000;
             if (ciTime > bgTime) {
-                console.error("currentDeviation:",currentDeviation,avgDelta,bgi);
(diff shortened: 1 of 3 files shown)


scottleibrand:
This is well-tested and ready to merge to dev if no objections.

Issue #544, opened 2017-07-13
Support insulin curves to reflect realistic insulin curves and distinguish between regular fast acting and Fiasp


tim2000s:
Enhancement to iob/calculate.json to use real curves and allow a choice between Fiasp and standard fast acting during set-up. 

The curves are available at https://github.com/LoopKit/Loop/issues/388 

Potential issues include variation in peak time when scaling curves by DIA and when to cut off action. 

Also question whether we should have standard curves by default and require an enable flag to swap those out for fiasp or whether it should be included in the interactive setup. 

I'll submit a branch to allow users to test my Fiasp curves once tested over the weekend, but no setup configuration work has yet been done.

mgranberryto:
I have been using the referenced curves live for about 12h and have found a much better correlation with observed data over the last 2 weeks.  My modification to calculate.js, based on Tim's work, is available here:
https://gist.github.com/mgranberryto/e5116dbc314cdf78513f4d185684b8a5

sulkaharo:
My branch is at https://github.com/sulkaharo/oref0/tree/fiasp

Added an unit test, too. Will add more soon

tim2000s:
I'm running on the Fiasp curve successfully now for a couple of hours. Will report back when run for a longer duration. So far so good.

tim2000s:
Been running the Fiasp curve for the best part of a day now with a five hour DIA. It's definitely working more effectively than the older shape curve, and has the added benefit of not missing the tail and having the resultant stacking. All in all an effective change for me.

sulkaharo:
Just pushed a new version of my brach, which adds customizing the peak time OR DIA for Fiasp. See attacked graphs for what IOB and activity curves look at 45, 55 and 65 minute settings. 

![activity](https://user-images.githubusercontent.com/1001330/28492025-b60ee932-6f03-11e7-8afe-62bcc32943fa.png)

![iob](https://user-images.githubusercontent.com/1001330/28492026-ba4793dc-6f03-11e7-81f6-11098735c6d9.png)

tepidjuice:
I have a problem with the use of the function t^n \exp(-n*t/ \tau)  where n is a non-integer. I originally suggested that type of function as it is the response of a second/third order linear system with repeated roots to an impulse with 0 initial conditions. Such systems are commonly used to model insulin action.

[...]

PR #658, opened 2017-09-20
display curve in preferences by default

scottleibrand:
Since there are no strings displayed in preferences.json by default, it's non-intuitive to know how to add "curve".  Since we'll eventually want everyone to change over to an exponential curve (and most likely will eventually want to make "rapid-acting" the default and deprecate "bilinear"), we should include "curve" in displayedDefaults.

### DIFF
--- a/lib/profile/index.js
+++ b/lib/profile/index.js
@@ -59,6 +59,7 @@ function displayedDefaults () {
     profile.enableSMB_with_COB = allDefaults.enableSMB_with_COB;
     profile.enableSMB_with_temptarget = allDefaults.enableSMB_with_temptarget;
     profile.enableUAM = allDefaults.enableUAM;
+    profile.curve = allDefaults.curve;
 
     console.error(profile);
     return profile


pampopcorn:
I don't totally understand the difference between bilinear and rapid-acting.  Why have the 3 different ones now if the plan is to only have 2 options later?

scottleibrand:
https://github.com/openaps/oref0/issues/544#issuecomment-317187355 has a good visualization of the difference between a bilinear and exponential curve.  The rapid-acting one is pretty similar to that orange graph (just with a 75m peak instead of 65m), but as you can see that is way different from the yellow bilinear one.  The rapid-acting and ultra-rapid curves are probably closer to each other than either is to bilinear.

scottleibrand:
We do plan to deprecate bilinear eventually and make rapid-acting the default, but for now we want people to explicitly enable and test it so we can confirm it works better for everyone before changing the default and deprecating bilinear.

PR #662, opened 2017-09-21
Project remaining carb absorption time based on % carbs absorbed

scottleibrand:
Currently 0.6.0-dev relies on the user to set an eating soon to reduce the remaining carb absorption time for not-yet-absorbing carbs from 4h to 3h.  However, what we really want is for it to assume the shorter absorption timeframe when carbs are first entered, but *not* assume quick absorption at the tail end if actual absorption is not being observed.

This change makes the remainingCATime a function of what % of entered carbs have been absorbed vs. remain as COB.  When the carbs are first absorbed, remainingCATime is set to 3 hours.  At 50%, it'll use 4h.  And as COB approaches zero, remainingCATime will approach 5 hours.

This also disables the previous behavior of changing remainingCATime based on temp targets.


scottleibrand:
Ready to merge this if no objections.  @danamlewis ?

PR #663, opened 2017-09-21
Refresh pumphistory before setting temps and after SMB/enact

scottleibrand:
We currently allow oref0-pump-loop to try to set a temp basal before it tries to refresh pumphistory if it hasn't completed a pump loop recently, as a way to make sure that we can enact temp basals based on updated BG data even when pump comms are flaky.  However, this results in high temps being set and then immediately canceled when an SMB has been delivered that the rig doesn't know about yet.

This change skips the "setting temp before refreshing pumphistory" code path if the currently running temp basal has a duration of more than 20 minutes.

This also makes oref0-pump-loop refresh pumphistory after bolusing or enacting a temp, so we don't have to wait 5 minutes for it to be displayed in NS.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -223,7 +223,7 @@ function smb_bolus {
 }
 
 function refresh_after_bolus_or_enact {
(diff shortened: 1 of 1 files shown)


scottleibrand:
Working well.  Ready to merge if no objections.

PR #664, opened 2017-09-21
autotune/autosens on by default; don't ask about ~/myopenaps

scottleibrand:
... and remove some outdated warnings about SMB


scottleibrand:
Ready to merge if no objections.

PR #665, opened 2017-09-21
x12 automation

danamlewis:
Adding x12 prompt to the setup script to ease the way of x12 users. If x12 is a "yes", it will grab the raw files needed, and at the end of the setup script, it opens the basal file for editing. Users will still want to edit the settings file in order to adjust maxBasal or set DIA differently, and edit the targets file to adjust targets.

- [x] TODO: don't ask about SMB if they answer Y to x12 in oref0-setup


danamlewis:
(Instructions stored in the docs; these won't be merged to live until 0.6.0 nears release: https://github.com/openaps/docs/pull/1047)

Issue #630, opened 2017-09-02
No pump mmtune on SMB error for four consecutive failing pump loop runs


PieterGit:
@ecc1 @scottleibrand Follow up to https://github.com/openaps/oref0/pull/591#issuecomment-326703920 in a separate issue.

See screenshot of the pump loop:

![image](https://user-images.githubusercontent.com/6500826/29999135-8a2063ea-903d-11e7-9064-8e68dceccc81.png)

I changed the `RANDOM` to `$RANDOM` in `maybe_mmtune` about 23:57, causing the mmtune to run (in two consecutive runs) and fixing the rig-pump communication.

The odds of running mmtune after 4 unsuccusful pump runs should be  (25% ^ 4)  = 0,391 %, so 1 out of 256 tries. I suspect it's a bug in the maybe_mmtune logic.

PieterGit:
Issue with 0.5.3-dev

scottleibrand:
No, it's only supposed to mmtune 25% of the time. Your calculation is the odds of not tuning with 75% odds. The odds of not tuning after 4 runs with 25% odds each time are (1-0.25)^4, or 32%

I've already raised the maybe_mmtune probability to 30% in 0.6.0-dev. Maybe we should add some sort of check on the last mmtune timestamp and always mmtune if it's been more than 15m since the last one?

scottleibrand:
Opened https://github.com/openaps/oref0/pull/631 to do that.  If you'd like that merged to 0.5.3 before release, please test and comment.

PieterGit:
I just switched to a 0.6.0-dev checkout on a different rig. I think your patch will improve it, but I think the mmtune logic can be improved. That won't make it in time for 0.5.3, so I'll propose new things (and a new issue / PR) when I worked some thing out.

jaylagorio:
So is this in 0.6.0? I think I've noticed that after rebooting the rig it tries to loop three times and then does an mmtune on the fourth time through. I've done more than the usual number of reboots the past couple days and this has been pretty consistent. Successful loop before a reboot, and this right after reboot:

```
Starting supermicrobolus pump-loop at Fri Sep 22 21:56:02 EDT 2017 with 5 second wait_for_silence:   
Waiting up to 4 minutes for new BG: .glucose.json newer than pump_loop_completed                     
Radio ok. Listening: .No pump comms detected from other rigs                                         
Preflight mmeowlink.exceptions.CommsException: No acknowledgement from pump on wakeup. Is it out of r
[...]

PR #666, opened 2017-09-23
Remove pump_loop_completed on reboot to make mmtune run sooner

jaylagorio:
As noted in https://github.com/openaps/oref0/issues/630, I've never seen mmtune run sooner than on the fourth time through the loop after rebooting. Removing `pump_loop_completed` on reboot makes the loop run mmtune on the first loop, decreasing non-looping time from about 25 minutes after reboot to a little under 10 minutes.


PieterGit:
@jaylagorio @scottleibrand : I see two other solutions that I think are cleaner than adding another crontab line.

1. rewrite `pump_loop_completed`  to `/tmp/pump_loop_completed`. That way the cleaning isn't necessary because `/tmp` gets cleared by default on a reboot

2.  rewrite the `@reboot reset_spi_serial.py` line to a single line for all the pump init and mmtune, e.g.
`@reboot cd $directory ; openaps use pump init  ; openaps use pump mmtune`

scottleibrand:
I like #1. One of you want to PR that?

jaylagorio:
The search feature wasn't being my friend so the only place I found reference to that file was `oref0-pump-loop.sh`. If that's the only place it's used: https://github.com/openaps/oref0/pull/672

PR #667, opened 2017-09-24
wait for bolus to complete and refresh pumphistory after SMB/enact

scottleibrand:
This makes oref0 refresh pumphistory after an SMB/enact so that it can be displayed more promptly in NS.


jaylagorio:
`only refresh after SMB/enact; wait_for_silence 10 if blousing` - would this not refresh after a temp? That's OK if so, just trying to clarify. Otherwise this one looks good to me!

scottleibrand:
Yes, that is intentional. It also refreshes any time a >30m low temp is still running, even if not new, which is IMO an acceptable side effect for simplicity's sake. If we wanted to instead to modulus math we could, but my first attempt didn't work so I decided not to.

PR #668, opened 2017-09-24
autotune de-biasing

scottleibrand:
This fixes an upward bias in the new CR autotuning from dinners that don't finish absorbing by midnight, and removes the upward bias on ISF from ignoring (slightly) positive deltas.

It also fixes a bug that was preventing ISF data from being capped by the min/max limits before being applied, and only capping it afterward.  Fixing that allows us to stop relying on the autotune ISF adjustment ratio hack, so it changes the default value to 1.0 (in preparation for removing it entirely at some future date).

CSF is also removed from the recommendations report to reduce confusion now that ISF and CR are both calculated directly (and somewhat independently).


scottleibrand:
This has now been tested against 9 individuals' NS data and seems to be giving much better results overall, and all of the results seem to converge on reasonable settings that should be safe to loop with.

At this point I'd be comfortable merging this to dev and suggesting everyone run it, and then PM me their URL etc. if they see results they don't think are right.

PR #671, opened 2017-09-24
print hour numbers in autosens output

scottleibrand:


scottleibrand:
This one is entirely cosmetic and way better than before, so I'll merge it shortly if I don't notice anything else I want to change.

PR #672, opened 2017-09-24
Put pump_loop_completed in /tmp/

jaylagorio:
Proposed changes to the location of pump_loop_completed from $directory/monitor/ to /tmp/ so it disappears after reboot. Please double check lines 367, 474, and 529 since I'm not super up on Bash scripting?


scottleibrand:
This looks good.  Do we want to move monitor/pump_loop_enacted as well for consistency?  Not sure it matters otherwise.

We can also revert the now-unnecessary #666 changes in this PR if you want...

jaylagorio:
Yeah, keeping them both in /tmp/ is consistent and negates the need for the new cron job. Changes made!

PR #674, opened 2017-09-25
recalculate IOB and predBGs after each loop

scottleibrand:
Now that we're refreshing pumphistory after each loop, the only things remaining out of date until the next one are the IOB pill and purple prediction lines.  This recalculates those so they can be uploaded as well.


jaylagorio:
Looks good, more immediate feedback in Nightscout is always appreciated!

PR #677, opened 2017-09-26
add an enableSMB_always option

scottleibrand:
We often see non-meal-related rises that benefit from SMB/UAM correction.  For people who have been running SMB long enough to be comfortable running it overnight, this provides an option to enable such corrections 24x7 (except when overridden by a high temp target).  This is *not* recommended for new users, so it will not be exposed in displayedDefaults.

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -267,6 +267,9 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     // disable SMB when a high temptarget is set
     if (profile.temptargetSet && target_bg > 100) {
         enableSMB=false;
+    // enable SMB/UAM if always-on (unless previously disabled for high temptarget)
+    } else if (profile.enableSMB_always) {
+        enableSMB=true;
     // enable SMB/UAM (if enabled in preferences) for DIA hours after bolus
     } else if (profile.enableSMB_with_bolus && bolusiob > 0.1) {
         enableSMB=true;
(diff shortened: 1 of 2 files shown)


ceben80:
that'll be a great idea, for permanent SMB usage I often triggered with temp target

tim2000s:
Presume this is "enableSMB_always":"true" in preferences.json?

scottleibrand:
Yeah.  You can also confirm that by looking at the diff: https://github.com/openaps/oref0/pull/677/files

jaylagorio:
This one looks good to go

jaylagorio:
I like having SMB turn off for a big jump, but can the reason for being disabled on that run (for any reason) maybe go to the OpenAPS pill's hover text?

jaylagorio:
And what do you think about having the percentage be slightly higher? Maybe 15% is a good starting point? Thinking about my SMB-enabled meals lately, it's not uncommon for the first jump from being around ~100ish to be 11 or 12, then slowing down when it gets in the 150s, and capping out around 170s. 10% would slow the overall response time in at least my case. This is a hard one since it's all subjective and a judgement call.

scottleibrand:
Yeah, I think 10% might be a tad too low.  Do you see "real" 10-15% deviations for more than 10m after meals in your recent data?  If so, I'll probably raise the threshold from 10% to 15%.

jaylagorio:
I don't think so, but if so not for more than 15 minutes and so far not often.

scottleibrand:
Ok. Let's keep an eye out for it happening, and it if does we'll raise the threshold. I've already seen a few cases where 10% has helped avoid SMBing on false jumps from a noisy new sensor and from calibration, so there's benefit to having it as low as we can without messing up meal SMBs.

scottleibrand:
Gonna go ahead and merge this.  We can tweak further as needed with another PR to dev.

PR #679, opened 2017-09-27
randomize wait_for_silence before smb_verify_reservoir

scottleibrand:
I'm still seeing occasional instances where two rigs get all the way through to issuing small SMBs at the exact same time.  (SMBs larger than about 0.2U take too long to be an issue.)  This randomizes the wait_for_silence before smb_verify_reservoir to be $upto10s to try to make sure that other rigs are done talking before we do our reservoir check to see if any other rig has just SMB'd > 0.1U.


jaylagorio:
Also looks good to go

PR #682, opened 2017-09-29
allow UAM duration to extend to 4h

scottleibrand:
In order to limit UAM, we originally had a maximum of 3 hours over which UAM carb impact could be extrapolated.  Now that we're better able to limit UAM duration based on minimum deviation as well as maximum, extending that to 4 hours allows us to do a better job of projecting deviations from high and steady deviations from large meals.  Any increasing or decreasing deviations will still result in UAM durations shorter than 3h in most cases: this will just allow UAM to keep up with COB-based predictions when deviations aren't increasing or decreasing much over 45m.


scottleibrand:
Ready to merge if no objections.

scottleibrand:
I found a different approach I like better (in the blend-minGuardBG branch).  Closing this.

PR #683, opened 2017-09-29
look ahead an additional 15m to get insulin started faster

scottleibrand:
originally we were waiting 90m before setting minPredBGs in order to be sufficiently aggressive at bringing down BGs predicted to ramp up quickly from meals.  When experimenting with shorter remainingCA carb absorption times we shortened the minPredBG lookahead to 55m+15m=70m for Fiasp, leaving rapid-acting insulins at 75m+15m=90m.  This extends that another 15m, to 85m for Fiasp and 105m for RAIs, which will allow eSMB to be more aggressive at mealtimes, as seems to be necessary to accomplish no-bolus (carb guesstimate only) looping with Fiasp with 80% time in range.


jaylagorio:
This one looks good!

Issue #588, opened 2017-08-10
User definable ISF's


tepidjuice:
As per the discussion on gitter. I'd like to be able to give oref0 user-defined ISF's the use cases would be: 

- exercise, or lack of it, 
- illness
- large amounts of very slow acting food. 

For example say I plan to do medium intensity aerobic exercise then I'd like to be able to tell oref0 that my ISF will, e.g. double for one hour, in two hours time. It then calculates the amount of insulin I'd need to keep my bgl in range given this doubling. 

My preferred implementation would be a user inputted list:

`ISF_scale_factor, start_time, duration, time_to_return`

1. `ISF_scale_factor` -- the amount by which your `ISF` is increased so `new_ISF = basal_ISF * ISF_scale_factor`
2. `start_time` -- the time at which your `ISF` will change from `basal_ISF` to `user_ISF`
3. `duration` -- the length of time your `ISF` will be at `user_ISF`
4. `time_to_return` -- the length of time after `start_time+duration` that it takes your `ISF` to return to `basal_ISF` e.g. `ISF` would linearly return to `basal_ISF` over `time_to_return`. I think this would be most helpful for dealing with the after effects of exercise.

When initialised the algorithm would:
1. calculate the basal rate required to keep your bgl in range with the new ISF.
2. find the temp basal required to have the IOB at start time be the IOB you'd have if your basal rate was as in (1)
3. Do it's standard maths but with the new basal rate throughout the duration of the exercise
4. Periodically redo the above throughout the `time_to_return` period. For example this ramp or decay of ISF back to its basal value would happen over 8-24 hours for exercise.
5. Autosens would keep running but be ` +- autosens_ratio * new_ISF`.

Ideally (ignoring food and a slight raise/drop in bgl before the ISF actually changed) this would allow the bgl to be kept exactly on target rather than running higher or lower for exercise/laziness.

Also I think it means we would address the actual need for temp_targets during exercise.

Here's the back of the envelope calculations I propose:

Say you have a basal rate `x` u /hr (e.g. 1) which holds your basal bgl at `y` mmol\L (e.g. 5) with and ISF of `z` mmol\L\u (e.g. 3) then `xz = E` mmol\L\hr (i.e. 3) is your Endogenous Glucose Production (EGP), which we assume is constant. 

[...]

PR #684, opened 2017-09-29
Exercise mode

scottleibrand:
As described in https://github.com/openaps/oref0/issues/588 and https://github.com/LoopKit/Loop/issues/593, OpenAPS activity mode doesn't do enough (at reasonably low temp target levels) to handle extended aerobic exercise, and often overshoots if you set the temp target high enough to do so.  This implements a better approach, which uses the level of the temptarget to set a manual sensitivityRatio.  This is used the same way as the autosens ratio, to set the basal and ISF used for current decisions.  In addition, it also is used to set the basal used by IOB calculations, such that setting an activity mode raises OpenAPS' estimate of IOB, and causes it to zero temp long enough to get IOB down to a level more suitable for aerobic activity.

These extra adjustments kick in for temp targets > 110 mg/dL, and scale up as the target increases.  Some testing will be required to see if the formula used for scaling that up is appropriate or if it needs adjusting to better balance the basal and ISF adjustments against the actual BG target people want to run at during such exercise.

### DIFF
--- a/lib/determine-basal/autosens.js
+++ b/lib/determine-basal/autosens.js
@@ -148,6 +148,8 @@ function detectSensitivity(inputs) {
         avgDelta = avgDelta.toFixed(2);
         iob_inputs.clock=bgTime;
         iob_inputs.profile.current_basal = basal.basalLookup(basalprofile, bgTime);
+        // make sure autosens doesn't use temptarget-adjusted insulin calculations
+        iob_inputs.profile.temptargetSet = false;
         //console.log(JSON.stringify(iob_inputs.profile));
         //console.error("Before: ", new Date().getTime());
         var iob = get_iob(iob_inputs, true, treatments)[0];
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -94,10 +94,18 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     }
 
     var sensitivityRatio;
-    if ( profile.high_temptarget_sensitivity && profile.temptargetSet && target_bg > 110 ) {
+    var normalTarget = 100; // evaluate high/low temptarget against 100, not scheduled basal (which might change)
+    if ( profile.half_basal_target ) {
+        var halfBasalTarget = profile.half_basal_target;
+    } else {
+        var halfBasalTarget = 160; // when temptarget is 160 mg/dL, run 50% basal (120 = 75%; 140 = 60%)
+    }
+    if ( profile.high_temptarget_sensitivity && profile.temptargetSet && target_bg > normalTarget + 10 ) {
         // w/ target 100, temp target 110 = .89, 120 = 0.8, 140 = 0.67, 160 = .57, and 200 = .44
         // e.g.: Sensitivity ratio set to 0.8 based on temp target of 120; Adjusting basal from 1.65 to 1.35; ISF from 58.9 to 73.6
-        sensitivityRatio = 2/(2+(target_bg-100)/40);
+        //sensitivityRatio = 2/(2+(target_bg-normalTarget)/40);
+        var c = halfBasalTarget - normalTarget;
+        sensitivityRatio = c/(c+target_bg-normalTarget);
         sensitivityRatio = round(sensitivityRatio,2);
         process.stderr.write("Sensitivity ratio set to "+sensitivityRatio+" based on temp target of "+target_bg+"; ");
     } else if (typeof autosens_data !== 'undefined' ) {
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -94,7 +94,7 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     }
 
     var sensitivityRatio;
-    if (profile.temptargetSet && target_bg > 110) {
+    if ( profile.high_temptarget_sensitivity && profile.temptargetSet && target_bg > 110 ) {
         // w/ target 100, temp target 110 = .89, 120 = 0.8, 140 = 0.67, 160 = .57, and 200 = .44
         // e.g.: Sensitivity ratio set to 0.8 based on temp target of 120; Adjusting basal from 1.65 to 1.35; ISF from 58.9 to 73.6
         sensitivityRatio = 2/(2+(target_bg-100)/40);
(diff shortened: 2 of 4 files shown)


danamlewis:
Some future documentation notes:

How temporary targets impact operation of OpenAPS:

- For SMB, a temporary target of 99 or lower, can enable SMBs (when the option for SMBs from targets is turned to true). A temporary target of 101 (or higher) will disable SMBs.
- However, a temporary target of (anything less than normal target) will trigger the traditional "eating soon" mode for temporary basal rates - even if target is normally 130, and ES target is set to 100, that will trigger more basal (but per above, no SMB's, since that is >99).
- Historically, a high temp target solely caused the system to adjust the target and calculate behavior from there. (i.e. a normal target of 100, an 'activity mode' temp target for 2 hours of 140 means that OpenAPS would target 140 instead of 100 for those two hours, then revert upon canceling or expiration to targeting 100). Now, (*needs a name new exercise mode behavior*) when enabled means that for a target of 111 or higher will create a manual sensitivity ratio, thus adjusting ISF and basals in proportion to this adjusted target. The goal of this is to recalculate IOB down to zero sooner ahead of and during activity, when the human knows this increased sensitivity is coming.

drnoname82:
We just went on a tiny hike for science ;) Exercise mode did very well and the behaviour was much closer to what we would have done manually. IOB at the beginning was 0.5u with Fiasp/ultra-rapid curve/DIA=6h. The recalculation of IOB (0.5u->2.9u) , which led to long low temping, proved very useful.

In my opinion the relation between temporary target and sensitivity ratio (`sensitivityRatio = 2/(2+(target_bg-100)/40)`) makes very (too) high targets necessary. When adjusting manually for hiking, we have always had good results adjusting the basal rate to 50% for easy hikes and to 30% for more strenuous ones and approximately doubling CR, aiming for a target around 140.

Achieving the same in exercise mode with the current constant of 40 in the formula requires us to set a temp target of 180 for moderate hikes. At least for us a constant of 20 or 25 feels much closer to what we do manually. Also only such constants make it possible to calculate sensitivity ratios of ca. 0.3 which we find typical for more intense exercise.

Does oref0 provide a protection against endless low temping? Otherwise extensive low temping in exercise mode might lead to ketoacidosis.

activity mode

exercise constant

scottleibrand:
Thanks, good analysis.  I think lowering the constant to get even more aggressive low-temping in exercise mode will be good to do eventually.  I'd be interested in thoughts as to whether we should make such adjustments all at once in a single release, or start with a higher constant at first and lower it as people get comfortable with the idea...

We have chosen not to put any limitations in place on duration of low temping, as preventing low BG is more important than preventing low levels of ketones.  We should continue to recommend that people eat during activity if the activity is strenuous enough to keep BG levels low/falling despite long periods of low temping.  Dana has found that eating enough while hiking to bring BG levels up a bit and allow OpenAPS to administer some insulin helps prevents an icky ketoney feeling she otherwise gets by the end of a long hike.

drnoname82:
If exercise mode becomes an optional preference instead of the default behavior (at least during introduction), I think it would be ok to start with a low constant immediately. The current approach is not a problem during mild exercise, only during more strenuous activity, so relatively aggressive low-temping should be the goal for the new behavior in my opinion.

scottleibrand:
Seems reasonable. Any other thoughts from others with experience low temping for activity as to what target BG level should correspond to a 50% reduction in basal? To me, 180 does seem a little too high, but 140 doesn't give a lot of wiggle room to shoot for a higher target without overshooting for milder exercise...

jbwittmer:
We also target 170-180 for strenuous, prolonged exercise (long hiking, long cross country practice, etc)

drnoname82:
Do you really wish to have a BG of 170-180 or do you set this target to get OpenAPS to do what you want it to do?

jbwittmer:
@drnoname82 No we don't really aim for 170-180 but find it necessary to set this target to get OpenAPS to low-temp sufficiently.  And usually take extra carbs as well.

scottleibrand:
@jbwittmer so the question at hand is: if you have an activity strenuous enough that it would require basal to be cut in half, what would you like BG to be riding at?  If you could set a 160, 140, or 120 temp target and achieve that, which would give you the right amount of extra buffer for the activity?

tim2000s:
That's tricky. Often for the type of exercise I do, I don't want a high target as I initially drop and then rise as I go in to anaerobic exercise and put stress on my body. For me 120 would be a max.

jbwittmer:
Sorry if I didn't read this thoroughly the first time.  For a 50% basal and strenuous activity we would require a higher temp target, and we have our basal settings pretty well dialed in for "normal activity". For example for mountainous hiking we would typically set a temp target of 160 and dial basal down to 30% or so.  Less strenuous activities like heading to the water park, etc where you're active but not continuously so, we would use a 50% basal and target 140 or so.

scottleibrand:
If no objections, I'm leaning toward changing the 40 to 30 such that 50% basal corresponds to a 160 target instead of 180. We can lower it further from there if testing justifies.

tepidjuice:
Thanks Scott. I'm keen to test this.

A small suggestion. Could the formula be simplified from: 
2/(2 + (target - normal)/30) 
to: 
1/(1 + (target - normal)/60) 
or:
60/(60 + target - normal).

scottleibrand:
@tepidjuice good idea.  I went two steps further and made the formula calculate the constant halfBasalTarget - normalTarget, and then made the constant a variable by setting halfBasalTarget in preferences.json.  It now defaults to 160 mg/dL = 50% basal (instead of 180 mg/dL).

danamlewis:
Do we want to also have this adjust carb ratio accordingly?

danamlewis:
Based on offline convo w/ @scottleibrand and some other-branch-testing we did - leaving carb ratio off from here for now. Let's get the community testing as-is into dev now, with adjusting basals, ISF, and target and see how that does, and can decide and tweak any carb adjustments for this mode later.

danamlewis:
Tested successfully. Merging to dev for further changes. See the dev PR for updated notes on behavior and the variable names for preferences. https://github.com/openaps/oref0/pull/707

drnoname82:
The introduction of a variable is a great idea. I just tested this feature in dev, however changing `"half_basal_exercise_target"` in preferences does not lead to recalculation of IOB.  Could it be that a reference to `"half_basal_target"`instead of `"half_basal_exercise_target"` in [https://github.com/openaps/oref0/blob/0fa1e00b79c18af2c873f4582fe236cd10a8bb31/lib/determine-basal/determine-basal.js#L98](url) is the cause? (same in history.js - or maybe I just don`t understand why there are two different variables)

scottleibrand:
@drnoname82 good catch.  PR to fix is at #712

PR #687, opened 2017-10-02
Fix update of Decocare 

Bender1061:
Fix errors on updating to 0.1.0-dev on decocare

Very common error I get when updating an rig is 
```
Continue? y/[N] y
Checking openaps 0.2.1 installation with --nogit support
openaps 0.2.1-dev
Checking /root/johnopenaps: /root/johnopenaps already exists
Searching for decocare
Reading https://pypi.python.org/simple/decocare/
Best match: decocare 0.0.32.dev0
decocare 0.0.32.dev0 is already the active version in easy-install.pth

Using /root/src/decocare
Processing dependencies for decocare
Finished processing dependencies for decocare
/root/src/decocare/ already exists; pulling latest 0.1.0-dev
error: pathspec '0.1.0-dev' did not match any file(s) known to git.
Couldn't pull latest decocare 0.1.0-dev
```

and that kicks you out of finishing the setup.

So instead of trying to fetch the correct version of decocare it's just easier to remove it, then clone it back in, and away we go. 

Yes there are many ways to use git, but these are the simple ones that always works for me

<insert @scottleibrand  favorite git comic>

anyone see any problem with that?

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -543,7 +543,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
     # TODO: remove this and switch back to easy_install or pip once decocare 0.1.0 is released
     if [ -d "$HOME/src/decocare/" ]; then
         echo "$HOME/src/decocare/ already exists; pulling latest 0.1.0-dev"
-        (cd $HOME/src/decocare && git fetch && git checkout 0.1.0-dev && git pull) || die "Couldn't pull latest decocare 0.1.0-dev"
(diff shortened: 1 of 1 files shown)


scottleibrand:
This approach would remove any local changes that someone might be testing: avoiding that is the point of the whole "does the directory already exist" check.

When you next come across a rig that fails on that step, can you run the steps manually and see if you can get any more details on why it failed?  If the `git fetch` succeeds, you shouldn't get the `'0.1.0-dev' did not match any file(s) known to git` error.

If we do end up taking this approach, we should only do so if the fetch and pull method fails.

Bender1061:
i have another rig that I believe has not been upgraded yet, I'll test that, but ti would not be till next week sometime, as I'm near Toronto this weekend.

Bender1061:
I have been unable to reproduce this error. and since it does not sound like anyone has been having it, I'm going to close this.

PR #688, opened 2017-10-02
install correct locale en_US.utf8

CaptainBalou:
ask user to install correct locale to fulfill OpenAPS prerequisites. Better solution would be installation without user intervention. Follows if there is more time.


scottleibrand:
I don't think we want to force newbies into `dpkg-reconfigure locale` - it's extremely non-intuitive.  I would rather leave things as-is until we figure out the best method to install the necessary locales without user intervention.

I'm also not sure why exactly this is necessary to add to the default setup, as the jubilinux and raspbian images we recommend come pre-installed with an English locale that supports the proper numeric separators for autotune.  Obviously anyone who doesn't want to interface with their rig in English would need to add their own language settings, but I don't think that needs to be in the install script.

scottleibrand:
Closing for now.  Please re-open if you'd like to continue work on it.

PR #650, opened 2017-09-11
Prep for eventual 0.6.0 dev to master release

danamlewis:
**Note**: this next release will be a big release (from 0.5.3 to 0.6.0 to match the significant changes).

**WARNING:** This code is highly experimental and has many changes compared to 0.5.3. Do not implement this branch for testing unless you review the diff and plan to watch it closely. Also, do not change behaviors until you test the defaults of this branch. *There may be some things not yet documented here that need to be documented; please comment on this thread or in channel to help note what needs documenting.* **Make sure you follow this PR so you will get notifications about changes to this branch. It is still changing feature-wise, so you will want to update as new features and bug fixes are added in.**

**See the bottom-most comment for how to install and run 0.6.0.**

### CHANGED: **Default behavior change for carb absorption**
In 0.6.0, carb absorption now uses a `/\` shaped bilinear carb absorption model (replacing the previous `---` shaped constant model for predicting the future absorption of newly entered carbs that haven't shown any effect on BG yet.  As a result, the COB predBG purple prediction line immediately after entering carbs shows an S-curve shaped rise that starts out flat (in line with current BG trends), and then rises sharply after about an hour before flattening out.  

* A typical meal absorption time of 3h is assumed when carbs are first entered, which is then extended over time, so that oref0 gradually relies more on actual observed carb absorption as carbs are absorbed. When the carbs are first entered, remainingCATime is set to 3 hours. When 50% of carbs have absorbed, the remainder (that aren't seen to be absorbing already) are predicted to take another 4.5h. And as COB approaches zero, remainingCATime will approach 6 hours.
* As in 0.5.3, any actual observed carb absorption will be predicted to continue into the future.  The predCI assumed carb absorption described above only kicks in for carbs would not be absorbed within the current carb absorption time if the current carb absorption rate decreases linearly over that time.

### **Exponential Curves**
The code now supports three new configuration variables in preferences.json to allow exponential curves that better reflect the effects of insulin:

- curve, with legal values of blank, "bilinear", "rapid-acting" and "ultra-rapid"

1. if left blank or set to "bilinear", OpenAPS uses the old insulin activity model

2. If set to "ultra-rapid", the new curve model analyzed from publicly available data for Fiasp from Novo Nordisk is used, set to reach activity peak at 55 minutes
[...]

PR #690, opened 2017-10-03
DO NOT USE enableSMB_always or enableSMB_after_carbs with xDrip+, Libre, or similar

scottleibrand:
This PR adds a warning to lib/profile.index.js (the only place enableSMB_always is currently defined that people using DIY CGM receivers like xDrip+, LimiTTer, etc. should not enable enableSMB_always (or enableSMB_after_carbs), as there is a small risk of serious hypoglycemia if the sensor generates a long series of false high readings that do not come down normally as BG comes down.  The official Dexcom receivers and apps properly filter out such readings with a ??? when they're accompanied by a high noise level (as they almost always are).  Until such behavior is added to the DIY receiver options, they should not be trusted to trigger SMBs overnight.  The enableSMB_with_COB and similar options are safer with potentially erroneous CGM data, as they're self-limiting: high BGs will decay COB more quickly and shut off SMBs.  And of course enableSMB_with_temptarget can be used with temporary targets set to expire at bedtime.


scottleibrand:
If we agree that these configurations should not be used with DIY CGM receivers, we should update the https://github.com/openaps/oref0/pull/650 PR notes as well, and make sure these warnings show up in any documentation we add for these new flags.

sulkaharo:
Is the ??? behavior documented somewhere? My experience is Dexcom goes to ??? mode when there’s significant noise from the sensor, not when it’s been high for a long period.

scottleibrand:
The ??? behavior is indeed for significant sensor noise.  I haven't seen any reports where the Dexcom receiver displays a high false reading for a significant amount of time, except in the cases of acetaminophen use.  However, there was a recent report (https://gitter.im/nightscout/intend-to-bolus?at=59d2bbb1b59d55b82365bed8) of xDrip+ reporting falsely high (>300 mg/dL) values for at least 15m with high noise, which would've been filtered by Dexcom's algorithms.  @drnoname82 has discussed this with @jamorham and was going to open an issue about adding this safety check to xDrip+, but I don't know if that has been opened yet.

scottleibrand:
And it looks like it has now been opened: https://github.com/NightscoutFoundation/xDrip/issues/191

tim2000s:
xDrip+ does generate a "High Noise" warning when it sees jumps to high levels and higher noise, I'm just not sure whether that gets into the xDrip+ data that OpenAPS receives.

scottleibrand:
Per https://github.com/NightscoutFoundation/xDrip/issues/191 I think that's the problem: we just need that noise information to be communicated in the glucose.json file, ideally in a way similar to what Dexcom does.  (But if they choose to communicate it some other way, we should be able to adjust oref0 to handle that as well.)

drnoname82:
Exactly. As far as I understand, there is currently no way to judge the quality of BG data from the MongoDB entries XDrip+ generates. However, the required information (noise level) is available locally within the XDrip app. So in my opinion XDrip should be changed to either a) upload high-noise readings as raw instead of SGV and provide the noise level or b) only upload low-noise readings

sulkaharo:
Does someone have the appropriate json field names and value ranges for the noise information?

sulkaharo:
Did AAPS also pull G5 info direct from the transmitter? I recall there’s no Dex receiver involved

tim2000s:
It uses xdrip too.

tim2000s:
This is the data:

[                                                  
  {                                                
    "direction": "Flat",                           
    "noise": 1,                                    
    "dateString": "2017-10-04T06:36:58.286000+01:00
",                                                 
    "sgv": 131,                                    
    "device": "xDrip-DexcomG5",                    
    "unfiltered": 138016,                          
    "filtered": 139904,                            
    "date": 1507095418286,                         
    "rssi": 100,                                   
    "_id": "59d4737ab3fa33c832373efb",             
    "type": "sgv",                                 
    "glucose": 131

tim2000s:
@sulkaharo The issue lies in what xDrip uploads to NightScout. All values have a noise value of 1.

sulkaharo:
@tim2000s Sure - can @scottleibrand or someone else using an official receiver confirm what the expected values are? Nightscout implementation based on earlier data from the G4 receiver is:

```
  rawbg.noiseCodeToDisplay = function noiseCodeToDisplay(mgdl, noise) {
    var display;
    switch (parseInt(noise)) {
      case 0: display = '---'; break;
      case 1: display = translate('Clean'); break;
      case 2: display = translate('Light'); break;
      case 3: display = translate('Medium'); break;
      case 4: display = translate('Heavy'); break;
      default:
        if (mgdl < 40) {
          display = translate('Heavy');
        } else {
          display = '~~~';
        }
        break;
    }
    return display;
  };
```

sulkaharo:
Also - what is the definition of noise we're after and how does that definition map to above noise codes?

jmcrawford:
This is the packet definition from xDrip.c:
`typedef struct _Dexcom_packet
{
	uint8	len;
	uint32	dest_addr;
	uint32	src_addr;
	uint8	port;
	uint8	device_info;
	uint8	txId;
	uint16	raw;
	uint16	filtered;
	uint8	battery;
	uint8	unknown;
	uint8	checksum;
	int8	RSSI;
	uint8	LQI;
} Dexcom_packet;`

Could noise be interpolated from the difference between raw and filtered values?

jamorham:
I don't think noise and glucose values should be limited to Dexcom type data only.

Regarding the issue raised:

https://github.com/NightscoutFoundation/xDrip/issues/191

I am hopeful that this commit will go some way to resolving it:

https://github.com/NightscoutFoundation/xDrip/commit/b4b632fa993ccda71e45d2c2dc3b2794560275c5 

This change is available in the Oct 4th xDrip Nightly.

As far as looping is concerned, it isn't my area of expertise but I would say that with these changes, any reading with a noise value other than `1 / Clean` should be disregarded.

Not all data sources have a true "raw" value so I would prefer not to try to exclude sgv and provide just raw as a fallback as I think this could be misleading.

jamorham:
Anyone had a chance to verify if this is working correctly? Thanks.

drnoname82:
Unfortunately not because I could not get the latest nightly to work with my SW3 and had to go back to the latest beta. The few readings it reported all had low noise and so I could not test any further.

jamorham:
Ok, try the Oct 6th nightly as I think this should resolve the SW3 issues.

ceben80:
I do have enabled `enable enableSMB_always`, running xDrip+ last nightly with G4, but the rig is pulling the BGs directly from my phone (cgm=xdrip), no issues so far.

tim2000s:
I can confirm variable noise reports in the xdrip uploads. Most of the time I see nois:=1 but when getting ⬆️⬆️ On xdrip it uploads with noise:=2.

tepidjuice:
Wouldn't 'enableUAM' be effected similarly?

scottleibrand:
No, UAM is much safer if SMBs are disabled: it just calculates insulinReq differently based on recent deviations, but won't be able to deliver insulin very fast via temp basals.

sulkaharo:
So, instead of just warning, I propose changing determine_basal lines 63 to 66 to

```
var bg = glucose_status.glucose;
var noise = glucose_status.noise;
if (bg < 39 || noise >= 3) {  //Dexcom is in ??? mode or calibrating
rT.reason = "CGM is calibrating, in ??? state or noise is high";    
```

which should make the behavior between xDrip and Dexcom fairly consistent (assuming my read on the glucose object being used here is correct).

tim2000s:
Based on the dying sensor I experienced with xdrip, this looks like it makes sense.

sulkaharo:
https://github.com/openaps/oref0/pull/735

scottleibrand:
Given that #735 is somewhat stalled, should we go ahead and merge this PR before 0.6.0 release?

drnoname82:
I think this PR is worth merging, also because even after we get #735 to work, most XDrip users still use XDrip versions that do not provide noise data.

Do you have any idea why #735 might not work? It would be great to have that safety feature in 0.6.0 as well.

tim2000s:
I’ve been running SMB Always On for the last couple of weeks with xdrip and sensor placement has resulted in a fair amount of compression lows. The upshot of this is that on the rebound I’ve seen SMBing of 0.1u before the levels stabiliise. This doesn’t appear to be anymore insulin than the high temp would have provided without it, especially given the suspend that’s taken place every time beforehand. 

There’s been no issue with the drops and any moodiness on the way down as suspension is fine. 

What I’ve yet to see with xdrip is a long series of false highs. Given the data is calculated and provided from the G5 transmitter (where the algorithm resides) this is perhaps unsurprising.

scottleibrand:
Agreed, this should be less of an issue when using the xDrip with the G5 transmitter.  I don't think we need to change the warning, though, as anyone using DIY systems to read their CGM needs to really understand what they're doing before enabling SMB.

PR #696, opened 2017-10-06
Time-aware carbsReq

scottleibrand:
The current carbsReq notifications tend to be too sensitive when carbs aren't likely to be needed for an hour or more.  This changes the carbsReq calculations to be based on how much time BG is predicted to remain above the minGuardBG threshold (which comes out to 70 for a target of 100).  It uses naive_eventualBG so as not to be too sensitive to current BG slope, and attempts to make sure that there will be enough time for zero-temping to get IOB down to zero by the time BG hits the threshold.  If there is not, it will recommend enough carbs to compensate, and also note how soon they'll likely to be needed, so the user can decide whether to do carbs right away or wait and see if BG flattens out on its own and they're not actually needed.

This also disables insulinReq pushover notifications by default now that eSMB can usually handle insulin dosing unassisted, so that only situations potentially requiring action generate notifications.

- [ ] TODO: add an option to oref0-setup to enable insulinReq pushover notifications as well.

### DIFF
--- a/bin/oref0-pushover.sh
+++ b/bin/oref0-pushover.sh
@@ -5,11 +5,12 @@ USER=$2
 FILE=${3-enact/smb-suggested.json}
 SOUND=${4-none}
 SNOOZE=${5-15}
+ONLYFOR=${6-carbs}
 
 #echo "Running: $0 $TOKEN $USER $FILE $SOUND $SNOOZE"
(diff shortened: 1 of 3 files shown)


scottleibrand:
This is working well: I think it's good to merge.

PR #699, opened 2017-10-06
Attempting to --nogit by default for existing user installs

danamlewis:
Thanks to the amazing @Kdisimone as always for the first round of testing, confirmed it successfully installs and removes an existing git directory.

Looks like this after starting setup:
screen shot 2017-10-06 at 2 25 36 pm

And then it successfully removed .git

For anyone else who wants to test:
1. `cd ~/myopenaps` and `ls -la` (you should see .git listed in the alphabetical list - we need this tested on rigs that still have git)
2. `cd ~/src/oref0/ && git fetch && git checkout no-git-default && git pull`
3. `npm run global-install`
4. `bash ~/myopenaps/oref0-runagain.sh`
5. Once done, `ls -la` again in `myopenaps` and confirm that .git no longer exists there.
*(If you're planning to stay on this branch vs. switching back to dev, make sure to re-set your preferences!)*

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -499,6 +499,11 @@ echocolor-n "Continue? y/[N] "
 read -r
 if [[ $REPLY =~ ^[Yy]$ ]]; then
 
+    # Attempting to remove git to make install --nogit by default for existing users
+    echo Removing any existing git
+    rm -rf ~/myopenaps/.git
+    echo Removed any existing git
+
     # TODO: delete this after openaps 0.2.1 release
     echo Checking openaps 0.2.1 installation with --nogit support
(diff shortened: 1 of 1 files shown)


scottleibrand:
Nice.  We should also test it on a rig that's already had .git removed.  It will probably whine about being unable to remove .git - we'll need to decide if that's clear enough, or if we need to redirect the output of `rm` to squelch it.

Kdisimone:
@scottleibrand I'll rerun it now...want me to just try a runagain?

scottleibrand:
Ok, good.  No scary warnings.  :-)  Thanks.

Kdisimone:
no worries.  installed fine both ways.  and YEAH for nogit!

PR #702, opened 2017-10-07
Pumphistory safety checks

scottleibrand:
@hilarykoch recently experienced an A52 error that rebooted the pump and somehow caused decocare/openaps to prematurely stop parsing pumphistory such that it didn't return an error but didn't include recent events.  This in turn resulted in incorrect IOB, causing OpenAPS to overdose insulin, providing Leo with an unscheduled pump break and dessert.

This PR adds some safety checks to make sure that pumphistory is being updated, and that the lastTemp in pumphistory matches the currently running temp basal on the pump.

### DIFF
--- a/bin/oref0-determine-basal.js
+++ b/bin/oref0-determine-basal.js
@@ -219,7 +219,7 @@ if (!module.parent) {
     }
 
     console.error(JSON.stringify(glucose_status));
-    console.error(JSON.stringify(currenttemp));
+    //console.error(JSON.stringify(currenttemp));
     //console.error(JSON.stringify(profile));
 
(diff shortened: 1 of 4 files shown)


scottleibrand:
This is working now: I think it's good to merge.

mgranberryto:
I've had this happen as well.

scottleibrand:
@mgranberryto can you describe exactly what happened in your situation?  Did it just fail to read some records, or were new temps and SMBs also not being read?

mgranberryto:
New temps/SMBs were being unread.  I assumed that it was due to an edited record after the fact, but reading this after the fact I realize that this is the same issue.  It recovered about 6 hours later.

PR #705, opened 2017-10-08
Use autosens basal adjustments in IOB calculations

scottleibrand:
Up until now, we've used autosens to adjust the neutral basal assumed by determine-basal and used to calculate temps, and to adjust ISF and BG targets, but have not used autosens to change the basals used when calculating IOB.  Now that we're doing so for exercise mode, it makes sense to also do so normally based on autosens data.  This makes autosens much more effective at counteracting increased sensitivity or resistance.  In light of that, it also means that we no longer need to adjust targets for autosens by default.  We can also go back to keeping carb ratio independent of basal sensitivity, as we've noticed that basal sensitivity and meal sensitivity don't always correlate.


tepidjuice:
I'm not sure if this comment should go here or in #684. But if IOB is affected by the autosens ratio might not the large discontinuous changes in IOB cause determine-basal to overreact?

For example. Suppose a high temp-target is implemented for a long duration to cover exercise. After the end of the exercise the temp-target finishes at which point the autosens ratio returns to 1.0, or whatever its original value was. Now determine-basal sees a large negative IOB and tries to counteract that by administering more insulin which in this instance is probably not required as sensitivity is higher post-exercise.

scottleibrand:
Yes, if you let the exercise mode end while the increased sensitivity is ongoing, oref0 will start undoing some of the work it did to get IOB down.  If that is inappropriate, you can run the exercise mode temp target longer, or set another one not quite so high for the post-exercise period while autosens catches up. This PR will at least allow autosens to do the same thing exercise mode was doing once it detects the increased sensitivity, so you don't have to run the post-exercise mode temp target as long.  We'll have to test and see whether people see this as a problem and want a more automatic solution, or if those tools are sufficient.

scottleibrand:
@tepidjuice that last commit (https://github.com/openaps/oref0/pull/705/commits/c4e78c5949d62d080b5e7249f6594e8b8f88ac38) should partially address your concern around post-exercise sensitivity, as it will cause autosens to "inject" more negative deviations for time periods where you're running a high temp target, so that it will subsequently detect more sensitivity than by the basal deviations alone

scottleibrand:
To test: `cd ~/src/oref0 && git fetch && git checkout autosens-iob && git pull && npm run global-install`

scottleibrand:

drnoname82:
We also tested this branch without any problems.

PR #714, opened 2017-10-15
predBGs with zero temps

scottleibrand:
This adds a new purple prediction line to NS (which will require an update to NS to view: https://github.com/nightscout/cgm-remote-monitor/pull/2919) showing how long it will take BG to level off at/above target if deviations suddenly cease and we run a zero temp until then.  It then uses the lowest of those ZTPredBGs to ensure that we're dosing safely with UAM: if UAM predicts BG to stay high (based on deviations), but we wouldn't be able to prevent a low if those deviations suddenly stopped, we'll adjust the UAM-based insulinReq to be somewhat less aggressive.  Conversely, if UAM shows BG ending up below target in 4h, but the ZTPredBGs show that we can safely zero temp and level out well above target, it will allow oref0 to be slightly more aggressive and bring BG down faster, safe in the knowledge it can zero temp if needed.

This also removes the aCOB predBGs, which were only displayed in NS and not used for dosing in any way.  If anyone finds the new prediction lines insufficient without that and wants it back, I'd be happy to merge a PR to add a preference to re-enable them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ coverage/
 
 npm-debug.log
 bin/__pycache__
+
+package-lock.json
--- a/bin/ns-get.sh
+++ b/bin/ns-get.sh
@@ -9,7 +9,7 @@ QUERY=${3}
 OUTPUT=${4-/dev/fd/1}
 
 CURL_FLAGS="--compressed -g -s"
-NIGHTSCOUT_FORMAT=${NIGHTSCOUT_FORMAT-json}
+NIGHTSCOUT_FORMAT=${NIGHTSCOUT_FORMAT-jq .}
 test "$NIGHTSCOUT_DEBUG" = "1" && CURL_FLAGS="${CURL_FLAGS} -iv"
 test "$NIGHTSCOUT_DEBUG" = "1" && set -x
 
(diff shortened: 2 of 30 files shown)


drnoname82:
This change just did a brilliant job in bringing down a high. I wonder if advanced target adjustment is still needed since this is a much more elegant and precise way of dosing insulin. I will test the changes in Nightscout as soon as they are merged to dev.
PS: We do not need the aCOB prediction line

drnoname82:
Now also tested the Nightscout ZT branch, I especially like the prediction tooltips. The font size of the ".x" SMB captions is a tad small on an iPhone 6s usingthe Nightscout app, but fine on desktop computers.

scottleibrand:
We made a few more tweaks in the NS `rendering` branch to make SMB labels more readable (removing the leading 0 and the U), so I just bumped the font size back up a bit there.  Want to give that branch a try and comment on https://github.com/nightscout/cgm-remote-monitor/pull/2926 ?

scottleibrand:

scottleibrand:
It appear that @dmdfreak had an instance this morning of SMB dosing too much insulin due to high UAMpredBGs, which this would have helped prevent.  So I'm inclined to merge this sooner than later if no objections.

danamlewis:
+1 for merging in. have been happy w/ this behavior. -1 for ZTnaming stuff, but can figure that out in dev.

mgranberryto:
I've been running it for about 24h and have been reasonably happy too.

PieterGit:
Please test this PR with https://github.com/nightscout/cgm-remote-monitor/pull/2991 (Nightscout PR for showin this predication).

PR #716, opened 2017-10-15
Offline local-only wifi hotspot

scottleibrand:
This causes oref0-online to set up a local-only wifi hotspot if it is unable to get online via wifi or Bluetooth tethering.  This allows phones (including iPhones, which can't do Bluetooth tethering unless they have Internet) to connect to the rig using the local wifi, in order to ssh in (for now) or (in the future) potentially use a locally hosted web interface to interact with the rig.  When the rig is once again able to connect to the Internet using Bluetooth tethering, it tears down the local-only hotspot and goes back to trying to connect to the Internet via wifi.  As before, once it's able to connect to the Internet via wifi it tears down the Bluetooth connection until it's needed again.


Bender1061:
While this is a real neat idea. 
what about on the airplane during takeoff and landing? while yes most of us are bad passengers and keep things on anyways. But we are more than just bending the rules by having something like this automatically start up. And we would not want to just turn off our rig, because sometimes getting the rig to get the pump time does not work. (at least I had major issues with that back in June, ended up having to drive into down to get my phone hotspot working) 

but beside that issue.

Are we setting a password for that wifi hotspot? In my experience. as soon as you start a wifi hotspot, the number of people that try to hookup to it (no matter what the name is) is very large.
Try it sometimes with a network scanner running. 
Resourceful people might try to see what services that device is running (yes I'm guilty of that) and when someone is sitting there board on an airplane for example maybe they try playing with things... 

So I did not see it, but it was a quick read through, we need to enable a hotspot password.

scottleibrand:
Please do submit a PR for WPA2 support. 

In flight wifi stays on the whole flight. You're not going to interfere with flight avionics by having a wifi hotspot running. (Your cellular radio might if you're in first class, but so far there have been no confirmed reports of even that.)

Bender1061:
You can have your electronics on at all times but inflight wifi does not turn on till you get to 10K feet, and when landing turns off when you drop down to 10K feet. 

Yes I know there has been nothing confirmed about wifi or any consumer radio signals doing anything to an airplanes avionics. 

but that is still regulated by the FAA and they require that phone to be in Airplane mode once the flight door closes and the plane is below 10K feet.

I do want to have this ability, but by automatically enabling it to turn on, I see a problem here, basically turning anyone that flies and is trying to be a good passenger and turn off their phones into a FAA Criminal. 

Most of us road warriors do tend to ignore the rules anyways. But I think there at least needs to be another step involved on enabling this feature. maybe in a different Preference file or something that someone would have to manually enable in order for this feature to be used.

scottleibrand:
I think a preference for offline mode will be needed regardless.  One much more practical reason: if you run oref0-setup with a BT MAC, but don't get the pairing to work properly, you won't be able to easily get it back out of local-only wifi mode and into client mode.

I'll check on my next flight, but I think the gogoinflight SSID remains active at all times, and just doesn't let you connect to the ground until you get to 10k ft.  And every flight I'm on has a handful of other SSIDs active at all times: the FAA isn't about to start walking the plane aisles with 2.4GHz antennas, so I'm not worried about providing an easy way to turn it off and on mid-flight.

JELCRAWFORD:
So if you put your phone in airplane mode will this change actually turn airplane mode off on your phone to create a local network?  (We leave a local network on throughout the flight including takeoff and landing,  I guess making us non compliant).

scottleibrand:
If you use the `offline` branch, then when you go into airplane mode on your phone, your rig will spin up an `OpenAPS` local-only wifi network.  If you turn wifi back on on your phone, you can then connect to that local network and manage your rig.

JELCRAWFORD:
So if the phone stays in airplane mode and  that is the expectation of the FAA, then phone in airplane mode = compliance, no?

scottleibrand:
They theoretically also want you to put your medical devices into airplane mode.  In actual practice, they don't care about any of this.

tim2000s:
Have to agree with Scott on this. There are a few studies you can dig out that demonstrate that the signals from passenger devices have no effect on the avionics on a plane (and there have been no incidents as a result of passenger devices remaining on during take off, landing and inflight).

Bender1061:
It does not matter if we believe that our Wifi hotspot will interfere with the plans avionics or not. the Law is to follow all crew instructions and one of those instructions is to not have a radio device running after the doors close and below 10K feet.

I agree it does not do anything and I think it's an archaic rule.  

but while we are a small group (in the terms of amount of rigs in the air at the same time) I think its a better idea not to have a device that automatically goes against the FAA, straight out of the box. 

we have the FDA watching us already, I would hate another government organization having an issues with what we are doing. 

all I'm asking for is this feature not be On already after you've ran oref0-setup.
if we used something like
"enable_wifi_hotspot": false   as the default.
then if someone understands that they are potentially violating FAA Rules and Regulations (a Federal offence) then they can turn it on. 
Really not a big deal that would make me sleep better at night. Then the owner of responsibility is on the user, not the OpenAPS development team. 

Please don't get me wrong, I think this feature is a wonderful idea and Fully plan on using it. I'm actually kicking myself in the butt for not having thought of it. I own a Piratebox which can use WPA2 and has a small webserver running on it. So we can look to see how that works and maybe borrow some of the tools from it.

webserver-wise I see large potential. Such as being able to set a time or change your timezone while camping and for some unknown reason not always grabbing it automatically from the pump. (yes an issue that needs to be dug into) 
anyways I'm excited about the possibilities.  just worry about the on by default part.

scottleibrand:
Then I think we're in agreement to have this off by default, with an option to enable either in oref0-setup or in preferences.

jaylagorio:
Can there be a ``preferences.json`` line that allows you to configure the hotspot name? I'd hate for two of us to step on each other with the same hotspot name if it can be prevented.

thebookins:
I'm having a little trouble using this for offline looping using xdrip. [This line](https://github.com/openaps/oref0/pull/716/files#diff-f51a693ec933210db5d42ac8fa23a0a2R80) appears to be redundant and causes the local hotspot to stop and restart when we get a local IP address over bnep0 (when `check_ip` can still return `false`). https://github.com/openaps/oref0/pull/727 is an attempt to prevent this. Even with these changes, it seems that when there is no wifi and no cellular service, the BT PAN connects once every minute and then disconnects when no internet is available on bnep0. Consequently the endpoint for xdrip switches between http://10.29.29.1/api/v1/entries and http://172.20.10.x/api/v1/entries (on iPhone) and readings will be missed if they arrive at the wrong moment.

This looks to be a great feature for offline visualisation and works really well on Airplane mode (when BT tethering isn't possible). It would be great if it can work seamlessly for a kid to provide a local endpoint for glucose values when NS isn't available.

scottleibrand:
@thebookins I can't think of any reason we shouldn't leave the bluetooth connection up even after launching the wifi-only hotspot.  then you'll have connectivity to the rig over both interfaces, so either IP should work.  the bluetooth will get momentarily disconnected once a minute, each time it does bt_connect and tries to get online, but that shouldn't disrupt things much.

scottleibrand:
@jaylagorio looks like it'll be easier to just have people edit /etc/hostapd/hostapd.conf directly if they have multiple rigs and want different SSIDs.

scottleibrand:
@Bender1061 `offline_hotspot: false` is now the default in preferences.json.

scottleibrand:
Do we think it would be useful to set a well-know WPA2 password in /etc/hostapd/hostapd.conf to ward off casual connections ? Or would it be better to avoid security by obscurity and leave the hotspot open unless someone wants to edit hostapd.conf to add their own passphrase?

jaylagorio:
About changing the default hostname manully with ``hostapd.conf`` - that's cool. But I think we should set up a default WPA2 password to ward off casual connections. It's a low bar but anything's better than nothing.

Bender1061:
So Trying to get this to work.

I see the wifi hotspot, but when I try to connect it does not assign an IP address to either my iphone or my Android. 

I found where the hotspot is being started and I ran them through manually, this is what I'm getting

```
root@Wall-E:~#
root@Wall-E:~# wpa_cli terminate
Failed to connect to non-global ctrl_ifname: (null)  error: No such file or directory
root@Wall-E:~# ifdown wlan0
root@Wall-E:~# cp /etc/network/interfaces.ap /etc/network/interfaces
root@Wall-E:~# ifup wlan0
root@Wall-E:~# /etc/init.d/hostapd start
Starting hostapd (via systemctl): hostapd.service.
root@Wall-E:~# service udhcpd stop
root@Wall-E:~# /etc/init.d/dnsmasq start
Starting dnsmasq (via systemctl): dnsmasq.service.
root@Wall-E:~# systemctl daemon-reload
[ 1335.806706] systemd-sysv-generator[6979]: Ignoring creation of an alias galileod.service for itself
root@Wall-E:~#
```

I'm wondering if maybe I'm doing my wifi setup a bit different than what people are doing now, as I've not made any changes to how I've been setting up my wifi since I first got an Edison June of 2016. And I know there are some scripts people are running. 
but I won't be able to test that till this weekend.

Bender1061:
So my issue right now. It works. when the did the suggestion from above. 

but I can't get it to stop.

I can tell it's because the hotspot changes the interfaces file. in order to allow the hotspot to activate, but I'm believing its due to if can't run function stop_hotspot I'm thinking its because iwgetid -r wlan0 does not return anything.

so since the call to wpa_supplicant.conf is commented out, there is just no way to get your wifi to connect anymore.

Bender1061:
ok scratch the above. I see that it's supposed to get getting the bt network connected first. So I'm troubleshooting that now.

Bender1061:
Ok so oref0-bluetoothup does not actually do anything because the command hciconfig does not return anything.

Bender1061:
So Right now, I'm going to warn people do not try this if you don't have an extra rig.

i have 2 rigs that will no longer connect to my wifi at home or my phone. Something about this is breaking the system to not allow the wpa_supplicant.conf to run. even running commands does not do anything. I've tried repairing the connections and manually changing the files, but does not work.

Bender1061:
Ok, so this is what I've learned.

Now the Rig does not connect to WiFi automatically at bootup.

I don't know what would have changed that. 

so if you don't have a Bluetooth hotspot on, it automatically starts the hotspot. 

even a cycle power does not stop that. 
you must have the bluetooth hotspot on your phone running in order for the WiFi to start working.

this is very strange behavior. I don't understand why the normal /etc/network/interfaces does not start at bootup anymore. 

Now when we start the local hotspot, we do change the /etc/network/interfaces. So if you were a hotspot when you cycled power, that /etc/network/interfaces would still be in play and would explain in that situation for /etc/wpa_supplicant/wpa_supplicant.conf not starting up.

but I have even disabled oref0-online and make sure that the /etc/network/interfaces is correct for a cycle of power and it still does not come up connected to WiFi like normal. 

I also have tried cycling power when it is connected to a WiFi connection. So something must have been changed in the startup services or whatever is run to start services in Linux.

bin/oref0-setup.sh
```diff
@@ -650,6 +650,16 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
...
             echo bluez v ${bluetoothdversion} already installed
         fi
+        echo Installing prerequisites and configs for local-only hotspot
+        apt-get install -y hostapd dnsmasq || die "Couldn't install hostapd dnsmasq"
+        ls /etc/dnsmasq.conf.bak 2>/dev/null || mv /etc/dnsmasq.conf /etc/dnsmasq.conf.bak
+        cp $HOME/src/oref0/headless/dnsmasq.conf /etc/dnsmasq.conf || die "Couldn't copy dnsmasq.conf"
+        ls /etc/hostapd/hostapd.conf.bak 2>/dev/null || mv /etc/hostapd/hostapd.conf /etc/hostapd/hostapd.conf.bak
+        cp $HOME/src/oref0/headless/hostapd.conf /etc/hostapd/hostapd.conf || die "Couldn't copy hostapd.conf"
+        sed -i.bak -e "s|DAEMON_CONF=$|DAEMON_CONF=/etc/hostapd/hostapd.conf|g" /etc/init.d/hostapd
+        cp $HOME/src/oref0/headless/interfaces.ap /etc/network/ || die "Couldn't copy interfaces.ap"
```
Bender1061: is this supposed to say cp $HOME/src/oref0/headless/interfaces.ap /etc/network/interfaces.ap  ?

scottleibrand: It's the same thing.  Repeating the filename is redundant.

Bender1061:
So this morning, I started with a completely Fresh rig.

Reinstalled Jubilinux0.1.1.
-Installed OpenAPS via the manually entered in script from manuals.
-did not run master oref0-setup.
-did a cd ~/src/ && rm -r oref0
-did a $ mkdir -p ~/src; cd ~/src && git clone -b dev git://github.com/openaps/oref0.git || (cd oref0 && git checkout dev && git pull)
-Ran cd && ~/src/oref0/bin/oref0-setup.sh --dir=johnopenaps --serial=622890 --tty=/dev/spidev5.1 --max_iob=6 --ns-host=https://myaps.herokuapp.com --api-secret=*********** --cgm=MDT --btmac=AA:AA:AA:AA:AA:AA --btpeb=AA:AA:AA:AA:AA:AA --pushover_token=***************** --pushover_user=****************** --enable='autosens autotune meal microbolus'
-this installed Dev Branch.
-Once Dev Branch running I connected my phone on bluetooth, made sure the Rig would connect to multiple wifi networks and would connect to bt pan. Rebooted, cycled power. everything came up perfectly. Exactly as it should be performing.
-did a $ crontab -r and then did a cd ~/src/ && rm -r oref0
-did a $ mkdir -p ~/src; cd ~/src && git clone -b offline git://github.com/openaps/oref0.git || (cd oref0 && git checkout offline && git pull)
-did a $ cd && ~/src/oref0/bin/oref0-setup.sh --dir=johnopenaps --serial=622890 --tty=/dev/spidev5.1 --max_iob=6 --ns-host=https://bender1061.herokuapp.com --api-secret=042607258313 --cgm=MDT --btmac=24:F0:94:19:D2:0D --btpeb=00:18:33:75:0F:66 --pushover_token=a2ugephuya73w83ig8oax63jhxomzj --pushover_user=ucbh5ahbw64riiszc9np7co62amw7i --enable='autosens autotune meal microbolus'
- This started running, rig was hooked up to wifi. everything seamed happy. 
- did a reboot.
- rig did not automatically have wifi when I rebooted, instead bt-pan started.
```
Starting oref0-online.
At Sun Oct 22 15:13:01 EDT 2017 my local IP is: 172.20.10.4

At Sun Oct 22 15:13:02 EDT 2017, my wifi network name is , and my public IP is: 166.137.12.120
Finished oref0-online.
^Z
[1]+  Stopped                 tail -n 100 -F /var/log/openaps/network.log
root@Gort:~# ipconfig
-bash: ipconfig: command not found
root@Gort:~# ifconfig
bnep0     Link encap:Ethernet  HWaddr 98:4f:ee:04:5e:94
          inet addr:172.20.10.4  Bcast:172.20.10.15  Mask:255.255.255.240
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:525 errors:0 dropped:0 overruns:0 frame:0
          TX packets:572 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:182296 (178.0 KiB)  TX bytes:71196 (69.5 KiB)

lo        Link encap:Local Loopback
          inet addr:127.0.0.1  Mask:255.0.0.0
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:8 errors:0 dropped:0 overruns:0 frame:0
          TX packets:8 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0
          RX bytes:1104 (1.0 KiB)  TX bytes:1104 (1.0 KiB)

usb0      Link encap:Ethernet  HWaddr 02:00:86:bc:ea:c1
          inet addr:10.11.12.13  Bcast:10.11.12.255  Mask:255.255.255.0
          UP BROADCAST MULTICAST  MTU:1500  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

wlan0     Link encap:Ethernet  HWaddr 78:4b:87:a4:e5:0f
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:84 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:28728 (28.0 KiB)
```

- Decided to see if I could get the wifi to force start. 
```
root@Gort:~# ifdown wlan0; ifup wlan0
Removed stale PID file
Internet Systems Consortium DHCP Client 4.3.1
Copyright 2004-2014 Internet Systems Consortium.
All rights reserved.
For info, please visit https://www.isc.org/software/dhcp/

Listening on LPF/wlan0/78:4b:87:a4:e5:0f
Sending on   LPF/wlan0/78:4b:87:a4:e5:0f
Sending on   Socket/fallback
DHCPRELEASE on wlan0 to 192.168.1.1 port 67
[  800.051410] wl_android_wifi_off in
[  800.058966] dhd_wlfc_deinit():3248, ampdu_hostreorder get failed Err = -23
[  800.066347] dhd_wlfc_deinit():3268 successfully disabled bdcv2 tlv signaling, 0
[  800.074514] wifi_platform_set_power = 0
[  801.353914]
[  801.353914] Dongle Host Driver, version 1.141.59 (r)
[  801.353914] Compiled in /home/robin/src/edison-linux-helper/edison-bcm43340 on Apr  4 2017 at 22:42:52
[  801.354090] wl_android_wifi_on in
[  801.374049] wifi_platform_set_power = 1
[  801.943684] F1 signature OK, socitype:0x1 chip:0xa94c rev:0x2 pkg:0x0
[  801.952253] DHD: dongle ram size is set to 524288(orig 524288) at 0x0
[  801.962006] dhdsdio_download_firmware: firmware path=/lib/firmware/fw_bcmdhd.bin, nvram path=/lib/firmware/bcmdhd.cal
[  802.092169] sdioh_request_buffer: [1] doing memory copy buf=f58d2000, len=2037
[  802.103889] dhdsdio_write_vars: Download, Upload and compare of NVRAM succeeded.
[  802.273131] dhd_bus_init: enable 0x06, ready 0x06 (waited 0us)
[  802.282858] wifi_platform_get_mac_addr
[  802.286770] wifi_get_mac_addr_intel: unable to open /config/wifi/mac.txt
[  802.298437] Firmware up: op_mode=0x0002, MAC=78:4b:87:a4:e5:0f
[  802.319797] Firmware version = wl0: Aug 15 2014 15:09:17 version 6.20.190.2 (r497061) FWID 01-bb9bbd07
[  802.330637] dhd_preinit_ioctls wl ampdu_hostreorder failed -23
[  802.337605] dhd_wlfc_init(): successfully enabled bdcv2 tlv signaling, 79
[  802.344929] dhd_wlfc_init(): wlfc_mode=0x0, ret=-23
Internet Systems Consortium DHCP Client 4.3.1
Copyright 2004-2014 Internet Systems Consortium.
All rights reserved.
For info, please visit https://www.isc.org/software/dhcp/

Listening on LPF/wlan0/78:4b:87:a4:e5:0f
Sending on   LPF/wlan0/78:4b:87:a4:e5:0f
Sending on   Socket/fallback
DHCPDISCOVER on wlan0 to 255.255.255.255 port 67 interval 7
DHCPDISCOVER on wlan0 to 255.255.255.255 port 67 interval 11
DHCPDISCOVER on wlan0 to 255.255.255.255 port 67 interval 3
No DHCPOFFERS received.
No working leases in persistent database - sleeping.
root@Gort:~# ifconfig
bnep0     Link encap:Ethernet  HWaddr 98:4f:ee:04:5e:94
          inet addr:172.20.10.4  Bcast:172.20.10.15  Mask:255.255.255.240
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:618 errors:0 dropped:0 overruns:0 frame:0
          TX packets:676 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:217163 (212.0 KiB)  TX bytes:84749 (82.7 KiB)

lo        Link encap:Local Loopback
          inet addr:127.0.0.1  Mask:255.0.0.0
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:8 errors:0 dropped:0 overruns:0 frame:0
          TX packets:8 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0
          RX bytes:1104 (1.0 KiB)  TX bytes:1104 (1.0 KiB)

usb0      Link encap:Ethernet  HWaddr 02:00:86:bc:ea:c1
          inet addr:10.11.12.13  Bcast:10.11.12.255  Mask:255.255.255.0
          UP BROADCAST MULTICAST  MTU:1500  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

wlan0     Link encap:Ethernet  HWaddr 78:4b:87:a4:e5:0f
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:96 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:32832 (32.0 KiB)

root@Gort:~#
```
- I decided it's kind of looking like it does not like some of the updates that were done with the upgrade.
decided to run $ sudo apt-get update; sudo apt-get upgrade
-gave errors that things were being used. So I commented out crontab
-Waited, but it still was giving me the same error. 
-rebooted. system would not connect to wifi like always since runing offline branch. and now the bluetooth would not connect. First tired oref0-online AA:... and even tried to run oref0-bluetoothup and then sudo bt-pan client 24:... Still not working. 

So once again I have a rig that will no longer connect to the internet, no matter what I try. 
Cause I am unable to run $ sudo apt-get update; sudo apt-get upgrade since not online. 

this is three rigs that have been messed up beyond easy repair, basically I'm going to have to re-flash these edisons.

Bender1061:
Ok after way too much trial and error, I have discovered the issue.
How to fix it, is beyond me at the moment.

we are adding a line to the /etc/init.d 
`DAEMON_CONF=/etc/hostapd/hostapd.conf`

When I start up the Edison after a reboot or cycle power, the wifi does not connect.  Only way I can get it to connect is to first let the oref0-online start running at least once, even if the `"offline_hotspot": false` that has to run at least once, with my bluetooth hotspot on phone turned off. 
once that is run, then I can turn my bluetooth hotspot back on, and the system will then turn on the wpa supplicant and I connect to wifi just fine.

This is extremely reproduce-able and I tried a multiple of ways to see if I could manually get the wifi to start up. (yes I even tried commenting out the oref0-online and see if I could manually do it)

If I comment out the line in /etc/init.d the edison starts up like normal, but of course the hotspot does not work. 

looks to me like we need a way to activate that line only when we want the hotspot to work. 
but we need to make sure it's not put in a place that's going to break the system if we do a reboot.

Bender1061:
Looks like I found the solution

update-rc.d -f hostapd remove
update-rc.d -f dnsmasq remove

will need to be run after we install the hostapd and dnsmasq services in the setup. 

a further issue is what happens when you reboot the rig when it's in the hotspot. you still have an issue of having to hook up with the bluetooth before it does goes to connect to wifi

my solutuion was to add to /etc/rc.local
```
sudo dmesg -n 1
sudo ifdown wlan0
sudo cp /etc/network/interfaces.client /etc/network/interfaces
sudo ifup wlan0
```
the top line is not nessicary, but with all the messages that pop up while running this stuff it sure makes it easier to work on your putty.
But
All of this seems to have fixed the issues!

@scottleibrand I do have a question, is there any reason we can't use the $Hostname for the wifi SSid? That way if two rigs are in the same place they would not cause a problems.

scottleibrand:
Awesome!  Great detective work.  Do you feel up to PR'ing in those changes sometime soon?

I'd be ok with $HOSTNAME as the SSID (and either #OpenAPS or #WeAreNotWaiting as the well-known default PSK), if others like that idea.

Bender1061:
I'll work on a PR tomorrow. And ill test out the PSK stuff

Bender1061:
I think we need a way to check after a set amount of time, if wifi is available. maybe with the crontab? Every 15 mins if we are in hotspot mode, turn off the mode and do a check for wifi, and if not turn the hotspot back on?

scottleibrand:
I added support for disabling the hotspot based on the presence of a /tmp/hotspot_disabled file, with the idea that we would have a web or menu interface option for shutting down the hotspot to switch back to wifi if Bluetooth isn't available (like if you Hootoo/Edimax onto in-flight wifi).

scottleibrand:
@Bender1061 I also added cron jobs to `touch /tmp/disable_hotspot` once an hour, and then `rm` it  a minute later, to allow rigs to eventually reconnect to wifi in the absence of BT.

scottleibrand:
After extensive testing in a challenging-connectivity environment (an otherwise very nice beach house on the Florida panhandle), I was able to merge in @Bender1061's great work, and I think this branch is once again ready for wider testing.

Bender1061:
when crontab is being wrote it actually write $directory oref0-online ...

other than that it looks like it's working quite well. 

how exactly is the .app stuff used? is that actually doing anything right now? or do we need to find something like Lighttpd to run in order to use any of that? and place a simple webpage?

scottleibrand:
Not sure what you mean by "it actually write $directory oref0-online".

The .app stuff is not yet working, so nothing uses it yet.

Bender1061:
after the oref0-setup runs it writes in the crontab `|| cd cd $directory && oref0-online 24:`

Bender1061:
because of that it won't run the hotspot cause it can't read the preferences.json

scottleibrand:
Ah, ok.  I just escaped the $directory with single quotes.  Try now?  Not sure where the double `cd` would be coming from if you're still seeing that...

Bender1061:
yes that fixed that.

Bender1061:
interesting enough the `1 * * * * rm /tmp/disable_hotspot` is not there now.

Bender1061:
@scottleibrand what do you think about adding a `rm /tmp/disable_hotspot` to the rc.local? I know normally that disable_hotspot is not there when we are running, but if by chance you were in that minute when the file is touched there, and you cycled power. you would not have your hotspot turn back on till a minute past the next hour. I know it's small potatoes. but would be easy enough to add to rc.local.

Bender1061:
oh and the syntax error was fixed, it all wrote the crontab correctly now.

scottleibrand:
That's why I wrote it to /tmp - that gets erased on each reboot.  :-)

Bender1061:
so I'm getting times where I'm getting stuck in a loop
```
Starting oref0-online at Wed Nov 1 23:17:02 EDT 2017.
At Wed Nov  1 23:17:02 EDT 2017 my wifi is not connected
At Wed Nov  1 23:17:02 EDT 2017 my local wifi IP is: 10.29.29.1
At Wed Nov  1 23:17:02 EDT 2017 my Bluetooth is connected to John's iPhone
At Wed Nov  1 23:17:02 EDT 2017 my local Bluetooth IP is: unassigned
At Wed Nov  1 23:17:02 EDT 2017 my public IP is: not found

At Wed Nov  1 23:17:02 EDT 2017 my wifi is not connected

No Internet access detected, attempting to connect BT to 24:F0:94:19:D2:0D
Connected to Bluetooth with IP: unassigned

Local hotspot is running.
Finished oref0-online at Wed Nov 1 23:17:28 EDT 2017.
```
I have my wifi hotspot turned back up, and my phone is got it's bluetooth hotspot on, phone says it's connected, but we are not getting a ip address. I've even tried turning off and on the bluetooth hotspot on the phone. Only way I can get out of it is to reboot the edison.
```
root@johnyfive:/tmp# ifconfig
bnep0     Link encap:Ethernet  HWaddr 98:4f:ee:04:2a:79
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:4 errors:0 dropped:0 overruns:0 frame:0
          TX packets:55 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:488 (488.0 B)  TX bytes:17323 (16.9 KiB)

lo        Link encap:Local Loopback
          inet addr:127.0.0.1  Mask:255.0.0.0
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:24 errors:0 dropped:0 overruns:0 frame:0
          TX packets:24 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0
          RX bytes:2343 (2.2 KiB)  TX bytes:2343 (2.2 KiB)

usb0      Link encap:Ethernet  HWaddr 02:00:86:95:f2:3b
          inet addr:10.11.12.13  Bcast:10.11.12.255  Mask:255.255.255.0
          UP BROADCAST MULTICAST  MTU:1500  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

wlan0     Link encap:Ethernet  HWaddr 78:4b:87:a0:ff:fd
          inet addr:10.29.29.1  Bcast:10.29.29.255  Mask:255.255.255.0
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:86 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:27760 (27.1 KiB)

```

scottleibrand:
Does a `dhclient bnep0` do anything?

scottleibrand:
Anything else we want to test before merging this to dev?  If not I think it's good to go.

Bender1061:
lol, yeah I think we are good. Somehow my bluetooth got disabled and would not come back till I cycled power (maybe an edison problem) my other issue with the disable_hotspot being on after reboot was actually (i looked at time stamp, it had just written the file had just turned past the hour, but since the rm cron line was not there, it had stayed. I've been playing with this since 6 EST tonight and its been smoothly been changing and doing exactly as expected. Let push it to dev, lets get some more hands testing it.

PR #717, opened 2017-10-15
raise SMB-disable threshold to > 20% of bg to catch fast carb rises

scottleibrand:
We added a safety check to avoid SMBing when delta is >10% of BG, in order to avoid overreacting to a noisy sensor, calibration jump, or other "fake" rise.  However, that threshold also prevented SMBing when it's needed most: when BG is rising fast from uncovered fast carbs.  This changes the threshold to 20%, which we've found is sufficiently high to allow SMBs when really needed.

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -714,9 +714,9 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
         //rT.reason += "minGuardBG "+minGuardBG+"<"+threshold+": SMB disabled; ";
         enableSMB = false;
     }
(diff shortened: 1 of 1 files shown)


tim2000s:
Just been hit by the 10% limit. Agree this is needed (one or two fast carbs 😉).

mgranberryto:
Same here.  The rise didn't last terribly long, though, so I don't know how much it really matters and whether it's worth defeating a safety measure.

tim2000s:
Given the amount of carbs I'd taken, the SMB was needed....

applehat:
Yeah, this got me when I was low and ate some (ok, way too many) rescue
carbs. Spiked pretty good before SMB kicked in.

PR #718, opened 2017-10-15
Upload entries from G5 to nightscout

jaylagorio:
By choosing ``G5-upload`` as the CGM type during setup OpenAPS will automatically upload CGM data from a directly connected Dexcom G5 receiver as new data is read. If the device loses the Internet connection the entries will buffer until the network connection returns, then the buffered data gets uploaded to Nightscout.

Similarly, if the receiver is disconnected from the device for an extended period of time it will upload all data since the time of disconnection, or the previous 24 hours, when it is reconnected.


scottleibrand:
Nice.  This is quality work.

This looks like a good approach if the g5-upload rig is the only uploader.  But if there are other uploaders (other g5-upload rigs, or more commonly the G5 app and Share bridge), would we rather it only upload to fill in gaps?  Or is it better to re-upload everything?

jaylagorio:
My idea behind this is that this is the only uploading rig, so whatever data points it has are relevant and should go to Nightscout. I think for there to be other G5-uploader rigs you'd have to be wearing a second transmitter because although you can only pair one receiver and one phone, you can't pair two receivers. And if the G5 app and Share bridge are an option, the person would use that.

In my particular situation I don't have a phone that supports the G5 app, and since now they don't support Android < 6.0 for new downloads I'm probably not the only person in that boat.

scottleibrand:
Ok, for that use case this would work well.  I'm already thinking about extending this to take over for the current mess that is g4-upload, so thinking about what the requirements would be there.  :-)

jaylagorio:
I did test it on a fresh install but would like someone with a non-Edison device give it a shot too. I need to resurrect my Raspberry Pi but haven't had a chance yet.

scottleibrand:
Ok.  If no one volunteers here, I'll go ahead and merge it so we can get that testing done from the dev branch.

jaylagorio:
Great!

Issue #720, opened 2017-10-15
Warning about currenttemp rate when updating rig


bdr1177:
Reinstalled 0.6.0 dev and got the following error: `Warning: currenttemp rate 0.4 != lastTemp rate 0.3 from pumphistory; setting neutral temp of 0.3.. Setting neutral temp basal of 0.3U/hr`  An openaps-set temp was running at the time that I reinstalled and that seemed to start off the errors.  Despite the errors, the rig seemed to be successful at setting new temps.

scottleibrand:
Did you keep getting the same warning repeatedly? Or just once and then it went back to looping normally?

bdr1177:
The warning has been repeating for the past two hours ever since I did a clean install of 0.6.0 dev.

scottleibrand:
I'm confused. If it's giving you that error it can't set temps normally. Has it only been setting neutral temps this whole time? You said it also set a zero temp...

scottleibrand:
Can you post the contents of settings/temp_basal.json and the most recent snippet of pumphistory.json?

bdr1177:
```
Oct 15 19:35:52 natopenaps pump-loop.log: Autosens ratio: 1; Basal unchanged: 0.3; target_bg unchanged: 130; ISF unchanged: 190
Oct 15 19:35:52 natopenaps pump-loop.log: currenttemp: { duration: 25, rate: 0.3, temp: 'absolute' } lastTempAge: 7 m tempModulus: 2 m
Oct 15 19:35:52 natopenaps pump-loop.log: determine-basal://text/shell/enact/suggested.json
Oct 15 19:35:52 natopenaps pump-loop.log: reporting enact/suggested.json
Oct 15 19:35:52 natopenaps pump-loop.log: {"temp":"absolute","deliverAt":"2017-10-15T23:35:52.250Z","reason":"Warning: currenttemp rate 0.3 != lastTemp rate 0 from pumphistory; setting neutral temp of 0.3. 25m left and 0.3 ~ req 0.3U/hr: no temp required"}
[...]

PR #725, opened 2017-10-17
fix broken safety checks with a single pump-loop type

scottleibrand:
Per https://github.com/openaps/oref0/issues/720 the recently added safety checks caused a regression for users of the non-SMB pump-loop.  To fix that, and make sure that future changes to oref0-pump-loop are more thoroughly tested, this deprecates the old pump-loop and consolidates both SMB and non-SMB use cases onto the one formerly known as smb_pump_loop.  The actual bolus commands are not executed if --microbolus is not passed, but everything else is done the same for both loop types.  This results in some extra status checks for non-SMB users, but that shouldn't hurt anything, and helps avoid problems like we saw with the safety checks making assumptions about the state of various data files that weren't met when the basal-only pump-loop did things in a different order.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -53,19 +53,23 @@ main() {
             && refresh_old_profile \
             && touch /tmp/pump_loop_enacted -r monitor/glucose.json \
             && ( smb_check_everything \
-                && if ( [[ $1 == *"microbolus"* ]] && grep -q '"units":' enact/smb-suggested.json); then
-                    ( smb_bolus && \
+                && if ( grep -q '"units":' enact/smb-suggested.json); then
+                    if [[ $1 == *"microbolus"* ]]; then
+                        ( smb_bolus && \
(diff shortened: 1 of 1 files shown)


scottleibrand:
To test: `cd ~/src/oref0 && git fetch && git checkout pump-loop-type && git pull && npm run global-install`

bdr1177:
I'll plan to test this tonight.

scottleibrand:
Merged to dev for testing there, as it fixes a major bug that is known to be preventing people from looping.

Mackwe:
Have been off-grid for a week. Updating the rig with latest 'dev'. Have you got any feedback from users yet? (Before I go live with this update)? Thanks!

scottleibrand:
Seems to be working well.

Mackwe:
Awesome! And those people are running without the 'microbolus' switch in runagain?

scottleibrand:
Yep.  And in related news, looks like we're going to retire that switch to simplify things further.

PR #724, opened 2017-10-17
BlueZ update - min  version 5.47 (BlueBorne vulnerability fix)

infoneo:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-1000250
http://www.bluez.org/release-of-bluez-5-47/

PR #726, opened 2017-10-17
Fix bluez upgrade

scottleibrand:
If Bluetooth is already installed, oref0-online will attempt to start it back up while we're downloading and making bluez (to upgrade it per #724).  This change adds another killall bluetoothd again immediately before trying to install the newly compiled binary to prevent the copy from failing.


scottleibrand:
To test, `cd ~/src/oref0 && git fetch && git checkout fix-bluez-upgrade && git pull` and re-run oref0-setup / oref0-runagain.

jdhigh:
Works!

PR #728, opened 2017-10-17
timeout ns-get curls after 30s

scottleibrand:
@cbrenner reported at https://gitter.im/nightscout/intend-to-bolus?at=59e6566ad6c36fca31635aeb that his rigs have been losing connectivity and ns-loop has been getting stuck and not recovering, stopping his loop.  This is likely because a curl gets into a state where it never completes until timing out.  We already limit ns-upload curls to 30s, so this does the same for ns-get.

### DIFF
--- a/bin/ns-get.sh
+++ b/bin/ns-get.sh
@@ -57,7 +57,7 @@ case $1 in
     fi
     test -z "$NIGHTSCOUT_HOST" && usage && exit 1;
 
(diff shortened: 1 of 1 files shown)


scottleibrand:
To test: `cd ~/src/oref0 && git fetch && git checkout ns-get-timeout && git pull && npm run global-install`

cbrenner:
Fix seems to work. Haven't had this issues since I updated a few days ago.

PR #729, opened 2017-10-18
Fix pump-loop

scottleibrand:
The fixes to oref0-pump-loop require the SMB reports and aliases to be defined.  This installs them, regardless of whether SMB is enabled.


bdr1177:
This fix worked for me for looping without SMB.

PR #730, opened 2017-10-18
use old ns-loop for MDT CGM for now

scottleibrand:
MDT CGM relies on a custom ns-loop.  Until we update oref0-ns-loop to support that, we'll want to keep using the old one for MDT CGM users for now.


CaptainBalou:
I tried mdt-ns-loop which used the old code for MDT and it worked well for me. Upload restarted immediately.

PR #732, opened 2017-10-19
round insulinReq down to safely allow eSMB on small basals

scottleibrand:
Now that eSMB allows maxSMBBasalMinutes > 30, and ZTpredBGs do a better job of preventing excess insulinReq from UAM unless there's enough carbs or time to zero temp to prevent a low, it should now be safe to allow the use of SMB on smaller basals.  The main change required to support that is that we need to round down all SMB amounts, such that (when in insulinReq/2 mode) a 0.1U SMB is not given until insulinReq reaches 0.2U, and 0.2U requires 0.4U, etc.  (Previously, a 0.11U insulinReq would generate a 0.06U SMB suggestion, which would be rounded to the nearest 0.1U.  Now, that will be rounded down to zero.)

All the same "please be careful and test everything thoroughly" caveats for SMB and eSMB still apply, but there should no longer be any technical limitation on how small basals can be to allow SMBs.  (If basals are small, SMBs will be rare, except when larger amounts of insulin are required, such as if you want to use no-bolus eSMB and enter carbs but don't bolus, for example).

We may want to keep the recommendation in place for now against SMB on small basals, but this should make things safer for anyone whose basals are near the threshold, and allow SMB to kick in when really needed.

Thoughts?  I'd like to get some discussion on this one before merging.

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -985,8 +985,7 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
         //console.error(lastBolusAge);
         //console.error(profile.temptargetSet, target_bg, rT.COB);
         // only allow microboluses with COB or low temp targets, or within DIA hours of a bolus
-        // only microbolus if 0.1U SMB represents 20m or less of basal (0.3U/hr or higher)
-        if (microBolusAllowed && enableSMB && profile.current_basal >= 0.3 && bg > threshold) {
+        if (microBolusAllowed && enableSMB && bg > threshold) {
             // never bolus more than maxSMBBasalMinutes worth of basal
             mealInsulinReq = round( meal_data.mealCOB / profile.carb_ratio ,3);
             if (typeof profile.maxSMBBasalMinutes == 'undefined' ) {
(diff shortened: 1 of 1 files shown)


andrew-warrington:
Could this be a conditional, i.e. round down if basal is near or below threshold, and keep current behavior if above? And then in preferences.json there could be an override which defaults to "auto" but could be changed to "round-down" or "round-up" (or whatever is appropriate as a descriptive label).

scottleibrand:
Yes, it could, but the round-down behavior is safer for everyone.  Can you think of a case where it'd be inappropriate?  With larger basals and SMBs, you're just going to calculate another one in 5m anyway...

andrew-warrington:
My understanding was that the SMBs are currently delivery 1/3rd of the calculated need. I did have one case of having a hypo post SMBs which were delivered which glucose was on a down trend, but on the up trend I find SMBs are only barely able to keep up with glucose rises, even with announced meals. I personally would not choose to further limit SMB activity, at least not during the up trend. YMMV. (As implied by all this, I would love a preference switch to prevent SMBs during a down trend.)

scottleibrand:
@andrew-warrington If you're not already, you should try eSMB (with maxSMBBasalMinutes) and the latest dev.  The SMB calculations are now based on insulinReq/2, not insulinReq/3.

bdr1177:
@scottleibrand I'll test this over the weekend.  Is this now in the main dev or do I need to install a particular branch to test?

scottleibrand:
This PR is open, not merged, so you'll want to test the branch it's proposing to merge (tiny-basal-eSMB).

PieterGit:
The code changes looks good to me and rounding smb's downs seems good to me. 
I didn't test it on a rig yet. 

I would limit the number of algorithm options as @andrew-warrington suggests in https://github.com/openaps/oref0/pull/732#issuecomment-337812063 and try to only limit options too stuff that is meaningful in diabetes treatment and minimize the features on the algorithm. E.g. in this case perhaps a (minimum and) maximum limit on each SMB might be useful.
On the other hand that limit is currently implemented with `maxSMBBasalMinutes` which seems appropriate enough for me.

sulkaharo:
Re: defaults / configuration discussion above - FWIW my observation of designing user interfaces over the years is that vast majority of users never change the defaults they use, unless absolutely necessary and when they do, the settings are often changed based on a suggestion from someone and not based on understanding the settings. So in terms of OpenAPS, less configurability AND always defaulting to the safest config possible is the way to go, through assuming the users will most of the time not have the level of understanding needed to really fine tune the OpenAPS settings. (Myself included.)

scottleibrand:
So based on the discussion so far, it sounds like we're positively disposed to merging this?  Any objections or considerations before doing so?

bdr1177:
We've been testing this over the weekend and this has worked well so far.  It's been very conservative and we've had few SMBs, but that seems appropriate given our small programmed basals of between 0.25 and 0.35.  We'll continue testing, but I haven't seen any reason to cause concern so far.

PR #733, opened 2017-10-20
make sure all apt-get packages are up to date

scottleibrand:
Among other things, this should make sure that rigs get updated to protect against the Krack attack on WPA2 (wifi encryption).


scottleibrand:

PR #734, opened 2017-10-20
optionally allow SMBs with activity mode

scottleibrand:
When using eSMB in no-bolus mode, setting an activity mode at mealtime becomes a somewhat binary proposition, as it by default disables all SMBs, making it dangerous to do so before most of the mealtime SMBs have been administered.

This change allows setting allowSMB_with_temptarget to true in preferences.json to leave SMBs enabled even when high temp targets are active.  In combination with exercise mode, this allows high temp targets to be used to make SMB less aggressive, but still properly handle meal insulin delivery.  This should make it much easier to inform oref0 that post-meal activity is likely to cause BG to drop, and cause it to be more conservative in those situations.

This also switches CSF calculation back to using the non-autosens adjusted CR, so that high temp targets take full effect at mealtime.  It applies an upper limit to the amount of carb impact that will be projected to continue, limiting the aggressiveness of SMB with large fast-absorbing meals (whose absorption often falls off sharply once insulin activity levels rise high enough to switch the liver over from a net source to a net sink of BG).  And it allows COB to still be calculated even when one or more BG readings are LOW.  (While BG is less than threshold, it will still zero-temp based on minGuardBG, but as BG starts rising again it will have a correct COB calculation to decide when to resume insulin delivery.)


PieterGit:
@scottleibrand Why not split `enableSMB_with_temptarget` to `enableSMB_with_low_temptarget` and `enableSMB_with_high_temptarget`. Those preference names seem more appropriate, and I would avoid using enable and disable boolean preferences.

scottleibrand:
Because we're not using high temp targets to enable SMB, just turning off the default behavior of having them disable it. So the double negative is more accurate. 

Perhaps allowSMB_with_high_temptarget would be more intuitive?

PieterGit:
👍  for `allowSMB_with_high_temptarget`. Using booleans with a double negative is quite confusing.

scottleibrand:
Changed to `allowSMB_with_high_temptarget` and updated the PR description.

scottleibrand:
I'm ready to merge this to dev for further testing there if no objections.

PR #739, opened 2017-10-22
Optimize oref0-ns-loop.sh

scottleibrand:
This includes various optimizations to make oref0-ns-loop run faster, particularly for Pi0 rigs.  It also causes it to skip non-essential ns-loop functions (everything other than getting new BG data) if system load is too high, to ensure that oref0-pump-loop can run.


scottleibrand:
Merging this to the `offline` branch.

PR #740, opened 2017-10-22
Limit carb impact predictions to 30g/h

scottleibrand:
When SMB'ing for very large meals of 120g (the maximum COB oref0 allows by default before capping it), we've seen low BG situations with lots of COB, presumably because insulin activity is so high that the liver starts storing all the incoming carbs as glycogen.  In order to help mitigate this, we want to prevent oref0 from assuming that very high current deviations will continue.  This change caps the maximum projected deviations at a rate equivalent to 30g/h for purposes of calculating COBpredBGs.  The slightly lower projected deviations slightly reduce how much it SMBs up front, and instead causes it to predict the carbs to hit a bit later on.  This does not affect UAMpredBGs, so they can start out rising faster than COBpredBGs, but are limited to 3h of total impact, so end up lower in this scenario.

### DIFF
--- a/lib/determine-basal/cob.js
+++ b/lib/determine-basal/cob.js
@@ -116,7 +116,7 @@ function detectCarbAbsorption(inputs) {
         var delta;
         if (typeof(bucketed_data[i].glucose) != 'undefined') {
             bg = bucketed_data[i].glucose;
-            if ( bg < 40 || bucketed_data[i+3].glucose < 40) {
+            if ( bg < 39 || bucketed_data[i+3].glucose < 39) {
                 process.stderr.write("!");
                 continue;
             }
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -345,6 +345,8 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     // calculate current carb absorption rate, and how long to absorb all carbs
     // CI = current carb impact on BG in mg/dL/5m
     ci = round((minDelta - bgi),1);
+    // limit Carb Impact to 12 mg/dL per 5m
+    ci = Math.min(12, ci);
     uci = round((minDelta - bgi),1);
     // ISF (mg/dL/U) / CR (g/U) = CSF (mg/dL/g)
     // use autosens-adjusted sens to counteract autosens meal insulin dosing adjustmenst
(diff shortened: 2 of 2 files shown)


scottleibrand:
This just adds a safety check, so I think it's good to merge if no objections.

tepidjuice:
30g/h seems too low to me. However, I rarely eat a single meal with more than 45g of CHO so my experience is probably not applicable here. I tend to see SMB's not occurring quickly enough.

Perhaps in this case people need to drop their maxIOB instead?

scottleibrand:
@tepidjuice what is the highest carb impact you routinely see (in mg/dL/5m) for more than 15m in a row?

We also lowered max_iob at the same time I made this change, but I don't want to rely on user-set maximums to bandaid over deficiencies in the algorithm.

tepidjuice:
I'd say up to 4 g/5m of CHO absorption would be typical giving a CHO impact of 4*csf. A typical post meal rise without insulin would be somewhere between 0.5-1.25 mmol/L/5m. But like we discussed on gitter there is a steep rise in CHO absorption which is not captured by the algorithm. 

It seems like this would mainly be an issue for those on FIasp? The main advantage of SMB's is preloading insulin to prevent hyperglycaemia. In my experience delaying SMB's with novorapid would take away a lot of the benefits.

Also are you sure it is a deficiency in the algorithm? Isn't this what maxIOB is there to prevent? If you're correct is saying that:

> presumably because insulin activity is so high that the liver starts storing all the incoming carbs as glycogen.

then maxIOB should be set to prevent that from occurring.

I don't know if this is would be a problem in real world use. If this gets merged and I see problems I can always change 30 to 40 or 50 on my own rig.

scottleibrand:
Thanks.  I do think there are some other interactions that come into play here.  As you noted, aggressive SMBs are mostly needed up front, when remainingCI dominates, and this 30g/h limitation on the extrapolation of current carb absorption would not have any effect.  (If necessary, remainingCI can exceed 30g/hr at its peak: this change doesn't touch that part of the algorithm.)  The maxCI we're setting here comes into play when actual observed carb impact is at its peak.  In that case, assuming that really rapid carb absorption will continue can be problematic in two scenarios, both related to large meals that won't absorb all at once at the same rate.  The first situation is where there are some fast-acting carbs in the meal, followed by slower-acting ones.  If we assume the whole meal will absorb at the rate of the fastest carbs that hit first, we might overestimate how quickly the rest of them will absorb.  The second situation is where the speed of absorption (BG impact) is actually a function of the insulin concentration (due to the liver kicking in).  That can be partially mitigated by setting maxIOB, if you know to do that, but this seems to me like a relatively costless way to do so in the algorithm as well.

I suspect you won't see this kicking in for <= 45g meals, as by the time carb absorption has a chance to ramp up to full speed, half of those carbs have already absorbed, and UAM is likely taking over anyway.  But if anyone does observe mid-meal SMBs coming too slowly on fast-absorbing meals, to the point where they're actually wanting to change this in the code, we can make it a (non-displayed) preference at that point.

PR #742, opened 2017-10-22
disable SMB if either avgDelta is >20%

scottleibrand:
When hot-swapping a sensor and looping off raw data, there is a very large initial jump in the reported raw BG value, which then quickly comes down and levels out at whatever level the new sensor is reading.  The current SMB safety check disallows SMBs on the initial jump, but allows them on subsequent readings (if IOB is low enough to warrant dosing off the high value despite the downtick).  To prevent SMBing off a falsely high reading, this change disallows SMBs if any of the delta, short_avgdelta, or long_avgdelta is > 20% of the current BG.

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -168,6 +168,7 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     //var minDelta = Math.min(glucose_status.delta, glucose_status.short_avgdelta, glucose_status.long_avgdelta);
     var minDelta = Math.min(glucose_status.delta, glucose_status.short_avgdelta);
     var minAvgDelta = Math.min(glucose_status.short_avgdelta, glucose_status.long_avgdelta);
+    var maxDelta = Math.min(glucose_status.delta, glucose_status.short_avgdelta, glucose_status.long_avgdelta);
 
     var profile_sens = round(profile.sens,1)
     var sens = profile.sens;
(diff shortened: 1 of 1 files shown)


sulkaharo:
Related - Nightscout has the sensor age feature, so another approach would be to disable SMB for first n hours of a new sensor?

scottleibrand:
If you'd like to add and test that, feel free.  We don't enter sensor change events, and *want* SMB enabled even on a brand new sensor (after the first few minutes), so that wouldn't be helpful for us.

scottleibrand:
Unless anyone wants to try to specifically test this somehow, I'll merge it to dev for real-world testing there.

tim2000s:
If all you’ve done is change the configured threshold in determine-basal from 0.1 to 0.2*bg, then I hardwired the rig to do that in the previous dev version with no issues.

scottleibrand:
@tim2000s no, that was a previous commit that's already been merged.  This is about checking the short and long average deltas as well as the 5m delta.

jaylagorio:
It wasn't readily apparent to me, what timeframe are the short and long average deltas measured over?

scottleibrand:
15ish and 30ish minutes. I think short is and average of the data from 7-22m ago, and long is 22-42m or something like that.

jaylagorio:
And this won't prevent SMB during those sudden rises at the beginning of a meal when doing no-bolus?

scottleibrand:
We don't see deltas > 20% from that, so we shouldn't see avgDeltas >20% either. (They're an average in units of mg/dL/5m, not a total.)

Diadon81:
Doesn’t work well for my child, there are was several cases, when BG rise up faster then 20% threshold and one case when rig  disable smb, after short period of connectivity problems between xdrip and rig.

scottleibrand:
Are you bolusing or doing no-bolus? How long were SMBs disabled?

Diadon81:
Yes I am doing prebolus 1u, after this BG starts to rise up very quickly and there was no smb. When I start to check why, I found the message like smb disables because of avgDelta > 20%, after approximately 20 minutes (I amn’t sure about a time) it start to do smb.

Another case which happened several days ago. I am using freestyle Libre together with NS, after the meal my child rig lost connectivity with internet for 15 minutes approx. when connection came back, bg rise up from 5.5 to 8.9 mmol and I saw same message in the log.

tim2000s:
This has had some interesting side effects. I’m seeing fast carbs (and in some cases a fast reaction to less fast carbs) disabling SMB precisely when it is most needed. 

I’m wondering whether the Noise observation will render this invalid as it does seem to affect how SMB works quite dramatically. Maybe it needs a minor modification that something along the lines of:

if COB > 10 & avgDelta > 20% allow SMB
Else disable SMB

Diadon81:
I suggest to add parameter to preferences.json, which allow to turn off this feature, so everyone can analyze a risksand make their own decision.
At least for the fast carbs and for child with very quick absorption rate, this feature doesn’t work well.

PR #743, opened 2017-10-22
pad up to 90m of zero deviations after site change

scottleibrand:
Last night we saw an example of a site change followed by lots of positive deviations (likely meal-related, but not all excluded as such) that caused autosens to detect its maximum allowed resistance, and subsequently caused oref0 to be a bit too aggressive with SMB.  This change makes autosens pad up to 6 additional zero deviations after a site change, so it'd be less likely to detect resistance based on too-little data.


scottleibrand:
Unless someone wants to explicitly test this, I'll go ahead and merge this to dev for real-world testing there.

PR #744, opened 2017-10-22
Only SMB every 3m; retry up to 2m

scottleibrand:
Now that we have a mechanism to retry SMBs when we are Waiting to microbolus again, we can allow SMBs a little less frequently, to reduce the chance of getting two SMBs based on the same BG data (and potentially delivering the full insulinReq, not just half of it).


jaylagorio:
These look good!

PR #745, opened 2017-10-22
check validity of iob.json and profile.json and refresh if invalid

scottleibrand:
Sometimes `get-settings` fails to refresh something from the pump, resulting in a bad/empty iob.json or profile.json.  Rather than letting the loop fail and waiting for it to refresh naturally, this does so before trying to use it.

This also refactors the `openaps get-settings` alias into a function in oref0-pump-loop for better logging, efficiency, and to make things easier to tweak in the future.


scottleibrand:
I think this one is ready to merge to dev if no objections.

PR #746, opened 2017-10-22
adding cron start/stop aliases

Kdisimone:


bin/oref0-log-shortcuts.sh
```diff
@@ -20,6 +20,8 @@ grep -q git-branch ~/.bash_profile || echo "alias git-branch="'"cd ~/src/oref0 &
 grep -q runagain ~/.bash_profile || echo "alias runagain="'"bash ~/myopenaps/oref0-runagain.sh"' >> ~/.bash_profile
 grep -q edison-battery ~/.bash_profile || echo "alias edison-battery="'"cd ~/myopenaps/monitor && cat edison-battery.json"' >> ~/.bash_profile
 grep -q cat-reservoir ~/.bash_profile || echo "alias cat-reservoir="'"cd ~/myopenaps/monitor && cat reservoir.json"' >> ~/.bash_profile
+grep -q stop-cron ~/.bash_profile || echo "alias stop-cron="'"cd ~/myopenaps && /etc/init.d/cron stop && killall -g oref0-pump-loop"' >> ~/.bash_profile
```
scottleibrand: No need to cd here.

Kdisimone: I like it because usually when I'm going to use stop-cron, it's because I'm going to be using aliases in myopenaps dir.  might as well kill two birds with one stone.

jaylagorio: Nice. I would have made another shell script but this is smarter.

PR #747, opened 2017-10-25
Calculate UAM and eventualBG when insufficient data for COB

scottleibrand:
When there is less than 3h worth of glucose data available, oref0-meal can't calculate COB, so returns zero.  However, if UAM is enabled, it also fails to return UAM information, which was propagating through to undefined / NaN for eventualBG in that situation (basically turning oref0 into just a predictive low glucose suspend system). 

This change only sets COB to zero when insufficient glucose data is available, but still calculates the UAM data if it can.  It also adds an undefined check to prevent missing UAMpredBGs from causing an undefined/NaN eventualBG.  And it adds a check to the 24h glucose refresh to force it to retry every run until it gets enough glucose data to calculate COB.

This also includes various optimizations from #741 to make oref0-ns-loop run faster, particularly for Pi0 rigs. It also causes it to skip non-essential ns-loop functions (everything other than getting new BG data) if system load is too high, to ensure that oref0-pump-loop can run.

### DIFF
--- a/bin/oref0-meal.js
+++ b/bin/oref0-meal.js
@@ -85,11 +85,6 @@ if (!module.parent) {
       basalprofile_data = temp;
     }
 
-    if (glucose_data.length < 36) {
-        console.error("Optional feature meal assist disabled: not enough glucose data to calculate carb absorption; found:", glucose_data.length);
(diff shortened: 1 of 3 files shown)


scottleibrand:

Issue #737, opened 2017-10-22
Edge case when SMB variables set in preferences, but not in enable line


danamlewis:
@Kdisimone reported a possible edge case, where a new rig was set up, the preferences for SMB were added and turned to true, but was not added via the setup script (i.e. with safety catch phrases) or to the enable line.

Behavior resulted in it setting zero temps; calculating and attempting to SMB; but not succeeding with SMB. 

Possible proposed ways forward:
1) Remove the requirement of the safety phrases in the setup script, and just have people enable SMB by setting the requisite preferences in `preferences.json`
2) Add a safety guard against the above scenario by a stronger check for SMB in the enable line regardless of the preferences set.

Thoughts on option 1? People seem to know what they're doing nowadays with regards to self-selecting into SMB and otherwise waiting to get their heads wrapped around the basics before they get started.

tynbendad:
i'm good with letting just preferences.json enable it.  simplification is better anyway, and it should be pretty clear you should read about/understand a variable in there before changing it.

scottleibrand:
If we want to go with option 1, that would look like https://github.com/openaps/oref0/compare/dev...smb-in-preferences-only (It's ready to just hit the button to make that a PR if that's the consensus).

Kdisimone:
I’d vote for preferences option...lol. Obviously my brain needs simple.

jaylagorio:
``preferences.json`` only sounds good to me.

scottleibrand:
@danamlewis or @kdisimone I'll leave it to one of you to open the PR if that's the way you want to go.

bdr1177:
I'd also vote for the preferences option.

scottleibrand:
#748 is now merged, so closing this.

PR #748, opened 2017-10-25
Removing "enable" requirement for SMB - use preferences only 

danamlewis:
Per discussion in https://github.com/openaps/oref0/issues/737, moving forward with the community-voted option of removing the requirement to "enable" SMB, and only have to turn it on via preferences, to prevent edge case and unnecessary confusion. This needs testing before it merges.


scottleibrand:
To test: `cd ~/src/oref0 && git fetch && git checkout smb-in-preferences-only && git pull && npm run global-install` and then re-run oref0-setup.

scottleibrand:
Any volunteers to test this so we can merge it?

efidoman:
I'm testing it now. I switched to this branch, ran global install, took microbolus out of oref0-runagain.sh and re-ran setup. It is looping now with  `Starting oref0-pump-loop at Thu Nov  2 10:51:02 EDT `

However, I need to wait for lunch time to test SMB's. My current preferences are as follows:

`{
        "max_iob": 9,
        "max_daily_safety_multiplier": 13,
        "current_basal_safety_multiplier": 14,
        "autosens_max": 1.5,
        "autosens_min": 0.5,
        "rewind_resets_autosens": true,
        "adv_target_adjustments": false,
        "exercise_mode": false,
        "unsuspend_if_no_temp": true,
        "enableSMB_with_COB": true,
        "enableSMB_with_temptarget": true,
        "enableUAM": true,
        "curve": "rapid-acting"
}
`

efidoman:
SMB isn't working for me. I think the problem may be unrelated to this branch's changes. My NS site is up and running and getting treatments, but openaps isn't able to communicate with it for some reason. I can't figure out why yet.

efidoman:
OK - I got my unrelated npm install issues all sorted out and fixed. Testing this branch and it works great for me. 👍 

SMB's are working and I didn't use enable microbolus when I re-ran oref0-runagain.sh.

scottleibrand:
Thanks.  Based on @efidoman's testing I think we're good to merge this if nobody else wants to test anything specific before it goes to dev.

PR #750, opened 2017-10-26
only use monitor/edison-battery.json if it exists

scottleibrand:
oref0-ns-loop was failing to format_ns_status on non-Edison rigs because edison-battery.json doesn't exist.  This changes it to only try to include that if it exists.


jbwittmer:
I wondered what was going on.  Glad I hadn't drilled down to try and figure it out.  Do you need someone to test this?  I can pull and install on a pi rig when I get home from work in the morning.

scottleibrand:
I just tested it on my pi, and it's a simple file existence check, so I think I'll merge it into dev for further testing there.

PR #751, opened 2017-10-26
match on spidev5.1 to indicate explorer (for runagain); fix if syntax

scottleibrand:
PR #622 introduced a syntax error (bash if comparisons require square brackets) and failed to match when run non-interactively via oref0-runagain.  This changes the conditional to just match on spidev5.1 as we do elsewhere to indicate that we're on an Edison + Explorer Block rig.  cc @jsmurray1

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -26,7 +26,6 @@ DIR=""
 directory=""
 EXTRAS=""
 radio_locale="US"
-explorer1="NO"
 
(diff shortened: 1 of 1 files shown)


scottleibrand:
To test: `cd ~/src/oref0 && git fetch && git checkout edisonvoltage-fix && git pull && npm run global-install`, then re-run oref0-setup or oref0-runagain.

tynbendad:
this worked for me - fixed my rig's ns-loop, thanks!

PR #752, opened 2017-10-26
Fix edisonbattery for MDT and xDrip.

scottleibrand:
When I removed the edisonbattery import, that broke MDT (and xDrip) CGM users who still need the old `openaps ns-loop` for now until we update oref0-ns-loop.sh to support them.

This restores edisonbattery upload functionality for MDT, and restores that plus ns-loop functionality for xDrip.


scottleibrand:
To test: `cd ~/src/oref0 && git fetch && git checkout mdt-edisonbattery && git pull && npm run global-install` then re-run oref0-setup or oref0-runagain.

jdhigh:
Works! The Edison voltage pill now appears in NS for mdt. (Can't speak for xDrip)

PR #753, opened 2017-10-26
 press ESC three times on the pump to exit Bolus Wizard before SMBing

scottleibrand:
To reduce the risk of A52 errors, @mgranberryto resurrected the idea in https://github.com/openaps/oref0/pull/749#issuecomment-339441573 of having OpenAPS send ESC key presses before issuing SMBs.  This does so, three times, to make sure the pump is out of the pump Bolus Wizard before SMB'ing.  That will force the user to go back into the bolus wizard after the SMB completes, when it is again safe to do so.  This might also help with accidentally leaving the pump in the fixed prime screen after a rewind, but that will need to be tested.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -240,6 +240,10 @@ function smb_bolus {
     # and administer the supermicrobolus
     find enact/ -mmin -5 | grep smb-suggested.json > /dev/null \
     && if (grep -q '"units":' enact/smb-suggested.json); then
+        # press ESC three times on the pump to exit Bolus Wizard before SMBing, to help prevent A52 errors
+        openaps use pump press_keys esc
+        openaps use pump press_keys esc
+        openaps use pump press_keys esc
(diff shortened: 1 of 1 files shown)


bin/oref0-pump-loop.sh
```diff
@@ -240,6 +240,10 @@ function smb_bolus {
     # and administer the supermicrobolus
     find enact/ -mmin -5 | grep smb-suggested.json > /dev/null \
     && if (grep -q '"units":' enact/smb-suggested.json); then
+        # press ESC three times on the pump to exit Bolus Wizard before SMBing, to help prevent A52 errors
+        openaps use pump press_keys esc | jq .completed | tr '\n' ' '
```
mgranberryto: `openaps use pump press_keys esc esc esc` works for me, returning an array of three confirmations.  It would lead to less process start overhead to use that form.

scottleibrand: Thanks.  And that also gives a single summary boolean for `completed`, which we can use to bail out and not SMB if false.

PR #754, opened 2017-10-26
Fixed handling of x12 pumps - now forces them to use old_main instead…

dcacklam:
2 separate changes, one you guys will probably like, the other you very well may not:

1) oref0-pump-loop

 = Rewrote preflight to work off of 'openaps-use pump model', so it doesn't break x12 pumps
 = Changed behavior of 'main' loop, so that (A) it preflights first (sparing extra calls to prep & stuff for new_main, if we are just going to jump off to old_main because the pump is an x12), and (b) it detects x12 pumps and boots them off into old_main (which works for the x12, provided that 2 other things are fixed (the 'simulated' status call in decocare/commands.py, and the removal of the 'currenttemp != lastTemp from pumphistory' check)

The reason for forcing x12 to old_main, is that 'new main' uses a bunch of reports/commands from SMB that are not available once you say 'Yes' to x12 in the setup script (eg, x12 has enact/suggested.json, but not enact/smb_suggested.json in it's valid-report list - so the 'new' loop fails with an openaps-report syntax error). 

Feel free to use anything you like & trash what you don't.

If I have time, I'll try to come up with cleaner fixes, but at present, the ones I have here (Combined with the modification to the status function in decocare) produce a reliable loop.

One other note: 
At least on the x12, the 'currenttemp' check has the effect of randomly reverting to the hardcoded basal targets for long periods of time. 

A user who isn't watching the log obsessively may think they are looping (because the check doesn't always go haywire (defined-as: Overrides the previously enacted temp-basal with the hardcoded value at the end of every loop because it doesn't like what it sees for the last-known rate in pumphistory - even though everything was 'OK' when the now-overridden rate was enacted a few lines of code back in the same 'run' of the loop), but they are really on their (maybe too high? maybe too low?) hardcoded basal rates from settings/basal_profile.json - which could lead to excessive insulin delivery if the basal rate is on the high-side...

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -34,6 +34,20 @@ old_main() {
 
 # main pump-loop
 main() {
+        preflight
+        case $? in # Use the return status code to determine what loop
+                1)
+                        #512/712
+                        old_main
+                        exit 0
+                ;;
+                2)
+                        fail "$@"
+                ;;
+                0)
+                        echo Preflight Complete
+                ;;
+        esac
     if [[ $1 == *"microbolus"* ]]; then
         looptype="supermicrobolus";
     else
@@ -46,7 +60,6 @@ main() {
         echo && echo "Starting $looptype pump-loop at $(date) with $upto30s second wait_for_silence:"
         wait_for_bg || fail "$@"
         wait_for_silence $upto30s || fail "$@"
-        preflight || preflight || fail "$@"
         if_mdt_get_bg || fail "$@"
         refresh_old_pumphistory_24h || fail "$@"
         refresh_old_profile || fail "$@"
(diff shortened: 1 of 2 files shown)


scottleibrand:
Side question:
Are you using an IDE, or did you manually go remove all my newlines between functions?  (Those are there on purpose for readability.)

This looks like a good approach: I think I'll go ahead and incorporate something similar.

scottleibrand:
Care to test out https://github.com/openaps/oref0/pull/775 and see if it accomplishes the same thing?

dcacklam:
Everything was done with vi/sed/textutils, on the edison rig itself...
I don't remember why/wherein I did any specific formatting.

I will go ahead and test 775.

PR #755, opened 2017-10-26
don't treat suspended pump as a complete failure

scottleibrand:
When pump is suspended, the rig currently treats that as a failure and mmtunes unnecessary when pump comms are fine.  This changes that behavior to continue to refresh profile, pumphistory, etc. instead of failing out if the pump is suspended.


scottleibrand:

PR #757, opened 2017-10-26
Sensitivity and targets

scottleibrand:
Currently autosens_adjust_targets is false by default, as autosens now adjusts the basals used in calculating IOB, eliminating the need to adjust targets to compensate for IOB previously not reflecting sensitivity.  However, there are still cases where autosens doesn't react strongly enough, such as for persistent low BG due to sensitivity.  To help address that, this splits autosens_adjust_targets into two preferences: sensitivity_raises_target (true by default) and resistance_lowers_target (false by default).  As before, temp targets override any autosens-based target adjustments.

Additionally, some people find that 80 mg/dL temp targets are insufficient to bring BG down when they're running resistant, so this adds a low_temptarget_lowers_sensitivity preference (false by default) to allow people to use low temp targets (>=99) to signal resistance mode.   This is the converse of what exercise_mode does (for >=111), so for consistency I've added high_temptarget_raises_sensitivity as well as a (hidden) synonym for exercise_mode.

### DIFF
--- a/lib/determine-basal/autosens.js
+++ b/lib/determine-basal/autosens.js
@@ -280,8 +280,8 @@ function detectSensitivity(inputs) {
             //console.error(tempTarget)
         }
         if ( tempTarget > 100 ) {
-            // for a 110 temptarget, add a -1 deviation, for 160 add -6
-            tempDeviation=-(tempTarget-100)/10;
+            // for a 110 temptarget, add a -0.5 deviation, for 160 add -3
+            tempDeviation=-(tempTarget-100)/20;
             process.stderr.write("-");
             //console.error(tempDeviation)
             deviations.push(tempDeviation);
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -302,6 +302,7 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
         , 'insulinReq': 0
         , 'reservoir' : reservoir_data // The expected reservoir volume at which to deliver the microbolus (the reservoir volume from right before the last pumphistory run)
         , 'deliverAt' : deliverAt // The time at which the microbolus should be delivered
+        , 'sensitivityRatio' : sensitivityRatio // autosens ratio (fraction of normal basal)
     };
 
     // generate predicted future BGs based on IOB, COB, and current absorption rate
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -354,8 +354,14 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     ci = round((minDelta - bgi),1);
     uci = round((minDelta - bgi),1);
     // ISF (mg/dL/U) / CR (g/U) = CSF (mg/dL/g)
-    // use unadjusted profile.sens to allow activity mode sensitivityRatio to adjust CR
-    var csf = profile.sens / profile.carb_ratio;
+    if (profile.temptargetSet) {
+        // if temptargetSet, use unadjusted profile.sens to allow activity mode sensitivityRatio to adjust CR
+        var csf = profile.sens / profile.carb_ratio;
+    } else {
+        // otherwise, use autosens-adjusted sens to counteract autosens meal insulin dosing adjustments
+        // so that autotuned CR is still in effect even when basals and ISF are being adjusted by autosens
+        var csf = sens / profile.carb_ratio;
+    }
     var maxCarbAbsorptionRate = 30; // g/h; maximum rate to assume carbs will absorb if no CI observed
     // limit Carb Impact to maxCarbAbsorptionRate * csf in mg/dL per 5m
     maxCI = round(maxCarbAbsorptionRate*csf*5/60,1)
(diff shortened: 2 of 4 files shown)


scottleibrand:
To test: `cd ~/src/oref0 && git fetch && git checkout sensitivity-and-targets && git pull && npm run global-install` then re-run oref0-setup or oref0-runagain and edit your preferences.json.

AdrianLxM:
Hi, I have some remarks: https://github.com/openaps/oref0/pull/757

(1) Redundancy
--------------
My personal preference is not to have synonym or redundant settings (`sensitivity_raises_target`, `exercise_mode`) if both are adjustable. What if one is set to true and the other one to false. The user might have only one in mind and might even double-check that it is disabled - and it is still active. But that's really just a personal preference of me.

(2) normalTarget
----------------
`normalTarget` seems to be fixed to 100. Some use lower targets. Wouldn't it be better to use the real default target from the profile instead of checking for `target_bg > normalTarget + 10`?
For people using higher normalTargets as 100 couldn't the current behaviour even lead to the behaviour that a **low**-temptarget would be considered a **high**-temptarget?

(3) Precise wording
-------------------
Comment states it is synonym for `high_target_raises_sensitivity`, actual setting is called `high_temptarget_raises_sensitivity`. Unify the names?


(4) Confusion about thresholds:
-------------------------------
Description says: **this adds a low_temptarget_lowers_sensitivity preference (false by default) to allow people to use low temp targets (*>=99*) to signal resistance mode**
Comment says: `low_temptarget_lowers_sensitivity: false // lower sensitivity for temptargets <= 94.`
I think it should read *< normal_target* in both cases?


(5) Suggestion: Use reason field:
---------------------------------
As in my opinion too much is fixed with the settings: Temp targets do have a *reason* field. It already has "eating soon" and "activity". Why not add "resistance" and "controlled high" (let a native speaker invent better phrases ;) ) By using this field the user could decide if (s)he wants to have a high temp target with activity mode enabled or just be safe for driving for an hour or two when setting the temp target.

scottleibrand:
Fixed (3) and (4).

On (1), I'm open to input from others (particularly @danamlewis and @Kdisimone) on what we prefer.  I don't much care what the preferences are named.

On (2), the normalTarget is hard-coded mostly because determine-basal doesn't know what your normal target is if you have a temp basal set, as the temp target is communicated by oref0-get-profile changing the profile target to the temporary one.  But I also think it's better to have the same temp target always produce the same sensitivity changes given the same preferences, regardless of how your pump targets might shift around during the day.  I don't think anyone (is? should be?) running a normal target of 80 and then setting a high temp target of 90, or running a normal target of 140 and then setting a low temp target of 120.

On (5), I didn't really want to have to update the NS codebase (and everyone's personalized IFTTT recipes) and then have to troubleshoot those with people who didn't update them right.  So far we've stuck to ignoring the temp target reason, and just using the numerical target value.  If someone has a use case for wanting some low temp targets to do different things than others based solely on how they're entered, I'd like to understand that use case before making things more complex.

AdrianLxM:
Re (2): I usually do a low temp target at 72 mg/dl for eating soon mode. (I know this needs adjustment in determine_basal now anyway). In this case I would only want to do a simple low-temp-target. So far I haven't thought of if that should trigger the "break sensitivity mode", but would guess not.

I usually have a target of 90mg/dl. When my sensor is stable and other conditions are met (not much exercise, no alcohol,...), I rather have it a tad lower during the night. This I also do with a low temp-target for convenience. It times out for the next day.
 It could be done with a profile switch though that has a lower "default" target and and also has a duration that it falls back to the previous profile switch. This would allow to have "temporary targets with sensitivity boost" and "temporary default targets without sensitivity boost".

scottleibrand:
I don't think I understood that, so I think it would be too complicated to support that kind of use case.  :-)

AdrianLxM:
The last one was not a use case but just the idea that the default target range - at least since the latest Nightscout version, but probably much earler - also is temporary if wanted.

scottleibrand:
Are you confusing the NS target range (which is only used for display in NS) with the target range used by OpenAPS (which comes from the bolus wizard settings on the pump)?

AdrianLxM:
Ah, so OrefX does not use the NS targets? That's different in AndroidAPS. It uses the one from the current profile switch (last who is started before and whose duration has not timed out if it has a duration).

scottleibrand:
Correct: oref0 never even downloads the NS profile.

hilarykoch:
I loaded this today. I'm used to targets adjusting automatically but the target has stayed at what we've got set on the pump. Is that correct?

scottleibrand:
What are your preferences?

hilarykoch:
@scottleibrand 
img_5142

scottleibrand:
Then yes, if your autosens ratio is above or below 1, it should lower or raise target.

hilarykoch:
Got it!

scottleibrand:
@hilarykoch How's it working for you?

hilarykoch:
So alone, I don't know that it's been aggressive enough for me. I've been temp targeting a lot. But our numbers have been overall pretty good the last few days, so maybe the combination of the version plus temp targets is good? I

scottleibrand:
Anyone want to specifically test anything else before merging this to dev?  If not I think it's good to go.

PR #758, opened 2017-10-26
oref0-version

applehat:
`oref0-version` Returns version [branch] by default.
 
With `--check-for-updates` it will fetch and compare local to remote, and alert user if they are up to date or should update.

Added this to report to pump-loop (the one everyone actually watches) every hour, with instructions to update before posting any issue requests if `branch != master`.


applehat:
@scottleibrand or @danamlewis -- Feel free to change any language in the file.

 I'm not a wordsmith in the slightest, and won't be offended. :P

applehat:
Closing this PR since #759 includes it.

PR #759, opened 2017-10-27
oref0-version

scottleibrand:
Updated copy of @applehat's https://github.com/openaps/oref0/pull/758 that runs check-for-updates after each successful pump-loop instead of once an hour.


applehat:
Running ever loop could potentially cause offline loop slowdown.

I'm not sure what the timeout on a `git fetch` is, but it might be worth some testing with a rig with no internet connection before merging.

Or just run it in a non-blocking way - tho that could potentially counter the (what I'm assuming is) idea of having it super visible in the pump loop for everyone to see.

scottleibrand:
Good point: I'll make sure `git fetch` times out reasonably.  At least by running at the end of the loop like this, it's not likely slowing anything down: the next thing the rig is going to do is wait for new BG or for radio silence.

scottleibrand:
Ah, even better: `timeout 30 git fetch` kills the `git fetch` after 30s if it isn't done.  :-)

scottleibrand:
@applehat want to test this version?

PieterGit:
I think this is quite an expensive check that will waste quite some CPU cycles and possible help drain the rig battery. In a normal day I see about 1 unsuccessful pump-loop each hour (mostly because rig-pump communication error or failure). e.g.

```
# grep -i "^Unsuccess" /var/log/openaps/pump-loop.log
Unsuccessful supermicrobolus pump-loop at Fri Oct 27 07:52:00 CEST 2017
Unsuccessful supermicrobolus pump-loop at Fri Oct 27 08:25:32 CEST 2017
Unsuccessful supermicrobolus pump-loop at Fri Oct 27 08:47:29 CEST 2017
Unsuccessful supermicrobolus pump-loop at Fri Oct 27 10:12:16 CEST 2017
Unsuccessful supermicrobolus pump-loop at Fri Oct 27 10:52:04 CEST 2017
Unsuccessful supermicrobolus pump-loop at Fri Oct 27 10:56:47 CEST 2017
Unsuccessful supermicrobolus pump-loop at Fri Oct 27 11:01:22 CEST 2017
Unsuccessful supermicrobolus pump-loop at Fri Oct 27 13:02:10 CEST 2017
```

But in a situation where the pump and rig are too far apart this will do a continuous network check on the version of oref0.

Perhaps we should only mmtune and do a check like this if the pump run fails for exactly 2 times in a row,

applehat:
Im sort of torn here.

I agree that post-loop is potentially expensive (especially since we really want to move forwards using less resources as much as possible since Pi0 will be the future), but I also question how much value there is to a once-every-hour approach.

Maybe once an hour we could check git for updates and create a status file like `out-of-date.json` in the `myopenaps` directory, and in the pump-loop check for the existence of that file to show the warning? Then oref0-setup can `rm` the file if it exists, since rerunning it is part of the update process generally.

I probability of someone having an issue and then an update that fixes is going out before that 1 hour mark seems pretty low.

applehat:
^ That concept above could segway into a decent self updater script that displayed release notes and performed updates 🤔

scottleibrand:
I think @PieterGit's concern misunderstands what we're implementing here.  This is not being implemented on loop failures, but rather after each successful loop.

`time oref0-version --update` shows that the full update check, including git fetch, takes less than 2 seconds.  And presumably most of that time is spent waiting on the network, not doing anything with the CPU, so I don't think that's very expensive.

If we wanted to optimize further, I'd check the timestamp of .git/FETCH_HEAD and only update that if it's more than X minutes old.  Then we can still do the same out-of-date check without any additional complexity, but just `git fetch` ever X minutes.

applehat:
Ahh thats a good point. If only happening on success, then yeah, perfect.

Tho it may be valuable to add a check against the  timestime of `.git/FETCH_HEAD` on a failed pump loop as well, just so maximum visibility happens in the logs.

PieterGit:
I didn't know you wanted an up to date check on a succesful pump loop, and I don't see the added value of that.

What's the point of having an up to date check that frequent? Why not integrate it with e.g. the autotune script as a daily message. We had a serious regression in the 0.6.0-dev cycle (https://github.com/openaps/oref0/pull/695 ), which caused quite some hypo's for quite some days for quite some people. I don't think that (checking for/doing) updates that often is necessary. That would require far more unit tests and some kind of real world feedback loop which we don't have into place now. After 695 I don't update my dev rig if I don't see a bug fixes or new features that I'm interested in.

More reasons to be careful with non-essential extra processes:
- in case of a network problem, this PR will hold up the pump loop for 30 seconds and thus potentially cause the new pump loop to start a minute later
- I would like to be able to use rigs offline, or with very limited mobile access
- If we want to switch back to pi0 we need to be more carefull with adding extra processes.

Im 👍  on integrating this script and running it, hourly or daily. I'm 👎  on auto updating development rigs. And I would be 👍  on increasing the number of release per year (e.g. a montly or bi-montly release), so that more people use a stable version instead of lurking on the dev channel.

PieterGit:
I thought about it a bit more. I really would like to propose we stick to seperations of concerns ( https://en.wikipedia.org/wiki/Separation_of_concerns ) from a safety standpoint. I know the check won't waste a lot of network / CPU resources, but I'm still 👎  on implementing this in the pump loop.

The concern of the `pump-loop` is to communicate between rig and pump, and the `cgm-loop`'s concern is too make sure the rig has recent and correct cgm data. These loop are essential and should not be mixed with other supplementary concerns.

The concern that people need to update their rigs (especially for dev branch rigs) should be implemented differently. I propose to implement a `oref0 check` and run that in a daily cron (similar to autotune). The `oref0 check` can also be run at the end of `oref0-setup` and the discipline should also be to run it after any configuration change or in case of trouble shooting.

This is similar to `postfix check`, see http://www.postfix.org/BASIC_CONFIGURATION_README.html#syslog_howto
Postfix administrators will always do a `postfix check` after every configuration change or update.

sulkaharo:
Agree with Pieter both on the SoC principle - pump-loop should do what it indicates it should do and nothing else and in principle avoiding extra battery drain. Also not sure how many people look at their logs regularly - I know I don’t. Sending this over pushover would maybe make more sense and I would actually consider mandating the use of pushover for dev branch for increased safety.

PieterGit:
Thanks @sulkaharo for also advocating separtion of concerns.
I'm 👍  on enforcing a mandatory notification mechanism for the dev branch for increased safety. I would not implement it completely based on pushover. 

For example:  I would rather have some generic `oref0-log` or `oref0-notify` command that will log the notification (with log/syslogd in first implementation). For loglevels we should make some rules, e.g. applying the logic of http://www.postfix.org/DEBUG_README.html#logging for our case:

`panic`  indicates a problem in the software itself that only a programmer can fix. oref0 cannot proceed until this is fixed.
`fatal` is the result of missing files, incorrect permissions, incorrect configuration file settings that you can fix. oref0 loops cannot proceed until this is fixed.
`error` reports an error condition, the user should fix a.s.a.p., e.g. "no succesful pump loop has been run for 30 minutes"
`warning` indicates a non-fatal error, such as that the rig software is out of date, running on the dev branch with more than a week old software

Off course the basic logging of a certain level should be sent to the user (or notified) in one or more ways:
- very basic: a cron with a `egrep '(warning|error|fatal|panic):' /some/log/file > /some/logextract.txt` where `logextract.txt` can be shown with a webserver
- a small webservice that @hartvig suggests in https://gitter.im/nightscout/intend-to-bolus?at=59f45a32e44c43700a8a7e71 , that shows the most recent error messages
- a pushover implementation that @sulkaharo suggests
- a pebble notify implementation, that I would prefer (can also work offline)
- ...

The notification for a rig to be out of date should be implemented in a `oref0-check` or `oref0 check` outside of the essential loops. 

Separating concerns will be a bit more work, but will make the software much better to maintain and to extend in a safe way.

tim2000s:
I’ve read through this discussion and considered my use of the system. 

This version management process is something that I agree with Pieter and Sulka on. It’s not an essential component of the loop, for either the NightScout or actual loop. It’s more of a supervisor process (that you could also implement for management my some of the system items like network control and I agree with the separation of concerns viewpoint. 

I also think that auto-updating The dev branch is a bad idea. There are too many users using dev  to use features that aren’t in Master (curves for example) for whom getting a dev branch with the hypo bug could cause issues. I see no issue with an updated script diwnloading and caching the latest code version until a user initiated an install, but it shouldn’t auto update running code. 

Any auto update would also have to guarantee consistency between currently running behaviour and newly installed behaviour so may add an additional overhead.

sulkaharo:
(Technical side note -looks like tthe git check logic is likely to fail on dev rigs that have multiple remotes configured? Did you try this with a secondary oref0 fork added?)

scottleibrand:
@sulkaharo I think anyone configuring multiple remotes who wants this feature probably has the ability to submit a PR to fix that.  :-)

@tim2000s no one is suggesting auto-updating anything.  This is just an informational message.

Based on the feedback so far, I think I'll make a cron job that runs `oref0-version --check-for-updates` once an hour and writes the results to a one-line text file, and then have oref0-pump-loop `cat` that at the end of each run for visibility.

While I understand the desire for another notification mechanism, I don't want to hold up a "good enough" solution for something better.  If anyone wants to implement a pushover or other form of notification as well, feel free.  I think for most users who are testing dev and who show up in ITB to ask questions, a note at the end of each pump-loop.log run will be a suitable reminder to keep up-to-date (and to update before asking for support on an outdated version of dev).

scottleibrand:
Any other feedback or anything anyone wants to test on the revised version running hourly in cron and displaying in pump-loop.log?  If not I think this is ready to merge to dev.

PR #761, opened 2017-10-28
Changes for Offline Branch

Bender1061:
For Wifi local only hotspot
SSID is now $Hostname-Hotspot
PSK passkey is #OpenAPS 

removed hostapd  and dnsmasq from starting up automatically during boot, this was stopping rig from trying to connect in client mode. 

add to rc.local to check etc/network/interfaces file and fix if it was last left in Hotspot mode, so rig will always attempt to connect in client mode when rebooted.

Changes to oref0-online to help reduce rig getting caught in a loop that did not turn on wifi. 
as well as provide more information about what it's doing.


bin/oref0-online.sh
```diff
@@ -43,13 +64,19 @@ main() {
             #restart_networking
         fi
     fi
-    echo Finished oref0-online.
+    echo Finished oref0-online. 
+}
+
+function print_bluetooth_name {
+    echo -n "At $(date), my bluetooth is connected to "
+    echo -n ${MACs}
```
scottleibrand: This won't support configurations that let the rig pair with multiple BT phones.  MACs is a list of all MAC addresses to try.  If we want to print_bluetooth_name we need to figure out how to actually print which device it's connected to.

bin/oref0-online.sh
```diff
@@ -13,26 +13,47 @@ main() {
...
-    if check_ip; then
-        # if we are back on wifi (and have connectivity to checkip.amazonaws.com), shut down bluetooth
-        stop_hotspot
+	if ifconfig | egrep -q "bnep0" >/dev/null; then
+		print_local_ip bnep0
+	fi
+	echo -n "At $(date) my Public IP is: "
+    check_ip
+    #echo
+	if [[ $(ip -4 -o addr show dev wlan0 | awk '{split($4,a,"/");print a[1]}') = $(print_local_ip wlan0) ]]; then
```
scottleibrand: What do these checks do?  Are you just trying to check whether `print_local_ip wlan0` returns a non-empty string?  If so, `if ! [[ -z $(print_local_ip wlan0) ]]; then` would do the trick.

bin/oref0-online.sh
Bender1061: How have you been running with multiple Bluetooth Hotspots? I was thinking that would be nice. since I carry around two phones. and well yeah we all sometimes forget to leave it one bluetooth hotspot. 
I was looking into having it say the device name that it's connected too, but only way I can find even to see that is to go into bluetoothctl. would be nice to have it do that though

bin/oref0-online.sh
Bender1061: this is determining if we are connected to bluetooth or Wifi. and only telling us the name of the network, or the Mac Id that it's connected too.

bin/oref0-online.sh
tim2000s: I run with multiple Bluetooth connections by putting the Mac addresses in serially in the oref0-online cron line, separated by a space.

scottleibrand: Same. I changed this to actually print the name(s) of the phone(s) it’s connected to instead.

tim2000s: Does this require a reinstall to be made to work or will it work off a standard pull and npm global install in the dev branch?

scottleibrand: If you have BT hotspot enabled it will mostly work without re-running oref0-setup, but there are some modifications there to rc.d that will change the behavior you get when restarting the rig.  @Bender1061 had quite a few issues with his rig getting stuck in hotspot mode before he made those changes.

PR #764, opened 2017-11-01
Don't swallow errors in the pump loop.

alimhassam:
It hides errors and makes troubleshooting harder.
Showing errors printed to stderr in the pump loop
logs will encourage having a more stable system.

Alternatively if there was a reason for this, it
could be made configurable, but for now i think
it's better to keep info printed to stderr

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -115,7 +115,7 @@ function smb_reservoir_before {
     # Refresh reservoir.json and pumphistory.json
     gather \
     && cp monitor/reservoir.json monitor/lastreservoir.json \
-    && openaps report invoke monitor/clock.json monitor/clock-zoned.json 2>&1 >/dev/null | tail -1 \
+    && openaps report invoke monitor/clock.json monitor/clock-zoned.json >/dev/null | tail -1 \
     && echo -n "Checking pump clock: " && (cat monitor/clock-zoned.json; echo) | tr -d '\n' \
     && echo -n " is within 1m of current time: " && date \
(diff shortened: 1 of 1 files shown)


alimhassam:
I was using this for debugging issues, and maybe it's useful in general.
Needs to be tested again, as i had to extract these changes from a larger commit with debugging logs, and i didn't re-test it after.

scottleibrand:
These are well optimized to balance log readability with making sure to surface actual errors. Most of the underlying commands report awful page-long tracebacks on something as simple as a radio comms error, so printing all of that is not useful. In other cases, the stderr output of standard unix commands is not useful to us: depending on what exactly we're running, "file not found" might be the normal desireable condition, not an error condition. 

If there are specific error conditions that are being hidden we can adjust for those, but logging everything to pump-loop.log would definitely not be an improvement.

alimhassam:
Hmm, i don't have any awful page-long tracebacks in my logs (now that i fixed my problems) with this change. Let me review again. I do have some occasional errors, but i think they are actually real issues which needs fixing, and making them more obvious might help to get a more stable system.

You have a lot more experience than on this project, so i'm probably missing something. I still think that swallowing the errors should be the exception more than the rule.

I'll review each of these commands and check if any of them print to stderror on the normal flow.

scottleibrand:
If you'd like to work on improving error handling, the biggest need there is for someone to add proper error handling to the openaps toolkit for pump comms, such that errors there return human-usable error messages like "did not receive a valid response from pump" rather than "ValueError: need more than 0 values to unpack" or similar. For most such errors we redirect stderr through tail -1 in order to preserve just that last (and most useful) line and hide the traceback.

alimhassam:
right, i've seen that one, that's probably the main one which occurs regularly on a normal operation, thought it may manifest in different ways. 
OK, since these are well optimised, i'll close this pr, and if the need comes back work on a more polished error handling pr, rather than this quick and dirty hack.

scottleibrand:
Thanks for jumping in so quickly and looking for places you can contribute. I'm looking forward to what else you find. :-)

alimhassam:
I have another possible change, i did yesterday, not sure if it's useful let me chat you on gitter

PR #765, opened 2017-11-01
Run autosens in ns-loop

scottleibrand:
Currently we run autosens in its own loop in cron.  This results in autosens occasionally running at the same time that ns-loop is refreshing carbhistory.json, which causes autosens to generate different results (since it can't properly exclude meal periods).

This PR updates oref-ns-loop.sh to run autosens after uploading to NS, which ensures that it happens after the carbhistory refresh.  It also checks that glucose data is fresh before and after running autosens, to ensure that a slow autosens run doesn't prevent us from getting new BG data for long.

### DIFF
--- a/bin/oref0-detect-sensitivity.js
+++ b/bin/oref0-detect-sensitivity.js
@@ -99,12 +99,12 @@ if (!module.parent) {
         , temptargets: temptarget_data
         //, clock: clock_data
     };
-    // calculate sensitivity using 8h of non-exluded data
+    console.error("Calculating sensitivity using 8h of non-exluded data");
     detection_inputs.deviations = 96;
     detect(detection_inputs);
     ratio8h = ratio;
     newisf8h = newisf;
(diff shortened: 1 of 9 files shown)


PieterGit:
I'm 👎  on integrating the autosens loop in the ns-loop. It's not separating concerns, but combining them. The same rationale as https://github.com/openaps/oref0/pull/759#issuecomment-340147866 applies. 

The autosens should also work for offline usage, and therefore it's should not be tied/integrated into ns-loop.

The ns-loop cgm part is essential for those using CGM from Nightscout. The loop getting the CGM data from Nightscout should be as small as possible.  The risk of "a slow autosens run doesn't prevent us from getting new BG data for long." should be made impossible from a safety perspective. Getting fresh BG data is much more important than running autosens.

mgranberryto:
Locking is the classic solution to this problem.

scottleibrand:
Ok, if you prefer https://github.com/openaps/oref0/pull/767 I'll close this one.

PR #767, opened 2017-11-01
oref0-autosens-loop

scottleibrand:
Per discussion in https://github.com/openaps/oref0/pull/765, we need to keep autosens in its own loop to ensure it can still run offline.  This PR refactors the `openaps autosens-loop` alias into oref0-autosens-loop, adding support for checking system load and temperature before running it.

I believe the original problem of the carbhistory.json race condition will be solved by commit 931be275b1a77ac88f3f610e00373847094f6e1e, which downloads carbhistory to carbhistory.json.new and verifies it before doing `mv` over to carbhistory.json, which should be atomic.


scottleibrand:

PR #768, opened 2017-11-02
Disable wide_bg_target_range by default

scottleibrand:
When first setting up OpenAPS many people seem to set their pump's bolus wizard target range rather wide, and then ask questions about why OpenAPS isn't doing more to prevent highs and/or lows.  Most experienced users end up with a much narrower target range, usually setting the low and high targets equal (100-100 or similar).  We've long had a preference named override_high_target_with_low that instructs OpenAPS to only use the low end of the target range, which allows the high end of the range to be set higher to disable pump bolus wizard BG corrections based on non-temp-basal-aware pump IOB.

This PR makes override_high_target_with_low the default behavior, and adds a wide_bg_target_range preference (false by default) to override that and restore the previous default behavior if someone really wants a wide target range.  For now as everyone transitions over, wide_bg_target_range is displayed by default, but we may want to hide it in a future version.

This PR will likely require docs updates when merged/released to master.


mgranberryto:
Would it be better to target the midpoint instead of the low target?

scottleibrand:
Then we'd need another preference to preserve the override_high_target_with_low behavior.

sulkaharo:
IMO this is a good change! In the docs, the critical bit is to also instruct to change the Bolus Wizard high target to be on the high side (180+) OR not enter BG value into the Wizard, so the Wizard doesn't issue a correction in addition to the OpenAPS corrections. I assume some of the "why does OpenAPS zero temp after boluses" cases are actually due to user using the Wizard to correct a high in addition to bolusing for carbs.

tim2000s:
I think the main reason leave a wide target in place is a lack of trust in the system as a new user. I wonder whether this helps address that?

scottleibrand:
Easy enough for new users to set the value to true if they want training wheels. That way hey have to keep re-setting it to true on reinstall/upgrade until they decide to go narrow like everyone else.

scottleibrand:
But I think it's better for someone new to just set a higher target, not a wider target range.

alimhassam:
Still new to the algorithm, but intuitively wouldn't it be better for the algorithm to have a target range rather than a single value to aim for by default? 
This would allow it to still be okay if it overshoot by an acceptable margin. If i want my bg to be between 80 -100 i would think i should set that as the target.
But if the range is 80-100 and the bg is too high, the algo aims for 100, and overshoots by 5 or 10, it's still in range... if it aims for 80, and end up at 70, we're now low.. 

I know it's built with safety in mind, and i don't know all the details, but it seems to me that this issue should be solved more with documentation.

scottleibrand:
That's what we (and everyone) thinks at first, and why we designed it to treat to range instead of treat to target.  But what we've found is that in fact it's better, in terms of taking action to prevent lows and highs, to have the rig take action whenever eventualBG is <> 90, rather than waiting for it to be <80 or >100.  The only reason to have a wide target range would be if you're optimizing for something other than BG control, like minimizing the number of temp basals set to save battery life or something like that.  (And oref0 is decidedly not optimized for that in many other more important ways.)

alimhassam:
Ok, that seems to make sense, though i still feel like its a bit of a surprising behaviour for the default to aim for the low range of the target. 

Personally, having just been through this, my first reflex without having seen this pr was to set the range to 80-100. I just changed yesterday after studying the algo and after i gained a bit of trust to the system. 

If you think the eventual BG should set be 90, then the idea of aiming for the middle would have been better (in my case with my though process i just went through). 

Now that i understand this better i have no real objection with the change (even if i think aiming for the middle seems more intuitive)

Also, as a response to the bg wizard correction comment, what's the suggestion there? does entering bg in there affect the loop / corrections in any way? I don't know if other do / did that, I personally used to only enter a bg in the wizard if i want to see my IOB in the pump.

scottleibrand:
If you enter your BG into the bolus wizard and it calculates your eventual BG (after pump IOB kicks in) will still be above target range, it will add a correction bolus in along with your meal bolus.

We'll be sure to clearly document that the low end of the bolus wizard BG range is what's used for target, and anyone who proceeds without reading the docs should figure that out pretty quickly when they see `Target: 80` in pump-loop.log.

jodivers:
Thank you Scott for bringing this here following my need for guidance in Looped yesterday.
If I can chime in, please note that I have no back ground knowledge in the programmers’ Universe us & customs, so if this is not appropriate, please let me know.

Wizzard
It took me a while in the first month (started with 5.2 in July 2017) to figure out that is was safer to bolus without the Wizzard. I came to the realisation after a few unwanted lows caused by the Wizzard adding correction insulin. I researched, read and figured the pump did not know about the actual IOB.  It also messes up the timing of carb absorbtion because of the prebolus, using NS to add carbs is also to stress for peole needing to prebolus. So yes, as sulkaharo suggest documentation on how to bolus avoiding the wizzard is crucial.

Trust - Target VS Range
In the beginning, as tim2000s mentioned, learning to trust that amazing system and because it is the traditional way of doing it, makes us trust the wider range, so making it a target default, I agree is a good thing. Habits are hard to break, even more when we do not clearly understand why; it’s a lot of info the first few weeks, so breaking the habit for us is a good thing. I agree scottleibrand that as a beginner we should have a higher target instead of a wide range. 

I did use 5.5 as a 24/h target for my 2nd and 3rd months on version 5.2. Then used a wide range trying to fix, I do not remember what. Results were good so I kept it that way when upgrading to 5.5, but here the wide range is really messing my results, haven’t had a decent run in 3 weeks and to many lo & hi. Anyhow, following your yesterday’s comment Scott, I’m back to 5.5 day and 6.5 night (hypoed twice last night…), observing in order to trust again, because my switch to 5.5 is not going well. 

Documentation
Fianally a lot of my daily operation knowledge I got from reading Dana's blog. I read it all in about 6 weeks. I felt and still feel sometimes that the documentation doesn't yet adress all the user's opreration needs and we get lost trying stuff grab here anf there and in my case sometimes confusing it. I'm overwhelmed with it all sometimes, but I could probably help there, obviously I can't code, but I can write ;) I just do not know how to go about it.
 
Lurking here, following Dana’s recommendation, to soon upgrade to Dev, since my insurance approved FIASP for 6 months only yesterday, so I need to start testing ASAP.

tim2000s:
Hey @jodivers, if there are areas where you see gaps in the docs, please join in. Many of us have added bits and pieces where we felt more information and clarity was needed!

jodivers:
Hi @tim2000s, 

Can you guide me where and how? 
Is there an example of how it's done? 

I'm not yet familiar with these plateforms and how to use them. 

Thanks

mgranberryto:
@jodivers https://openaps.readthedocs.io/en/latest/docs/Resources/my-first-pr.html

scottleibrand:
Good discussion, thanks.  I didn't notice any objections, so if anyone has any reasons not to want to merge this please speak up.  Otherwise I'll probably do so in another day or so.

PR #769, opened 2017-11-02
change default curve to rapid-acting

scottleibrand:
Everyone who's noticed any difference with exponential curves has reported it as an improvement, so I think it's best to go ahead and change the default curve from bilinear to rapid-acting before we release 0.6.0.


mgranberryto:
🥂  I think this is one of the better changes in recent history.

sulkaharo:
Yes, this is awesome, good call.

tim2000s:
Completely agree with this one. No brainer.

PR #770, opened 2017-11-03
Add a note with the latest calculated IOB after boluses.

alimhassam:
This is useful when multiple small micro boluses are being
administered. The notes will be visible in the pushover notifications from
nightscout.

### DIFF
--- a/lib/bolus.js
+++ b/lib/bolus.js
@@ -90,6 +90,17 @@ function reduce (treatments) {
         annotate("Normal bolus (solo, no bolus wizard).");
       }
 
+      if (has_insulin) {
+        var iobFile = "./monitor/iob.json";
+        var fs = require('fs');
+        if (fs.existsSync(iobFile)) {
+	  var iob = JSON.parse(fs.readFileSync(iobFile));
+          if (iob && Array.isArray(iob) && iob.length) {
+            annotate("Calculated IOB:", iob[0].iob);
+          }
+        }
+      }
+
       if (state.bolus) {
         annotate("Programmed bolus", state.bolus.programmed);
         annotate("Delivered bolus", state.bolus.amount);


tim2000s:
I’m not sure you need IOB after every SMB. Every time the loop runs it recalculates IOB and that info is published in SMB-enacted.json, And is also available in the OpenAPS pill. Given the frequency that there loop runs, I’m not sure that you need to recalculate IOB at the end of the loop run.

alimhassam:
It's already calculated I believe, this just publishes in the notes so it shows in the pushover notification.
Personally I care more about the IOB after smb than the individual 0.3 units being delivered so it's very useful to see it in the notification.

scottleibrand:
Are you using this with oref0-pushover or the NS bolus pushover notifications?  I'm not sure how the NS ones work, but I don't believe oref0-pushover would use this by default unless you've somehow re-enabled insulinReq notifications in addition to carbsReq.

alimhassam:
This goes through NS. Works well so far. I'll keep testing it today.

lib/bolus.js
```diff
@@ -90,6 +90,14 @@ function reduce (treatments) {
         annotate("Normal bolus (solo, no bolus wizard).");
       }
 
+      if (has_insulin) {
+	var myopenaps_dir = "/root/myopenaps/";
```
scottleibrand: We'll need to figure out another way to do this.  Hard-coding the path won't work for some rigs, and makes the tests fail.

alimhassam:
You're right, i had the same concern, but pushed anyway to get comments. Any suggestion on how to fix this?

alimhassam:
Fixed all the comments, except the tests need to look into these.

scottleibrand:
I just tested this on one of our rigs and it didn't add anything to enact/bolused.json as I expected.  Are you running a newer version of nodejs to support fs.*Sync commands?  We haven't upgraded oref0 from 0.10.x yet.

alimhassam:
my node version is 
```
# node --version
v0.10.29
```

I don't see the messages either in `enacted/bolused.json`  I don't think it's meant to appear there. But i do see them in nightscout treatment reports or as a hover on the treatment pill.
The main feature is that i see the note in the pushover notifications from NS:

image

PR #771, opened 2017-11-04
Dev

mhaeberli:
revise oref0-setup.sh so that, in case openaps WAN is using bluetooth while bluetoothd is updated, wget happens before bluetoothd is killed

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -640,8 +640,8 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
         bluetoothdminversion=5.47
         bluetoothdversioncompare=$(awk 'BEGIN{ print "'$bluetoothdversion'"<"'$bluetoothdminversion'" }')
         if [ "$bluetoothdversioncompare" -eq 1 ]; then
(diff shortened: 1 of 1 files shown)


scottleibrand:
This might not be sufficient to allow oref0-setup to complete if started over an SSH session riding the Bluetooth link.  It might work if you do it within `screen` though.

Probably best to also advise in the docs that oref0-setup should be run over console or wifi.

mhaeberli:
Scott,
Thanks for accepting.
Agreed about breaking ssl at that point, but at least likely (maybe) bluez
update would be applied for next try after reboot?
Indeed, might work within screen / might not.
Also agree re docs change - I'll see if I can muster the chutzpah for a
minor docs change along those lines in the next few days.  by the way, ssl
over tethered bluetooth seems to work for most other oref0-setup stuff,
just updating the bluetooth in that context is hard...
Martin

scottleibrand:
Maybe you could just add an oref0-online style check of whether the user has wifi or Bluetooth connectivity, and bypass the bluez upgrade if they're on Bluetooth..

PR #773, opened 2017-11-04
reset pump clock if >55s off; better error messaging

scottleibrand:
Per https://gitter.im/nightscout/intend-to-bolus?at=59fd9d41b20c642429880228 we should try to reset the pump clock before it drifts a full 60s.  This also gets rid of an extra debug statement and replaces it with proper error messaging if the pump clock is mismatched.


scottleibrand:
Merged this to `optimize` to integrate with the better error handling there.  Will merge that to dev when it's ready to reduce merge conflicts.

Issue #736, opened 2017-10-21
Remove "Warning: API_SECRET is not set when calling oref0-autotune.sh" from Autotune


gooseodyssey:
Remove the warning `Warning: API_SECRET is not set when calling oref0-autotune.sh` from Autotune when running on a Mac.

PR #774, opened 2017-11-05
only display API_SECRET not set warning if autotune fails

scottleibrand:
Per #736, the current "Warning: API_SECRET is not set when calling oref0-autotune.sh" message is confusing, as it only needs to be set if someone has locked down read-only access to their NS using token auth.  This PR moves the warning line to only display if autotune fails, and to clarify that the warning is only a problem if you have locked down read-only access to your NS.

### DIFF
--- a/bin/oref0-autotune.sh
+++ b/bin/oref0-autotune.sh
@@ -27,8 +27,13 @@
 # THE SOFTWARE.
 
 die() {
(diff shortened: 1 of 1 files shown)


scottleibrand:
I think this is ready to merge to dev if no objections.

PR #775, opened 2017-11-05
Restore 512/712 support

scottleibrand:
512 and 712 pumps don't support SMB, so the switch to always using the smb-pump-loop broke them.  Per @dcacklam's https://github.com/openaps/oref0/pull/754, this special-cases x12 pumps back onto old_main for now, and updates one of the currenttemp safety checks that breaks on x12s to only run if microBolusAllowed (which is always set when running determine-basal from the main oref0-pump-loop).


scottleibrand:
I purposefully didn't bother with any of that, on the assumption that it would run main() and refresh the model.json before failing out.  Does that work, or does it fail before even getting the new model.json?

dcacklam:
It does call main & main does generate settings/model.json properly.
But the fail-out of main isn't exactly smooth (runs preflight repeatedly, as well as mmtune and associated waits/listens several times)...

Either way works, once you get the file created.

scottleibrand:
Ok.  I just added a die() to preflight to get it to short-circuit all of that and restart with old_main() when detecting an x12 pump.  Can you test that and see if it works any better?

scottleibrand:
@dcacklam anything else you want to test, or should we merge this to dev for further testing there?

PR #779, opened 2017-11-06
Fix oref0-autotune-recommends-report for DST

scottleibrand:
The way oref0-autotune-recommends-report currently builds its time_list array of H:M in 30 minute increments to mirror pump basal schedule doesn't work on the day that DST ends and the clock falls back: it goes into an infinite loop from 1am to 1:30am and back instead of proceeding to 2am.

This builds that array much more simply without using `date`,with two nested `for` loops instead, to bypass that problem.


jmcrawford:
Seems to have fixed the problem - the report runs to completion, whereas it hangs on an older install

Issue #781, opened 2017-11-07
Extended 0 temp


sjolundjohn:
I am having issues with the longer duration of 0.00 temps that are present in the Dev build. The problem I encounter is when I get a 2h extended 0.00 unit temp, then go out of range of my rig.

If I was in range, it would have corrected for it - but if I am out of range for a while, then I ended up going too high.

I would rather have 30min cycles OR the option to choose the max length.

scottleibrand:
Are you using SMB?  If so, I would recommend going back to AMA on non-portable rigs that you tend to walk away from.  The SMB zero temps are a safety feature, and folks who find that safety feature problematic are typically using SMB in a less-safe way.

sjolundjohn:
Yes I am on dev branch. It has been working super well and the SMB is a key reason for using it, but I do wish that it was less aggressive for the zero temp duration. I would think for portable rigs that refreshing every 30 minutes would have the same effect?

Feedback noted and I will consider. Thanks!

JoseeT1Bionique:
I have experienced this also on Oref0 a few times and feel it is also an issue.
I agree that temp basal should be set for 30 minutes maximum and the be recalculated or defaulted to pump basal, at all time and with all settings. I would make me feel safer to have this in Dev 6 that I will start testing Thursday night.  We can predict or sometimes monitor/trouble shoot connection loss and 2 hours 0 basal is way to long.

PieterGit:
@sjolundjohn @JoseeT1Bionique : 
With 0.6.x you can decrease the SMB's (and therefore the length of 0-temps) by decreasing `"maxSMBBasalMinutes"` to say 20 minutes. That way SMB's will be at most for 20 minutes of basal. The 0-temps will be shorter in that case as well. I used that when the CR/ISF values were not tuned (because I  went from multiple CR/ISF schedule to a single one).

PieterGit:
Decreasing `maxSMBBasalMinutes` will probably not work (a lot) to shorten the 0-temps a lot, if no pump-rig communication is possible. I would like to 👎  implementing a way of shortening the 0-temps. Having a safe algorithm is more important than avoiding a high bg in case the pump-rig communication fails.

If you have thoughts that oref0 has an extended 0-temp that takes too long (and the rig can communicate to the pump), please open an issue for that.

JoseeT1Bionique:
> Having a safe algorithm is more important than avoiding a high bg in case the pump-rig communication fails.

I agree.

[...]

PR #782, opened 2017-11-08
make maxZeroTempLength configurable via hidden preference

scottleibrand:
Lots of people complain about the 120m zero temps set when SMBing.  I think most of them are either misunderstanding how SMB works, or don't want to use its safety features as designed (and would prefer increased risk of post-meal lows for convenience).  This PR adds a hidden preference to placate them.

Changing maxZeroTempLength from 120m to 90m or even 60m won't make much difference, and is probably pretty safe.  I do not recommend setting it to 30m, as that is likely too short to prevent post-meal lows when carbs are overestimated and the pump walks away from the rig.

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -1077,8 +1077,12 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
                 durationReq = 0;
             // don't set a temp longer than 120 minutes
             } else if (durationReq >= 30) {
+                var maxZeroTempLength = 120
+                if (profile.maxZeroTempLength >= 30) {
+                    maxZeroTempLength = profile.maxZeroTempLength;
+                }
                 durationReq = round(durationReq/30)*30;
-                durationReq = Math.min(120,Math.max(0,durationReq));
+                durationReq = Math.min(maxZeroTempLength,Math.max(0,durationReq));
             } else {
                 // if SMB durationReq is less than 30m, set a nonzero low temp
                 smbLowTempReq = round( basal * durationReq/30 ,2);
--- a/lib/profile/index.js
+++ b/lib/profile/index.js
@@ -50,6 +50,7 @@ function defaults ( ) {
     , insulinPeakTime: 75 // number of minutes after a bolus activity peaks.  defaults to 55m for Fiasp if useCustomPeakTime: false
     , carbsReqThreshold: 1 // grams of carbsReq to trigger a pushover
     , offline_hotspot: false // enabled an offline-only local wifi hotspot if no Internet available
+    , maxZeroTempLength: 120 // maximum zero temp length to set for hypo safety, in minutes. >= 60m recommended
   };
   return profile;
 }


sulkaharo:
IMO there should be a core set of safety features for which there is no override, similar to the new minimum DIA. This is probably one of those settings. If someone wants to break OpenAPSs safety model, they can fork the software and break it. Again IMO it’s the dev team’s responsibility to make OpenAPS as safe and reliable as possible (while aiming at the best results possible); I don’t see how adding features that make Openaps less safe and adds code that needs maintaining is good for anyone. Please at least add unit tests for this if the setting goes in?

sulkaharo:
If we do merge this, should we consider adding a new global flag called "EnableDangerousPreferencesThatVoidRightForSupportAndMightKillTheUser" which needs to be set to true before "known unsafe" settings are considered, which would also add a "DANGER MODE" string to the Nightscout status & logs, so it'd be super clear if someone with loop issues is running with these settings?

scottleibrand:
To be fair, I don't think 30m SMB zero temps are quite *that* dangerous.  But some sort of "I understand that I'm taking on additional risk" setting might be good for things like this, and perhaps things like enableSMB_always too.

bdr1177:
It is my understanding that with SMB enabled, long zero temps may be set even if no SMBs are actually delivered, especially with very small basals (see https://gitter.im/nightscout/intend-to-bolus?at=59f89face44c43700a9e2e63).  It seems that this is a problem for those with very small basals because it is not infrequent that you end up with a long zero temp in the absence of SMBs.  I don't know about others complaining about this issue, but when I've complained about the long zero temps it wasn't based on a misunderstanding of the safety features.  It was based on the fact that the SMB safety features don't necessarily make sense to me in the absence of SMBs.

scottleibrand:
@bdr1177 yeah, that’s something I still want to fix. Your case wouldn’t really be helped by this preference, as it would just keep resetting a 30m zero temp every 10m. So this wasn’t talking about you or your issue. 

Can you open a new issue for the “zero temp but no SMB on tiny basals” thing so we don’t forget it?

bdr1177:
@scottleibrand Will do.

PieterGit:
I prefer not to include it, instead of creating a preference `EnableDangerousPreferencesThatVoidRightForSupportAndMightKillTheUser` (even if this has a really low chance). The reference design of OpenAps clearly says:  "it is designed primarily for safety, understandability". It's not safe and and it doesn't make it more understandable.

We already noticed common users are having troubles with all the configuration parameters, and they are often used in a wrong way without proper reading of the documentation. So we should minimize the configuration parameters and make it clear and understandable how they should be used.

If people want to add possibly unsafe options, they should fork, and they are responsible for their changes to the software.

Is this related to: https://github.com/openaps/oref0/issues/781#issuecomment-342349762 ?
I agree with @scottleibrand that if SMB's are used and rig is often offline, it's better to avoid SMB and stick to AMA. Perhaps we should make more clear in the docs that it's recommended with SMB's to have a stable/mobile rig, and in case of lot's of problems with rig-pump communication OpenAps is implemented such that safety is more important than preventing high bg's in those cases.

scottleibrand:
Yes, #781 is the most recent example of this, but we see it all the time, mostly in the Looped group on Facebook.  

I agree with the feedback so far.  I'll leave this issue open until someone makes the documentation updates that @PieterGit suggests, then close it without merging and focus instead on fixing issues like #785.

diabeticdude2:
Just a note, another example of this happened to me today.  The fault is my own but would have been fixed with an option of not setting temps longer than 30 minutes or even one hour.  I had lunch at 12:30 and my sensor went out at 2:00 PM.  So I change my pod and wait for the 2 hour calibration process.  What I failed to do was pay attention to the fact that my pump had also been suspended for a long time and would continue to be suspended for the whole 2 hour calibration process.  

After 2 hours I check my BG one reading was 252 the other was 236.  I understand that I failed by not checking my BG more during this process.  But I also understand that if my normal basal would have been in place I would have gotten an additional 2.6 units which would have kept me more in line.

scottleibrand:
Another option to fix this kind of thing would be to have determine-basal check if BG is more than X minutes old, and set a neutral temp (or cancel temp) if so.  Currently it only does that if you have a CGM plugged in and it's getting the error status code: missing data simply causes it to not do anything and let the currently running temp basal run to completion.

scottleibrand:
I see you just opened #788, so I'll move my response over there.

PR #783, opened 2017-11-08
set >30m zero temps when needed for low minGuardBG

scottleibrand:
Up until now, minGuardBG < threshold has always set 30m zero temps, even when more than 30m of zero temping would be required to bring BG back up to target.  This sets longer zero temps (30-120m) as needed in those situations.

I also realized that since we added minGuardBG, low glucose suspend mode can only be triggered if BG < threshold but BG is predicted to be >= threshold in 5 minutes.  Rather than continue to special-case that corner case, I set minGuardBG = bg initially (so minGuardBG should always be <= bg), and consolidated the old bg < threshold code into the minGuardBG < threshold case.


scottleibrand:

Issue #785, opened 2017-11-09
Zero temp but no SMB on tiny basals


bdr1177:
With tiny basals and SMB enabled, long zero temps may be set even if no SMBs are actually delivered (see discussion here: https://gitter.im/nightscout/intend-to-bolus?at=59f89face44c43700a9e2e63 and here: https://github.com/openaps/oref0/pull/782#issuecomment-342859374). This is a problem for those with very small basals because it is not infrequent that you end up with a long zero temp in the absence of any SMBs.

KensingtonB:
We are also having this issue.  Happy to help test if any proposed solutions, but no technical savvy to contribute ;)

scottleibrand:
Sure, you can test #786 (assuming you're comfortable with 0.6.0-dev).

scottleibrand:
Closing this now that #786 is merged.

PR #786, opened 2017-11-09
don't SMB zero temp if insulinReq > 0 but too low for a microBolus

scottleibrand:
When SMB mode is active, zero temps are set based on naiveEventualBG, and actual insulin requirements are normally covered by SMBs.  However, as per #785, there are situations, particularly with tiny basals, where insulinReq is positive, but is not high enough (0.2U) to trigger the minimum SMB of 0.1U.  In those situations, extended zero temping can occur before insulinReq gets high enough to trigger an SMB.

This PR skips the SMB zero-temp when insulinReq > 0 but microBolus < 0.1, and instead falls through to setting temp basals normally.  This should also help avoid zero temping without SMBs in cases where maxSMBBasalMinutes is too low and maxBolus < 0.1.

(This is somewhat similar to the old SMB bolus snooze behavior, which checked for `snoozeBG > target_bg && lastCOBpredBG > target_bg` instead, so I uncommented and reused that code.  I also removed a bunch of other old commented-out snoozeBG code and un-indented the remaining code block.)

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -895,77 +895,61 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
             }
         }
 
-        // if we've bolused recently, we can snooze until the bolus IOB decays (at double speed)
-        //if (snoozeBG > min_bg) { // if adding back in the bolus contribution BG would be above min
-            // If we're not in SMB mode with COB, or lastCOBpredBG > target_bg, bolus snooze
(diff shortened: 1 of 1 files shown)


scottleibrand:
To test: `cd ~/src/oref0 && git fetch && git checkout tiny-insulinReq && git pull && npm run global-install`.

KensingtonB:
Ok, should be able to test this weekend

bdr1177:
Will start testing.

bdr1177:
Been testing this for 3 days and haven’t seen any problems.

PR #790, opened 2017-11-12
only add extra negative deviations if exercise mode is set

scottleibrand:
Per https://gitter.im/nightscout/intend-to-bolus?at=5a08a2a471ad3f8736d66aa4 we should probably only be adding extra negative deviation if a high temptarget is running if exercise_mode or high_temptarget_raises_sensitivity is set.


scottleibrand:
To test: `cd ~/src/oref0 && git fetch && git checkout extra-neg-deviations && git pull && npm run global-install`

scottleibrand:

PR #792, opened 2017-11-12
gzip old autotune json and log files to save space

scottleibrand:
Without intervention, the autotune/ directory can grow indefinitely, as it keeps all old entries, treatments, and autotune log files.  To reduce the disk space impact of doing so, this PR gzips all of those old files before each run.

### DIFF
--- a/bin/oref0-autotune.sh
+++ b/bin/oref0-autotune.sh
@@ -178,7 +178,8 @@ done
 echo "Compressing old json and log files to save space..."
 gzip -f ns-*.json
 gzip -f autotune*.json
-gzip -f autotune.*.log
+# TODO: figure out how to do this without gzipping the log we're currently writing to
+# gzip -f autotune.*.log
 echo "Autotune disk usage:"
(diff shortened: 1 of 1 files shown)


tim2000s:
Very sensible given the amount of manual deleting (and then cron jobs) I’ve been running.

tim2000s:
Probably want to look at something similar on the openaps logs. I've been reviewing my disk space this evening (using ncdu - great little tool for this stuff, install with `sudo apt install ncdu` ) and I can see that `/var/log/openaps` is consuming 10% of the overall disk space, with roughly 50% made up of the recently rolled logs, 25% with the current logs and 25% on the old gzipped logs.

scottleibrand:
We already have logrotate in place to handle those.  If you have a rig with ongoing disk space issues from log pressure, we need help adjusting logrotate settings to constrain that.

tim2000s:
I'm just testing something on logrotate to see if I can improve it.

PR #793, opened 2017-11-12
 don't reduce insulinReq for expectedDelta (we have minGuardBG for that)

scottleibrand:
Even if BG is dropping when we enter carbs, we generally want SMB to be able to get insulin started quickly, unless minGuardBG predicts BG will drop too low.  This PR comments out the code that previously reduced insulinReq when minDelta was < 0 but > expectedDelta.


scottleibrand:
This is ready to merge if no objections.

PR #794, opened 2017-11-13
run autotune 4am-4am instead of midnight-midnight

scottleibrand:
When COB > 0 at midnight, it messes up autotuning of CR and post-midnight basals.  Since most people's schedules mean COB=0 at 4am a lot more often than at midnight, this PR changes autotune to run 4am-4am instead of midnight-midnight to mostly avoid those issues.

Anyone installing this branch will need to re-run oref0-setup and answer Y to deleting their old cron (or manually update the autotune cron line from 12:05am to 4:05am) or their autotune will stop tuning midnight-4am basals until they do so.

### DIFF
--- a/bin/oref0-autotune.sh
+++ b/bin/oref0-autotune.sh
@@ -181,7 +181,8 @@ echo "Grabbing NIGHTSCOUT treatments.json and entries/sgv.json for date range...
 # Get Nightscout BG (sgv.json) Entries
 for i in "${date_list[@]}"
 do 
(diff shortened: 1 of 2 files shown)


PieterGit:
From safety perspective it's probably even better to run it just before people wake up. If it changes basals it's better that the people are (near) awake. So I would even propose a 5am run.

scottleibrand:
5am risks missing breakfast, which would be bad for morning basals.

jdhigh:
I'll just put this here...
https://www.ted.com/talks/rives_on_4_a_m

JELCRAWFORD:
I think this is a solid change.

PieterGit:
@jdhigh can't argue with that link on the "explanation" of 4am. 👍  for 4am (but I will still be asleep 😄 )

alimhassam:
Installed this yesterday works well so far

Issue #788, opened 2017-11-10
cancel long zero temps if rig is not getting BG data


in-10-city:
Just a note,  this happened to me today.  The fault is my own but would have been fixed with an option of not setting temps longer than 30 minutes or even one hour.  I had lunch at 12:30 and my sensor went out at 2:00 PM.  So I change my pod and wait for the 2 hour calibration process.  What I failed to do was pay attention to the fact that my pump had also been suspended for a long time and would continue to be suspended for the whole 2 hour calibration process.  

After 2 hours I check my BG one reading was 252 the other was 236.  I understand that I failed by not checking my BG more during this process.  But I also understand that if my normal basal would have been in place I would have gotten an additional 2.6 units which would have kept me more in line.

scottleibrand:
One option to fix this kind of thing would be to have determine-basal check if BG is more than X minutes old, and set a neutral temp (or cancel temp) if so. Currently it only does that if you have a CGM plugged in and it's getting the error status code: missing data simply causes it to not do anything and let the currently running temp basal run to completion.

in-10-city:
Is there a way to have determine-basal look at the last trend and make a decision based on that?  

For example:  
If I am trending up at 5 points every 5 minutes for the last 35 minutes then. then lost communications for x minutes, have it cancel temp basal if the temp basal is set lower than my default or suspended.  

Or if I am trending down at 5 points every 5 minutes for the last 35 minutes then. then lost communications for x minutes allow it to keep the temp basal if it is suspended or set lower than my stated default.

scottleibrand:
Ok, I'm thinking we consolidate the "BG too old" and "CGM is calibrating or in ??? state" code into a single block of logic, which will:

Cancel high temp:
```
{"rate":0,"duration":0,"temp":"absolute","deliverAt":"2017-11-14T01:11:00.349Z","reason":"If current system time Mon Nov 13 2017 17:11:00 GMT-0800 (PST) is correct, then BG data is too old. The last BG data was read 12m ago at Mon Nov 13 2017 16:59:00 GMT-0800 (PST). Canceling high temp basal of 3.3"}
```
Shorten long zero temp to 30m:
```
[...]

PR #796, opened 2017-11-14
Cancel high temps and shorten long zero temps when we lose CGM data

scottleibrand:
Per https://github.com/openaps/oref0/issues/788 it would be better to cancel/shorten long zero temps when the rig stops receiving CGM data.  It would also make sense to do the same thing for CGM errors as for no data, as they're often the same condition, just with/without the CGM plugged in.  In the event of >10m old CGM data, or CGM errors, this PR will:

Cancel high temps:
{"rate":0,"duration":0,"temp":"absolute","deliverAt":"2017-11-14T01:11:00.349Z","reason":"If current system time Mon Nov 13 2017 17:11:00 GMT-0800 (PST) is correct, then BG data is too old. The last BG data was read 12m ago at Mon Nov 13 2017 16:59:00 GMT-0800 (PST). Canceling high temp basal of 3.3"}

Shorten long zero temps to 30m:
{"rate":0,"duration":30,"temp":"absolute","deliverAt":"2017-11-14T01:11:42.771Z","reason":"If current system time Mon Nov 13 2017 17:11:42 GMT-0800 (PST) is correct, then BG data is too old. The last BG data was read 12.7m ago at Mon Nov 13 2017 16:59:00 GMT-0800 (PST). Shortening 59m long zero temp to 30m. "}

and let low temps run:
{"reason":"If current system time Mon Nov 13 2017 17:09:46 GMT-0800 (PST) is correct, then BG data is too old. The last BG data was read 10.8m ago at Mon Nov 13 2017 16:59:00 GMT-0800 (PST). Temp 1.25 <= current basal 1.3U/hr; doing nothing. "}

### DIFF
--- a/bin/oref0-determine-basal.js
+++ b/bin/oref0-determine-basal.js
@@ -191,26 +191,6 @@ if (!module.parent) {
       process.exit(1);
     }
 
-    //if old reading from Dexcom do nothing
-
-    var systemTime = new Date();
-    var bgTime;
-    if (glucose_data[0].display_time) {
-        bgTime = new Date(glucose_data[0].display_time.replace('T', ' '));
-    } else if (glucose_data[0].dateString) {
-        bgTime = new Date(glucose_data[0].dateString);
-    } else { console.error("Could not determine last BG time"); }
-    var minAgo = (systemTime - bgTime) / 60 / 1000;
-
-    if (minAgo > 10 || minAgo < -5) { // Dexcom data is too old, or way in the future
-        var reason = "BG data is too old (it's probably this), or clock set incorrectly.  The last BG data was read at "+bgTime+" but your system time currently is "+systemTime;
-        console.error(reason);
-        var msg = {reason: reason }
-        console.log(JSON.stringify(msg));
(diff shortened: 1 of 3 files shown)


scottleibrand:
This looks good in my testing, so I think it's good to merge if no objections.

scottleibrand:
What is the downside to canceling or shortening temps if the MDT sensor data is delayed by more than 10m? That sounds like a problem we shouldn’t be hobbling the algorithm’s safety features to work around.

scottleibrand:
Opened https://github.com/openaps/oref0/issues/799 to address the issue @Bender1061 raised.  Until that gets addressed, I think canceling high temps and shortening zero temps when the MDT sensor gets >10m delayed is the safest course.

Any other comments on this PR?  I'd like to get it merged for 0.6.0 release if no objections.

danamlewis:
I'd say let's do 12 minutes, for 2 CGM data refreshes, since sometimes they're ~6 minutes. And let's resolve MDT via 799 separately. LGTM with 12 instead of 10.

PR #797, opened 2017-11-14
Reason improvements

alimhassam:
added some missing unit conversions, and added carb ration to the reason field

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -364,7 +364,7 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
                 console.error("enableSMB_with_temptarget not supported within 6h of using Bolus Wizard");
             }
         } else {
-            console.error("SMB enabled for temptarget of",target_bg);
+            console.error("SMB enabled for temptarget of",convert_bg(target_bg, profile));
             enableSMB=true;
         }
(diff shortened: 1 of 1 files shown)


scottleibrand:
This looks fine to me, but I won't be able to test it with mmol/L.  Maybe if you could just paste the resulting log lines from your pump-loop.log so I can make sure they look reasonable, that'd be good enough to merge to dev for wider testing there?

alimhassam:
sure, i'll post a screenshot of NS reason field, and a log excerpt

alimhassam:
image

scottleibrand:
Hmm, is convert_bg() adding newlines?  If so, that won't really work.  Or is that something custom you added to your NS?

alimhassam:
no, sorry, my version of NS is replacing ", " with "<br/>"

alimhassam:
and i just realised i put the wrong screenshot, i moved CR after ISF in the latest push

scottleibrand:
Ok, good.  I don't think we'll want to change the default OpenAPS pill behavior to match that, but it wouldn't hurt to start discussion around what we actually want it to look like.

alimhassam:
sure, we can discuss once i do the pr, i already find this more readable so added the quick hack, but there is for sure room for even better improvement as someone suggested in i-to-b forum.

PR #798, opened 2017-11-14
Deprecate and remove profile.carbratio_adjustmentratio

alimhassam:
This could have surprising or dangerous effects when used with autotune or smb.

### DIFF
--- a/lib/profile/carbs.js
+++ b/lib/profile/carbs.js
@@ -24,7 +24,7 @@ function carbRatioLookup (inputs, profile) {
             if (carbratio_data.units == "exchanges") {
                 carbRatio.ratio = 12 / carbRatio.ratio
             }
-            return carbRatio.ratio * profile.carbratio_adjustmentratio;
+            return carbRatio.ratio;
         } else {
             console.error("Error: Unsupported carb_ratio units " + carbratio_data.units);
             return;
--- a/lib/profile/index.js
+++ b/lib/profile/index.js
@@ -32,7 +32,6 @@ function defaults ( ) {
     , unsuspend_if_no_temp: false // if true, pump will un-suspend after a zero temp finishes
     , bolussnooze_dia_divisor: 2 // bolus snooze decays after 1/2 of DIA
     , min_5m_carbimpact: 8 // mg/dL per 5m (8 mg/dL/5m corresponds to 24g/hr at a CSF of 4 mg/dL/g (x/5*60/4))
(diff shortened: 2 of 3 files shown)


scottleibrand:
What's the use case for keeping carbratio_adjustmentratio at all?  In what situations would you want to set it > 1?

scottleibrand:
If no one is successfully using it on an ongoing basis, I'm inclined to deprecate/remove carbratio_adjustmentratio entirely.  It seems to be a knob people try, unsuccessfully, to use before moving on to doing things another way.

alimhassam:
I thought it might be good to keep it for safe measure, i just wanted to make it difficult to set it lower than 1 as it's now dangerous, specially since there are old documentation suggesting to do so and that it might only apply to the pump CR which is not true. 

This also adjusts the autotuned CR.

i don't personally needed right now, but it might possibly be a useful advanced tweaking point once i re-enable autotune in case it suggest CR that are still too low for any reason.

alimhassam:
thought about this a little more, I guess you're right that there are too many parameters, so it if you prefer i'll remove it, and if needed again it could easily be re-added, maybe with a different name somethnig like `autotune_carbratio_multiplier` to make it clear.

alimhassam:
i removed the parameter completely have the code on my rig, but didn't test a re-install

alimhassam:
oops, didn't mean to close it, was supposed to just be a comment.

Bender1061:
I use this, I have it set to 1.2 making it more aggressive with meals. tends to do a great job for me.

scottleibrand:
That sounds backwards. 1.2 makes what more aggressive? Also, are you running autotune?

scottleibrand:
All:  If you've used carbratio_adjustmentratio in the past, or are using it currently, I'd like to know how you used / are using it, and what your experience has been with it. If anyone is using it successfully I'd like to understand exactly how you're using it so we can decide whether to keep it, and if so how to make it safer for new users and those using autotune. And if you used it but gave up on it, we'd also like to understand why you tried it originally and why you stopped using it.

Bender1061:
I started using this when I was having issues with SMB, this helped me at least during that transition. But I've for the past week set it as 1 so yeah it's no longer needed for me, so I'm OK with this.

scottleibrand:
Ok.  If no one else thinks this needs to be kept, I'm planning to merge this PR as-is.

PR #803, opened 2017-11-16
only try to get bnep0 IP if bt-pan succeeds

scottleibrand:
Our rigs occasionally experience a kernel bug panic after attempting to renew DHCP on bnep0, usually in flight, and have to be rebooted. This limits dhclient attempts to only occur when bt-pan runs successfully.


scottleibrand:
This looks good to merge if no objections.

dramageek:
I'll give this a shot and see if it helps my rig. 
 I'd already bumped my cron to disable the hotspot every 20 minutes, too!

PR #804, opened 2017-11-16
actually shorten zero temps

scottleibrand:
oref0-pump-loop was saying `No smb_enact needed.` when oref0-determine-basal tried to shorten a temp.  This PR changes it to make sure that the running temp is no more than 20 minutes longer than the suggested one.


scottleibrand:
This looks good to merge if no objections.

PR #806, opened 2017-11-20
retry bad_wifi after an hour

scottleibrand:
If a wifi network is marked bad and we're still connected via BT, retry the wifi after an hour to see if the issue was transient.

Also, bash-completion for easier TAB TAB.


scottleibrand:
Tested successfully; ready to merge if no objections.

PR #810, opened 2017-11-21
Process the Meal marker event from medtronic pump

cjo20:
Needs some testing (and cleanup) before it's merged - probably wait until after 0.6.0


scottleibrand:
Do you still want to do cleanup on this, or is it ready for review?

cjo20:
I’ve been using it in its current form for a few days with no problems. The only thing I think needs addressing is how it interacts with nsCarbs and bwCarbs, currently it isn’t being counted as either, just totalcarbs. Is that a problem?

PR #812, opened 2017-11-22
require boolean true on enableSMB preferences

scottleibrand:
Per https://gitter.im/nightscout/intend-to-bolus?at=5a15a1a2df09362e673ea448 @alimhassam discovered that setting a preference to the string `"false"` (instead of the boolean `false`) evaluates as `true`, which unexpectedly gave him enabledSMB_always behavior.  This PR looks for a boolean `true` (not just something that evaluates to true) for the enableSMB preferences to make things behave as you'd intuitively expect.

There may be other boolean preferences that need this treatment as well, but the enableSMB ones are the most important.

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -343,12 +343,12 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     var enableSMB=false;
     // disable SMB when a high temptarget is set
     if (! microBolusAllowed) {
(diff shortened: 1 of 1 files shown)


alimhassam:
tested, this correctly disabed SMB with  "enableSMB_always": "false", note that it also keeps it disabled when  "enableSMB_always": "true", it really needs to be set as  "enableSMB_always": true to enable.

scottleibrand:
Yes, that’s how I intended it to work.

PR #817, opened 2017-11-26
Simple Remote Access Scripts Test

Ebgineer:
regenerate-index.sh creates a new file index.html every time it is run. This contains simple html formatting with example data for now (intent is to add additional parameter-value pairs for display after approval of proof of concept).
setup-http.sh is intended to be called from oref0-setup.sh and adds cron entries to start SimpleHTTPServer at reboot and run regenerate-index.sh periodically. Both files should be in ~/myopenaps/enact directory. From a browser on any device, browse to http://[RIG_IP_ADDRESS]/index.html and the page will self-refresh periodically.

### DIFF
--- a/www/regenerate-index.sh
+++ b/www/regenerate-index.sh
@@ -1,0 +1,23 @@
+rm ~/myopenaps/enact/index.html
+(echo '<meta http-equiv="refresh" content="60">') > ~/myopenaps/enact/index.html
+(echo '<html>') >> ~/myopenaps/enact/index.html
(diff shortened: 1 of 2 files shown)


scottleibrand:
Should we put the index.html into the www/ folder instead of enact/ along with the other web stuff we've done?

tim2000s:
I would do. The old solution of running from enact was simply a work around. It’s better by far to have any web services running from their own sandboxed space with a user that has no ability to read and write to the APS file system (only to the web server one).  Especially if someone can interact with them.

PR #818, opened 2017-11-27
use pumpbasalprofile for autotune IOB calculations

scottleibrand:
When running autotune retrospectively against NS data containing many/long periods without temp basals (i.e. data from AndroidAPS, OpenAPS with profile.skip_neutral_temps = true, Loop with wide BG target ranges, or long periods without any closed looping), autotune would assume that the autotuned basals were in effect any time there was no temp basal.  That could result in a feedback loop where autotune continued trying to raise or lower basals indefinitely until it hit the autosens_max or autosens_min, rather than converging on a sensible basal schedule.  This PR changes autotune to assume that the basal schedule from pumpprofile.json was in effect instead.  That should ensure that autotune converges, and as long as the pumpprofile.json basal schedule matches what was actually in effect during the time period being analyzed, should result in much better basal recommendations for time periods without neutral temps.

### DIFF
--- a/bin/oref0-autotune-prep.js
+++ b/bin/oref0-autotune-prep.js
@@ -97,6 +97,7 @@ if (!module.parent) {
     var inputs = {
         history: pumphistory_data
     , profile: profile_data
+    , pumpprofile: pumpprofile_data
     , carbs: carb_data
     , glucose: glucose_data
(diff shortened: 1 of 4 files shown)


scottleibrand:
I think this is ready to merge to dev if no objections.

PR #819, opened 2017-11-27
Basal record search

cjo20:
Search more than 1 record ahead to find a duration to match a basal rate change. Only search to the end of the array, or until the record date changes.

Don't search using `date` to validate the records match, it isn't defined in those records. Instead use `timestamp`.


scottleibrand:
I believe the temp basal and temp basal duration records can be in either order.  It looks like your iter search only searches one direction through the array, so it might miss that...

lib/iob/history.js
```diff
@@ -148,12 +148,26 @@ function calcTempTreatments (inputs, zeroTempDuration) {
...
-                var duration = pumpHistory[i+1]['duration (min)'];
-            } else { console.error("No duration found for "+rate+" U/hr basal"+date, pumpHistory[i - 1], current, pumpHistory[i + 1]); }
+            var timestamp = current.timestamp;
+	    var duration;
+            if (i>0 && pumpHistory[i-1].timestamp == timestamp && pumpHistory[i-1]._type == "TempBasalDuration") {
+                duration = pumpHistory[i-1]['duration (min)'];
+            } else { 
+		var iter = i+1;
+
+		while (iter < pumpHistory.length && pumpHistory[iter].timestamp == timestamp) {
```
scottleibrand: Would it be more straightforward to just make this a for loop over the entire array?  Something like `for (var iter=0; iter < pumpHistory.length; iter++) {` maybe?

cjo20: Yeah, can do that. Any idea why it wasn't done like that in the first place?

scottleibrand: Efficiency.  But if we're checking the adjacent records first, this is just a fallback for a very small percentage of records that don't match those, so I'm not really concerned about performance here.

PR #820, opened 2017-11-27
Propagate the noise field.

alimhassam:
Propagate the noise field of glucose data so it can be used by determine_basal.

### DIFF
--- a/lib/glucose-get-last.js
+++ b/lib/glucose-get-last.js
@@ -65,6 +65,7 @@ var getLastGlucose = function (data) {
     return {
         delta: Math.round( last_delta * 100 ) / 100
         , glucose: Math.round( now.glucose * 100 ) / 100
+        , noise: Math.round(now.noise)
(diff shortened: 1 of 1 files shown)


alimhassam:
without this, the new noise logic doesn't work, see comments here: https://github.com/openaps/oref0/pull/735#issuecomment-347035387

lib/glucose-get-last.js
```diff
@@ -65,6 +65,7 @@ var getLastGlucose = function (data) {
     return {
         delta: Math.round( last_delta * 100 ) / 100
         , glucose: Math.round( now.glucose * 100 ) / 100
+        , noise: now.noise
```
scottleibrand: Should we round it to the appropriate precision to make sure we don't have floating point math fun like 3.00000000000001?

alimhassam: i guess we could, probably wouldn't hurt, but at the moment, i only saw integer values for this field, but let me fix that.

alimhassam: fixed

PR #822, opened 2017-11-27
Make oref0-version more useful if you're not running on openaps/oref0 directly

cjo20:
A few changes to make the message at the end of the loop more useful if you're not running on the main `openaps/oref0 repo`.
- Check OREF0_SRC env for source code location.
- Check for being behind upstream (if it exists).
- Only warn if repo is ahead of local (not local ahead of remote)

### DIFF
--- a/bin/oref0-version.sh
+++ b/bin/oref0-version.sh
@@ -1,13 +1,13 @@
 #!/bin/bash
-
+source ~/.bash_profile
 # Simple script to check current version / branch of oref0 installed and check for updates
-
-branch=$(cd $HOME/src/oref0/ && git rev-parse --abbrev-ref HEAD)
-version=$(jq .version "$HOME/src/oref0/package.json" | sed 's/"//g')
+location=${OREF0_SRC:-${HOME}/src/oref0}
+branch=$(cd ${location} && git rev-parse --abbrev-ref HEAD)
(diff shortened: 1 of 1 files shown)


scottleibrand:
What does the duplicated code do?  Can we just add that check to the code above and un-duplicate it?

cjo20:
When you're not running on the `openaps/oref0` repository directly (say, a personal fork), the easiest (possibly only?) way to have access to the personal repository and the base `openaps/oref0` I've found is to have it set up like this:

```
$ git remote -v
origin  https://github.com/cjo20/oref0 (fetch)
origin  https://github.com/cjo20/oref0 (push)
upstream        http://github.com/openaps/oref0.git (fetch)
upstream        http://github.com/openaps/oref0.git (push)
```
The first instance of the code checks how far behind the `origin` head you are, the second instance checks how far behind the `upstream` head you are.

If I put a bunch of changes to `cjo20/oref0` from somewhere other than my rig, and you push a bunch of changes to `openaps/oref0` this PR will give me a notification on my rig that I'm behind both my head and yours.

```
Your instance of oref0 [0.6.1-dev, dev] is up-to-date.
Your instance of oref0 [0.6.1-dev, dev] is up-to-date with upstream.
```

PR #824, opened 2017-11-27
ns-status js format

kkota1:
changed splice/pop to array element
cleaned up ns-status

### DIFF
--- a/bin/ns-status.js
+++ b/bin/ns-status.js
@@ -3,10 +3,9 @@
 
 var os = require("os");
 
-var requireUtils = require('../lib/require-utils')
-  , safeRequire = requireUtils.safeRequire
-  , requireWithTimestamp = requireUtils.requireWithTimestamp
-  ;
+var requireUtils = require('../lib/require-utils'),
+    safeRequire = requireUtils.safeRequire,
+    requireWithTimestamp = requireUtils.requireWithTimestamp;
 
 /*
   Prepare Status info to for upload to Nightscout
@@ -28,7 +27,7 @@ function mmtuneStatus (status) {
     if (mmtune_input) {
         var mmtune = requireWithTimestamp(cwd + '/' + mmtune_input);
         if (mmtune) {
-            if (mmtune.scanDetails && mmtune.scanDetails.length > 0) {
+            if (mmtune.scanDetails && mmtune.scanDetails.length) {
                 mmtune.scanDetails = mmtune.scanDetails.filter(function (d) {
                     return d[2] > -99;
                 });
--- a/bin/ns-status.js
+++ b/bin/ns-status.js
@@ -3,9 +3,9 @@
 
 var os = require("os");
 
-var requireUtils = require('../lib/require-utils'),
-    safeRequire = requireUtils.safeRequire,
-    requireWithTimestamp = requireUtils.requireWithTimestamp;
+var requireUtils = require('../lib/require-utils');
+var safeRequire = requireUtils.safeRequire;
+var requireWithTimestamp = requireUtils.requireWithTimestamp;
 
 /*
   Prepare Status info to for upload to Nightscout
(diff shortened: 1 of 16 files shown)


bin/ns-status.js
```diff
@@ -3,10 +3,9 @@
 
 var os = require("os");
 
-var requireUtils = require('../lib/require-utils')
-  , safeRequire = requireUtils.safeRequire
-  , requireWithTimestamp = requireUtils.requireWithTimestamp
-  ;
```
scottleibrand: Putting the comma at the beginning of each new line (and the ; on its own line afterward) makes diffs cleaner when you add new variables to the list.

If you'd rather have a different `var foo = bar ;` statement instead of a comma-separated list, that would also work.

bin/ns-status.js
```diff
@@ -25,30 +24,26 @@ var requireUtils = require('../lib/require-utils')
...
-        if (mmtune) {
-            if (mmtune.scanDetails && mmtune.scanDetails.length > 0) {
-                mmtune.scanDetails = mmtune.scanDetails.filter(function (d) {
-                    return d[2] > -99;
-                });
-            }
-          status.mmtune = mmtune;
+    var mmtune = requireWithTimestamp(cwd + mmtune_input);
+    if (mmtune) {
+        if (mmtune.scanDetails && mmtune.scanDetails.length) {
```
scottleibrand: Why are you removing the `> 0` here?

bin/ns-status.js
```diff
@@ -65,30 +60,29 @@ if (!module.parent) {
...
+        inputs = params._,
+        clock_input = inputs[0],
+        iob_input = inputs[1],
+        suggested_input = inputs[2],
+        enacted_input = inputs[3],
+        battery_input = inputs[4],
+        reservoir_input = inputs[5],
+        status_input = inputs[6],
+        mmtune_input = inputs[7],
+        uploader_input = params.uploader;
```
scottleibrand: What's the advantage of making things into a comma-separated list vs. having a separate var for each line?

bin/ns-status.js
```diff
@@ -98,12 +92,12 @@ if (!module.parent) {
...
-        var iob = null
-          , iobArray = requireWithTimestamp(cwd + '/' + iob_input)
-          , suggested = requireWithTimestamp(cwd + '/' + suggested_input)
-          , enacted = requireWithTimestamp(cwd + '/' + enacted_input);
+        var iob = null,
+            iobArray = requireWithTimestamp(cwd + iob_input),
+            suggested = requireWithTimestamp(cwd + suggested_input),
+            enacted = requireWithTimestamp(cwd + enacted_input);
 
-        if (iobArray && iobArray.length && iobArray.length > 0) {
```
scottleibrand: Why are you removing the `> 0` check?

bin/ns-status.js
kkota1: OK will revert in a bit

bin/ns-status.js
kkota1: It's redundant. If length is truthy, then it's value must be greater than 0

bin/ns-status.js
kkota1: Virtually none performance-wise, but I think it's more visually appealing if you're assigning a bunch of vars

bin/ns-status.js
kkota1: see other response

bin/ns-status.js
scottleibrand: Ok, I'd rather optimize for minimizing errors and clarity of diffs than visual appeal.  I don't care if we standardize on lots of `var =` statements or putting commas on following lines, but have had enough typos due to copying and pasting lines without adjusting commas that I'd rather not go with a comma-separated list with the commas and semicolon at the end of the line.

scottleibrand:
And thanks for simplifying the slice() stuff.  Assuming that works identically, your syntax is much easier to work with.

bin/ns-status.js
kkota1: Fixed. Regarding errors, a potential risk with var assignments I've seen in a number of places is declaring vars mid block instead of at the top. This can lead to errors if devs don't know how hoisting works

https://www.w3schools.com/js/js_hoisting.asp

kkota1:
I haven't done any testing

kkota1:
It's unlikely that I'll be able to test these. I haven't run the app myself and possibly don't have the hardware for it. Was just looking through out of interest for some family members and ended up doing this out of boredom

I can close the PR or wait for someone else to test. Just let me know. Sorry about that

scottleibrand:
Ok, no problem at all.  These look like good changes: let's keep it open and we'll test it before merging.  Thanks!

scottleibrand:
Tested successfully.  Thanks for the contribution!

PR #826, opened 2017-11-28
Improve the oref0-pump-loop debuging options by adding a debug flag.

alimhassam:
I took another stab at adding a verbose logging or debug option to oref0-pump-loop.sh if the approach is acceptable it could be done to other scripts such as ns-loop, i got tired of changing some of these manually every time, I hope it's done in a useful way, in this case the existing behavior is not changed unless an environment variable is passed.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
 
-# OREF0_DEBUG make this script much more verbose
-# and allow it to print additional debug information.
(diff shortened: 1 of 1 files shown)


alimhassam:
thanks, let me verify that it all works in the 3 mode, and maybe do it on other scripts too then.

alimhassam:
ok, tested locally without the variable set, and with values 1 or 2, does what's expected.

scottleibrand:
Cool.  Ready to merge then if no objections.

PR #828, opened 2017-11-28
commenting out unused and incorrect prepped_glucose code

scottleibrand:


scottleibrand:
Also includes a couple of other minor things from scott-dana-dev that never made it to dev.

PR #829, opened 2017-11-28
only calculate UAM deviations for the last ~30m

scottleibrand:
When BG is rising fast and then slows its rise, it currently takes ~45m for UAM to pick itself up off the floor and predict continued rise.  This shortens the UAM lookback to ~30m to make it more responsive.


scottleibrand:
If no objections, I think this is ready to merge to dev for testing there.

PR #831, opened 2017-11-28
don't require enableSMB to enableUAM

scottleibrand:
Now that enableUAM is on by default, we should probably make use of it in non-SMB contexts as well, to help AMA determine when high-temping is needed due to undercounted carbs, etc.


scottleibrand:
Unless anyone wants to test this beforehand, I'll go ahead and merge this to dev for testing there.

PR #836, opened 2017-11-29
only use splitHistory internally: return un-split tempHistory

scottleibrand:
Due to the way temp basal splitting works, the temp basal verification safety checks sometimes erroneously determine that a running temp basal doesn't match what's in pumphistory.  This PR continues to make use of the splitHistory internally, but returns the un-split tempHistory records for use in verification checks.


scottleibrand:
This looks good to me, so ready to merge it if no objections.

PR #837, opened 2017-11-29
fix Edison batteryVoltage

scottleibrand:
Since we removed the edisonbattery.json import for non-MDT and non-xDrip users, newly installed rigs may not properly refresh the monitor/edison-battery.json file or properly check for low voltage to shut down the rig if oref0-ns-loop can't talk to NS and ns_meal_carbs fails.  This PR fixes oref0-pump-loop and crontab to not rely on the alias.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -702,6 +702,7 @@ function glucose-fresh {
 }
 
 function refresh_pumphistory_24h {
+    sudo ~/src/EdisonVoltage/voltage json batteryVoltage battery > monitor/edison-battery.json 2>&3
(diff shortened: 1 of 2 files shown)


scottleibrand:
Actually, looks like this wasn't broken, because oref0-ns-loop refreshes it (when it can successfully upload): https://github.com/openaps/oref0/blob/master/bin/oref0-ns-loop.sh#L118

This PR is still needed for cases where the rig is offline and oref0-ns-loop terminates early, though.

scottleibrand:
This one just fixes a corner case, so probably best to merge it to dev for testing there if no objections.

PR #838, opened 2017-11-29
comment out broken CURL_AUTH API_SECRET line

scottleibrand:
When tokenauth was added, an improperly escaped CURL_AUTH= line was also added.  It hasn't done much harm, since anyone not using tokenauth has their NS site world-readable, but in @kandhoff's case (running autotune on a Mac) it seems to be breaking something: https://gitter.im/openaps/autotune?at=5a1e307c540c78242d5f36c3

This PR simply comments out the non-functional line to see if that helps.


scottleibrand:
To test: `cd ~/src/oref0 && git fetch && git checkout ns-get && git pull && npm run global-install`

scottleibrand:

PR #839, opened 2017-11-29
Automate the steps to install flask as a service

cduarte:
Automate the steps to install flask as a service so can monitoring your rig OFFLINE with an Offline web page

IPHONE USERS To access this from an iphone browser, enter something like the following: http://172.20.10:5000

ex : http://IP_address__rig:5000

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -848,12 +848,13 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
         done
     fi
 
+    sudo pip install flask || die "Can't add xdrip cgm - error installing flask"
+    sudo pip install flask-restful || die "Can't add xdrip cgm - error installing flask-restful"
+
     # xdrip CGM (xDripAPS)
     if [[ ${CGM,,} =~ "xdrip" ]]; then
         echo xdrip selected as CGM, so configuring xDripAPS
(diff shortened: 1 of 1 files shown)


scottleibrand:
LGTM.  Will merge, probably tomorrow, if no objections.

efidoman:
should we specify -p 5000 in the crontab command line for flask? I tested this manually and it only worked if I specify -p and the port number when running flask.

efidoman:
Also, any objections to adding a 60 second refresh into index.html? Use case is when I'm in a hotel, no wifi and just want to leave my phone or laptop on this page all night and glance at it from the bed to see BG.

scottleibrand:
Feel free to PR either/both of those.

scottleibrand:
Merging this to a new `flask` branch for easier PR targeting.

PieterGit:
Two remarks:

1. I think users should be asked if they want to have a flask webserver installed on their rig.  Having a local webserver running on the rig should be optional.  

2. Running `/root/myopenaps/enact/regenerate-index.sh` every minute seems a waste of cpu resources, why not generate it upon user request?

scottleibrand:
Feel free to PR improvements.  I'm generally going to favor merging people's work to dev rather than holding it up for hypothetical further changes, but would have no objection to merging further work to make it more secure and user-friendly.

PR #841, opened 2017-11-29
always enable autosens-loop

scottleibrand:
Now that oref0-pump-loop is no longer using `openaps report invoke enact/suggested.json`, disabling autosens in oref0-setup doesn't actually disable its use, just prevents it from being updated, which can result in the sensitivityRatio getting "stuck" on whatever the last value was when it was disabled.

To simplify everything, this PR removes the interview question about autosens and always enables the autosens-loop.  If someone really needs to prevent autosens adjustments while they troubleshoot an issue, they can set their autosens_min and autosens_max in preferences.json to 1.

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -143,7 +143,7 @@ if ! [[ ${CGM,,} =~ "g4-upload" || ${CGM,,} =~ "g5" || ${CGM,,} =~ "g5-upload" |
     DIR="" # to force a Usage prompt
 fi
 if [[ -z "$DIR" || -z "$serial" ]]; then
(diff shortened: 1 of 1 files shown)


scottleibrand:
I think this is good to merge if no objections.

drnoname82:
Do I understand correctly that after this change it will be impossible to run OpenAPS without autosens? How quickly will a rig without autotune factor in a change in pump settings? I think there are still (rare) situations where autosens can impossibly be fast enough and only direct changes of the pump settings provide a base for somewhat useful predictions.

scottleibrand:
IMO changing pump settings is almost never the best way to respond to short-term fluctuations in insulin needs. When you (or autotune) change pump settings, autosens recalculates sensitivityRatio based on the new settings. If the data doesn’t indicate the need for the change you just made, autosens will at least partially counteract it until the sensitivity or resistance shows up in the data. 

I would instead enable the preferences to indicate sensitivity and resistance with high and low temp targets, and then use those to override autosens as needed. 

If you really need to disable autosens, you can disable it (and force autotune to return to pump settings on its next run) by setting autosens_max and autosens_min both to 1 in preferences.json.

Issue #840, opened 2017-11-29
0.6.0: "skip_neutral_temps": true fails a safety check.


davidwagaps:
Per @scottleibrand There is a safety check that doesn’t work right with skip_neutral_temps
Because its corrective action is to set a neutral temp.

skip_neutral_temps is helpful to minimize beeps/vibrations which has been known to wake up pump users or disturb their classmates & teachers (said pump users will remain anonymous of course).

Possible solution is to deprecate that setting and instead cancel the temp just before the top of the hour and then restart it just after.

scottleibrand:
Please test https://github.com/openaps/oref0/pull/842 and see if it fixes this.

davidwagaps:
Checks:
1. Does NOT cancel zero temp at :55 - pass
2. Otherwise cancels temp at :55 - pass
3. Resumes normal operation at :00 (sets temp if necessary sometime after
that) - pass
4. Does not set temp if temp = current pump basal setting (original "skip"
functionality) - pass
5. "reason":"Warning: currenttemp rate 0 != lastTemp rate 2.35 from
pumphistory; " - no longer appears in pill or log: pass

Looks good. Thanks so much!

danamlewis:
Resolved via https://github.com/openaps/oref0/pull/842, so closing

PR #842, opened 2017-11-29
Cancel temp on error and > 55m w/ skip_netural_temps

scottleibrand:
Per #840, skip_netural_temps currently breaks the temp basal safety checks, because the neutral temp we attempt to set to fix the situation never gets set.

This PR fixes that by issuing a cancel temp when the safety checks fail, and making sure that such a cancel temp command actually gets passed through to the pump.

This also includes a new feature to cancel temp basals (unless they're low glucose suspend zero-temps) after 55 minutes past the hour if skip_neutral_temps is set, so that no temp basal will be running at the top of the hour and therefore the pump will not beep or vibrate.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -204,10 +204,10 @@ function determine_basal {
     oref0-determine-basal monitor/iob.json monitor/temp_basal.json monitor/glucose.json settings/profile.json settings/autosens.json monitor/meal.json --microbolus --reservoir monitor/reservoir.json > enact/smb-suggested.json
 }
 
-# enact the appropriate temp before SMB'ing, (only if smb_verify_enacted fails)
+# enact the appropriate temp before SMB'ing, (only if smb_verify_enacted fails or a 0 duration temp is requested)
 function smb_enact_temp {
     smb_suggest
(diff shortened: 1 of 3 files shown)


scottleibrand:
This needs testing both by someone using skip_neutral_temps and someone without it.  To test: `cd ~/src/oref0 && git fetch && git checkout set-temp && git pull && npm run global-install`

davidwagaps:
Sorry, my notification to test was filtered out of my inbox so I just saw it. 
Question: Will the temp cancelled at :55 resume after the top of the hour?

scottleibrand:
It won't necessarily resume the same temp basal that was running, as that would mean extending it to 30m, but it will set a new temp basal if needed after :00

davidwagaps:
OK, so it's "cancel temp basal at :55 and prevent setting temp until :01"
at which point system will determine if basal is needed (like at any other
time).

davidwagaps:
Started testing yesterday evening. So far so good. I'll post more once it's been 24hrs.

davidwagaps:
1. Does NOT cancel zero temp at :55 - pass
2. Otherwise cancels temp at :55 - pass
3. Resumes normal operation at :00 (sets temp if necessary sometime after
that) - pass
4. Does not set temp if temp = current pump basal setting (original "skip"
functionality) - pass
5. "reason":"Warning: currenttemp rate 0 != lastTemp rate 2.35 from
pumphistory; " - no longer appears in pill or log: pass
(Also posted in https://github.com/openaps/oref0/issues/840)

PR #843, opened 2017-11-29
Serve offline web page with flask

scottleibrand:
This merges #839 to dev, along with some minor tweaks as discussed there.

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -1060,7 +1060,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
             (crontab -l; crontab -l | grep -q "oref0-pushover" || echo "* * * * * cd $directory && oref0-pushover $PUSHOVER_TOKEN $PUSHOVER_USER 2>&1 >> /var/log/openaps/pushover.log" ) | crontab -
         fi
         (crontab -l; crontab -l | grep -q "cd $directory && oref0-version --check-for-updates" || echo "0 * * * * cd $directory && oref0-version --check-for-updates > /tmp/oref0-updates.txt") | crontab -
-        (crontab -l; crontab -l | grep -q "flask run" || echo "@reboot cd ~/src/oref0/www && export FLASK_APP=app.py && flask run -p 8080 --host=0.0.0.0") | crontab -
+        (crontab -l; crontab -l | grep -q "flask run" || echo "@reboot cd ~/src/oref0/www && export FLASK_APP=app.py && flask run -p 8080 --host=0.0.0.0" | tee -a /var/log/openaps/flask.log) | crontab -
 
         crontab -l | tee $HOME/crontab.txt
     fi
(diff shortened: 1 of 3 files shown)


scottleibrand:
Anything else before I merge this?

efidoman:
It looks great to me. Glad to see the -p port being specified and the refresh interval of 30 seconds. In my testing, I noticed that every once in a while, the json.load(open (... xxx/xxx.json)) code will get a ValueError exception. I think this is when the race condition occurs that the file is being overwritten by openaps. Sometimes, openaps actually deletes the file and then rebuilds it. Just a split second, but it can cause the exception to occur.  I suggest considering putting exception handling in to handle a ValueError. This is what I did ... ```   try:
        data['hostname']=socket.gethostname()
        glucose = json.load(open(os.path.join(myopenaps_dir, "monitor/glucose.json")))
        data['glucose']=glucose[0]
        # TODO: calculate delta properly when glucose[1] isn't 5m ago
        delta=glucose[0]['glucose']-glucose[1]['glucose']
        tick=""
        if delta >= 0:
          tick += "+"
        tick += str(delta)
        data['tick']=tick

        iob = json.load(open(os.path.join(myopenaps_dir, "monitor/iob.json")))

        data['iob'] = iob[0]

        data['meal'] = json.load(open(os.path.join(myopenaps_dir, "monitor/meal.json")))

        data['suggested'] = json.load(open(os.path.join(myopenaps_dir, "enact/suggested.json")))

        data['enacted'] = json.load(open(os.path.join(myopenaps_dir, "enact/enacted.json")))

        data['temp_basal'] = json.load(open(os.path.join(myopenaps_dir, "monitor/temp_basal.json")))
    except ValueError:
        return render_template('indexError.html', data=data )
    else:
        # print(data)
        return render_template('index.html', data=data )
```
Obviously, I created an indexError.html as follows. It will refresh back in 30 seconds to the right page, or hit F5 and its back ...

indexError.html
```
<html>
<head>
<meta http-equiv="refresh" content="20" />
<link rel="stylesheet" href='/static/style.css' />
</head>
<body>
<h1>OpenAPS</h1>
<p>Temporary JSON Value Error. Wait 20 seconds or hit refresh - F5 </p>
</body>
</html>
```

**Comment by scottleibrand** [comment]
Please PR improvements rather than just commenting.

File: bin/oref0-setup.sh [hunk 1]
```diff
@@ -1059,6 +1060,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
             (crontab -l; crontab -l | grep -q "oref0-pushover" || echo "* * * * * cd $directory && oref0-pushover $PUSHOVER_TOKEN $PUSHOVER_USER 2>&1 >> /var/log/openaps/pushover.log" ) | crontab -
         fi
         (crontab -l; crontab -l | grep -q "cd $directory && oref0-version --check-for-updates" || echo "0 * * * * cd $directory && oref0-version --check-for-updates > /tmp/oref0-updates.txt") | crontab -
+        (crontab -l; crontab -l | grep -q "flask run" || echo "@reboot cd ~/src/oref0/www && export FLASK_APP=app.py && flask run -p 8080 --host=0.0.0.0") | crontab -
```
PieterGit: Running this service should be optional and disabled by default. There are a lot of other ways of getting information about the rig status (e.g. Nightscout, Nightscout apps, Watchfaces, etc.), so not all users would like to run a webserver on their rig.

scottleibrand:
In the absence of any PRs to the `flask` branch in the next day or so, I'm inclined to merge this to dev and target further PRs there.

PR #844, opened 2017-11-29
limit SMB zero temps to 60m

scottleibrand:
Per https://github.com/openaps/oref0/issues/781 and other feedback, and my own experience, it seems that 120m SMB zero temps may be doing more harm than good.

This PR caps SMB zero temps to 60m, while allowing zero temps based on minGuardBG or eventualBG to still extend to 120 minutes.  This should minimize the situations in which really long zero temps are set where unneeded, while preserving the ability to set them when BG predictions actually justify.


scottleibrand:
If no objections, I'd like to go ahead and merge this to dev.

PR #845, opened 2017-12-01
Date error with MDT CGM

cjo20:
Fixes the date/jq errors towards the end of the loop in MDT CGM. Needs testing by someone with non-MDT CGM before merging.


luiesa:
fixed fine!

cjo20:
Exactly what error are you getting?

bin/oref0-pump-loop.sh
```diff
@@ -690,7 +690,11 @@ function wait_for_bg {
 
 function glucose-fresh {
     # set mtime of monitor/glucose.json to the time of its most recent glucose value
-    touch -d "$(date -R -d @$(jq .[0].date/1000 monitor/glucose.json))" monitor/glucose.json 2>&3
+    if grep "MDT cgm" openaps.ini 2>&3 >&4; then
```
scottleibrand: Would it be better to check for the existence of `.[0].display_time` here?  FWIW, my data from Dexcom includes both fields.

cjo20: This seems to be the check used to work out if it's MDT data or not elsewhere in the script, I was trying to be consistent. Do you have any preference?

scottleibrand: My Dexcom data has a display_time field as well, so it's not really a MDT-specific thing.  I think we should be able to just use display_time across the board, but probably safer for backwards compatibility to fall back to date if display_time doesn't exist?

cjo20: Ok, I'll try to take a look at it tomorrow.

PR #846, opened 2017-12-01
Display seconds in SMB wait time

cjo20:
Because x.y minutes bugs me :P 

Still need to check the regex stuff works properly and sort out tabs vs spaces - don't merge yet, I created this PR already in case anyone wants to comment.


scottleibrand:
Ready to test and merge now?

cjo20:
Think so, yup.

PR #848, opened 2017-12-02
Add indexError.html to handle JSON Value error

efidoman:


scottleibrand:
What causes this file to get used?

scottleibrand:
Opened https://github.com/efidoman/oref0/pull/3 to merge this into the same branch as the change that actually uses it.

PR #851, opened 2017-12-02
added check for custom OPENAPSDIR enviornment var

efidoman:


PieterGit:
@efidoman : good work to make the openaps dir configurable again. Currently the `myopenaps` is hardcoded on several places, see e.g. https://github.com/openaps/oref0/search?utf8=%E2%9C%93&q=myopenaps&type=

I would prefer `OPENAPS_DIR` instead of `OPENAPSDIR`, to make it similar to `API_SECRET`  and `NIGHTSCOUT_HOST`

efidoman:
@PieterGit - OK updated environment variable to OPENAPS_DIR. I will look into the other scripts.

efidoman:
@PieterGit - I think I've checked all those locations where myopenaps is used/referenced and changed the ones that require it to check OPENAPS_DIR environment  variable first. All in PR's.

PR #852, opened 2017-12-03
Create new webpage offline to monitoring the rig. Change the port of FLASK to port 80.

cduarte:


alimhassam:
The pull request should be to the dev branch I think. Maybe even a new branch to work on this offline feature? Fur example we need mmol conversion.

PieterGit:
This won't work when non root. Port < 1024 require root priviliges. With newer options for hardware coming up, and for use with PI3 rigs, I would advise to either:
- stick to a (free) port > 1024
- make the port configurable and only use 80 on an Explorer board

scottleibrand:
Since libmraa requires it, we’ve standardized on running everything as root for simplicity and ease of troubleshooting. If there is anywhere in the docs that doesn’t say to log in as root to install things, PRs welcome to fix.

PieterGit:
@scottleibrand @cduarte . I disagree. On Edisons we didn't find a user friendly way of upgrading the linux kernel, so we could not use the gpio sysfs kernel to access GPIO and we used mraa as a library to access GPIO. On raspberry pi's there is no need for that, so we should use the basic kernel features, and use a python or javascript gpio library and get rid of mraa. The fact that mraa does not work as a non-root is a major bug  in libmraa, and cannot be used as the sole reason to implement more stuff under root permission. It's not much work to create a library that makes it possible to interface to both libmraa and standard GPIO library and I'm willing to help on that when I get an Explorer HAT.

From a safety perspective, I would like that we work on improving the security of rigs and using root for everything is a very bad security practice, see for example 
https://askubuntu.com/questions/16178/why-is-it-bad-to-log-in-as-root or ask any security professional.

I would like to start improving the securing of the rig (and thus also the safety), and I disagree that everything on the rig should run under root permissions.

scottleibrand:
On running OpenAPS as root: what's your threat model here?  All of the good reasons (on that page and elsewhere I've seen) for not logging in as root either involve forced password entry on sudo (which we have never required even when logged in as `pi`) or GUI logins running web browsers, etc (which we don't allow).  I have yet to see a good argument for why users shouldn't log in as root when working on the CLI on a single-user embedded system. I am very reluctant to introduce additional complexity for users (i.e. supporting `pi` login on some systems and `root` on others) solely to blindly follow cargo-cult-style "best practices" that don't really apply in our use case.  If we want to improve the security of OpenAPS, we need to develop a threat model of what we're trying to secure against, and identify actions that will actually increase security for the average user without requiring significantly more work and confusion.  Since pump comms are wide open by default, and anyone nearby can bypass the rig and control the pump directly, anything we do to secure the rig against local attack is a wasted effort.  One area where I do think some additional security work might be useful is in enforcing secure randomly generated passwords (API secrets or tokens) in NS, to reduce the risk of a remote attacker maliciously entering carbs or uploading fake BGs.

With regard to flask etc, I would agree that it'd be better to run an nginx proxy on port 80.  My main objective there is actually to be able to run various different web apps on the same web server with different paths.  I actually tried this using https://www.digitalocean.com/community/tutorials/how-to-serve-flask-applications-with-uwsgi-and-nginx-on-ubuntu-14-04, but got stuck not really understanding what I was doing and didn't manage to get it working.  If someone wants to do that or something similar, that would be a welcome PR.

PieterGit:
I'll reply later on the security part. Running Nginx as a reverse proxy is quite easy, and also makes it possible not to use port 80 (http), but use port 443 and secure it with https or http/2. Hard part of securing nginx in this case is the fact that https requires certficates and certficates requires fqdn that resolve for the client. It would be a bit of work to get that properly bootstrapped and e.g. use letsencrypt to generate SSL certificates for your rig and be also able to use it offline.

I'm quite willing to help integrating the Nginx part in the coming Christmas period, even though I personally will not use it and am not interested on running a webserver on the rig. I think Nightscout does a good job of providing enough info on the the current status of the rig. If there is an error I use SSH to view the logs directly.

Main configuration command you use for nginx setup is http://nginx.org/en/docs/http/ngx_http_proxy_module.html#example

Search for "Nginx reverse proxy" to find tutorials such as: 
- https://www.nginx.com/resources/admin-guide/reverse-proxy/
- https://medium.com/@utkarsh_verma/configure-nginx-as-a-web-server-and-reverse-proxy-for-nodejs-application-on-aws-ubuntu-16-04-server-872922e21d38

Given the fact that the webserver will get very little requests, I will not uWSGI, but just use proxy_pass with a port for integration with the flask or even better nodejs backend.

scottleibrand:
Merged to the `flask` branch, and opening a new PR from there to dev.

PR #855, opened 2017-12-05
Updated oref0-online.sh to enable persistent BTPAN

auxlife:
Updated oref0-online.sh to enable persistent BTPAN, checks preferences.json for "presistent_btpan": true


drnoname82:
What is the advantage of persistent BT PAN?

auxlife:
@drnoname82 I am able to have xdrip push bg readings to xdripaps via the bt network and the rig can continue to access the internet via wifi (if available) if not it uses the bt pan for internet also.

I noticed a substantial drain on by cell phones battery when the rig was bt tethered, now since it doesn't route internet traffic over the bt pan the battery life is acceptable. I have not noticed any significant change in the rig's battery life.

drnoname82:
Ok nice, we currently solve this by having XDrip+ upload to 1) the rig via BT and 2) Nightscout. The BT upload simply fails when there is no BT PAN and the rig then gets glucose data from Nightscout via wifi.

auxlife:
@drnoname82 i was under the impression that if you did that then the bg readings would be doubled up in NS whenever you have the rig getting bg via btpan?

drnoname82:
That used to be the case, but the upload from the rig to NIghtscout fails/does not happen since one of the recent OpenAPS releases. I consider that a feature, not a bug :)

scottleibrand:
In light of @drnoname82's comments, do we still want to do this?

auxlife:
@scottleibrand If openaps is going to not push xdripaps data to NS going further i guess the primary reason for Persistent BT is removed. Was that a planned move or a bug?

scottleibrand:
I don't use xdripaps, so I don't really know how it's supposed to work.

scottleibrand:
@drnoname82 do you see benefit in this, or should we close it?

applehat:
I responded by email but apparently it didn't stick. I see benefit to this. I like having a Bluetooth connection always on (for easy rig access), but would also love to have it be able to use wifi as it's data connection when I'm at a location with it saved.

Visiting my mom's house I always have to grab her laptop and a USB cable if I have an issue. A constant Bluetooth connection would be great.

drnoname82:
I do not see any advantage, but no harm either.  @applehat We use an Android phone which will share any internet connection via BT, so the only wifi I have configured on the rig is our home wifi

applehat:
@drnoname82 same here, tho in some places I have really spotty internet on my cell at family houses, but fine connection on wifi. It still works fine looping, but with no Nightscout uploads (and pushovers every 20 minutes to let me know). I see no downside to leaving it as an option if people wanna use it.

scottleibrand:
If anyone wants to use this, please speak up.  If we want to do this, we'll need to fix all the extra whitespace changes (or make them separately if they're needed) so I can see what's actually changing here and figure out what needs to be tested for regressions affecting people not using the option.

scottleibrand:
Sounds like at least @applehat would like this feature, so we'll need to fix all the extra whitespace changes (or make them separately if they're needed) so I can see what's actually changing here and figure out what needs to be tested for regressions affecting people not using the option.  @auxlife would you like to give that a shot, or should I retarget this to a feature branch to do so there?

alimhassam:
I don't fully understand this change yet but at first sight it seems like an improvement to me. I would like to try properly, it but I can't do so for a couple days at least.

If understand this right, this would allow to use Bluetooth tethering by turning it on on the phone, and it would stay on even if there is wifi, is that correct?

tim2000s:
I used to run my portable rigs like this prior to the switching in oref0-online.sh. The main benefit of persistent bt tethering is that when you use hot buttons to set temp targets they always work from the android phone. 

In addition, with the rig always tethered to the phone, you can relieve the rig of Wi-fi responsibilities and allow android to take over, meaning that you get a better layer of security and ease of use on the various random networks you may want to join. I’ve previously hacked oref0-online.sh to do something similar so I think it’s worthwhile, even high it’s not for the original intent.

drnoname82:
I am still trying to understand why it was necessary to modify oref0-online.sh for that. We simply only have our home wifi configured in the rig itself. If the home wifi is not available, then it BT tethers to the phone and the phone provides internet connectivity via mobile network or wifi. I find it much more comfortable to configure new wifis on the phone than on the rig. This approach should work for all Android users. For iPhone users I do not understand the advantage because the iPhone will not accept BT connections if there is no mobile network coverage.

This change is perfectly fine if it helps anyone, I only wonder whether there is no simpler way to achieve the desired behavior.

auxlife:
@drnoname82 For example at my office we are required to disable all communications between wireless clients and between the wireless vlan and the rest of the network; the rig is able to pull bgs from ns and push its updates, but i can't ssh into it unless i plug it in. With the Persistent BT pan i can always ssh into the rig from my phone. I believe you are correct this wouldn't be of much help for iphone users as you can't have the bt pan if there is no internet connection on the iphone to share.

@alimhassam Correct, the rig will use the wifi for internet if its available, but will still have the bt network connection.

@scottleibrand I'd love to help in any way possible, sent you a message on gitter

tim2000s:
I do wonder what the impact on battery life of the rig will be? When I’m using only one radio I get about 14 hours from a 2000mAh battery. 

I’d also ask questions of your phone’s Bluetooth stack if you’re seeing high drain with BT tethering. My android phone doesn’t drain at all fast with bt.

auxlife:
@tim2000s  i'm using the 2500mAh battery and with just oref0 via wifi w/o any remote logging i was getting down to 60% from a 13 hour day, now with wifi & bt, along with fail2ban and splunk forwarder i'm getting to 57-58% so about a 5-8% increase in battery usage

dcacklam:
My 2 cents on this...

Right now, the 'BT switching' functionality can be rather unstable with some phones & situations. My Nexus 6 and 6P, combined with Pi BT, did not tolerate the up/down BT pan very well.

Also, Android 8's default behavior when it finds itself on wifi-with-no-internet, is to drop off the wifi network entirely (switch to LTE) - which has the nasty side-effect of disconnecting the rig from it's CGM data (xdripAPS cgm option). BT-PAN being up, would allow the rig a redundant channel to the CGM data when running in xdripAPS mode (to avoid this rig-on-isolated-wifi, phone on LTE split-brain scenario), but of course when the rig sees wifi it tears down the PAN....

I've commented out the 'if wifi turn off PAN' code in my own rig, so that it always maintains the BT connection. But it would be nicer if this was an official preferences option - even-if it was off-by-default. 

More choice is always good.

jaylagorio:
+1 for this feature. I had a PR a while ago that tried DHCP a few times before giving up connecting to BT because my phone doesn't issue an IP address when it's out of cell signal, but it'll keep an IP address assigned to the rig if it doesn't disconnect. It was good, but I still have some trouble when commuting via subway. This would probably further strengthen the connection between stations. Maybe it can go into `0.7.0-dev` and I can bang on it for a week to see?

jaylagorio:
I think the functionality this PR implements was just merged with #1048

scottleibrand:
Ok, closing this then.  If anything is missing in #1048 feel free to open a new PR.

PR #856, opened 2017-12-05
Pushover glances

efidoman:
Pushover has a new powerful glances API that sends complications to an apple watch or small bits of data to a smart watch face. More information at the following web page: https://pushover.net/api/glances

This change enhances the pushover notification script to optionally send information to the pushover glances API for watch face updates once every 10 minutes. It works well for apple watch complications. It has not yet been tested for android wear watchface updates yet.

If the openaps parameter, pushoverGlances does not exist in preferences.json or is set to false, then nothing is done or sent related to pushover glances. If pushoverGlances is set to true in preferences.json, then this code will fire every 10 minutes to update pushover glances / smart watch complications.

Note that apple limits the daily total number of complication updates to save watch battery, so we cannot update this glance data more often than every 10 minutes today.

Apple watch face screenshots below (large watchface and Modular)
img_5637
img_5639
img_5642
img_5645
img_5647
img_5648
img_5649
img_5651
img_5655
img_5676
img_5677


bin/oref0-pushover.sh
```diff
@@ -28,3 +28,76 @@ elif [[ $ONLYFOR =~ "insulin" ]] && ! cat $FILE | egrep "maxBolus"; then
...
+    echo "No preference or false preference for pushoverGlances"
+  exit
+fi
+
+
+GLANCES="monitor/last_glance"
+GLUCOSE="monitor/glucose.json"
+if [ ! -f $GLANCES ]; then
+  # First time through it will get created older than 10 minutes so it'll fire
+  touch $GLANCES && touch -r $GLANCES -d '-11 mins' $GLANCES
```
scottleibrand: I think your first `touch` here is redundant.

efidoman: I'll remove that first touch. Also, I'm currently refactoring it to perform faster by reading values from smb-suggested.json only. Give me a day or two to make these changes and test it.

efidoman:
@scottleibrand - I finished the refactoring, but at the same time I added another notification option for ifttt. Let me know if this needs to go somewhere else but I felt it fit best here so I added another optional notification that will send an ifttt maker event "carbs-required" if additional carbs are required and if environment variable "MAKER_KEY" is set. This is the ifttt webhooks Maker key https://ifttt.com/maker_webhooks

The "carbs-required" event can then be used with a ifttt recipe to perform an action based on carbs required. A good use case is to use the ifttt phone action to get a phone call with this event that will read out in human language the additional carbs and other vital facts. It will leave a voice mail if not answered.

bin/oref0-pushover.sh
```diff
@@ -28,3 +28,94 @@ elif [[ $ONLYFOR =~ "insulin" ]] && ! cat $FILE | egrep "maxBolus"; then
...
+      direction="--"
+    elif [ "${delta}" -lt "-2" ]; then
+      direction="-"
+    elif [ "${delta}" -lt "3" ]; then
+      direction=""
+    elif [ "${delta}" -lt "8" ]; then
+      direction="+"
+    else
+      direction="++"
+    fi
```
scottleibrand: Normally direction arrows are based on something like a 15m average.  Do we have the avgDelta available to use here?

bin/oref0-pushover.sh
```diff
@@ -28,3 +28,94 @@ elif [[ $ONLYFOR =~ "insulin" ]] && ! cat $FILE | egrep "maxBolus"; then
...
+    echo "No preference or false preference for pushoverGlances"
+else
+  GLANCES="monitor/last_glance"
+  GLUCOSE="monitor/glucose.json"
+  if [ ! -f $GLANCES ]; then
+    # First time through it will get created older than 10 minutes so it'll fire
+    touch $GLANCES && touch -r $GLANCES -d '-11 mins' $GLANCES
+  fi
+
+  if test `find $GLANCES -mmin +10`
```
scottleibrand: If you have more than one rig, this will be too often.  Do we need to run it less often in that scenario to avoid burning through the complications budget for the day? Or will the watch just ignore the ones sent more than once every 10 minutes?  I wonder if this is something that we should eventually port over to NS to avoid duplicate notifications from multiple rigs.

efidoman: Good point about multiple rigs. We really only want one rig to do this work ... the one that is currently active. I would need to test it, but I think we are still safe in this scenario because the non-active rigs would have an enact time > 5 minutes right? In this case the entire script exits at the beginning.

bin/oref0-pushover.sh
efidoman: Is there an avgDelta stored somewhere in an openaps json file? I could calculate it from glucose.json if not.

scottleibrand: No, we calculate avgDelta in determine-basal, but don't write it out anywhere.  I guess we could either add that to the enacted json, or recalculate it here.  Or you could do something more like what @cjo20 did with the mdt-trend-arrow stuff...

bin/oref0-pushover.sh
scottleibrand: When running multiple rigs in the same or adjacent rooms, it's very common for one rig to enact something, and then as soon as it's done another rig reads the pump and enacts something slightly different.

bin/oref0-pushover.sh
efidoman: I looked at the mdt-trend-arrow code and it definitely looks like something that xdrip-js-logger should be doing. If it is done there, then any downstream consumer of xdrip-js-logger glucose data would have better trend arrows including NS. Let me look at that in the next couple of days. Then this apple watch pushover glances update could just use the trend arrow from the glucose data directly. @scottleibrand - What do you think?

scottleibrand:
Yeah, I prefer the idea of using the CGM trend arrows. That also makes it possible to ensure the arrows always match the normal Dexcom receiver/app for everyone using that.

efidoman:
Ok - I'll refactor this now to use the glucose trend arrows. That way it shows whatever the CGM or collector-system put there.

efidoman:
@scottleibrand - I refactored it to use the direction field in the latest glucose.json entry. I tested it and it works good for me. look at the latest commit. I had some weird cut-n-paste stuff going on with the first 7546b55 commit.

efidoman:
I figured out why my carbs-required phone calls stopped working since 12/14. It seems ifttt limits the phone calls to 100 per month without documenting that limit. I did just find the email they sent me though 
---
Hi feibelman,You have reached IFTTT's limit of 100 phone calls.All of your active Applets that make phone calls are paused until Monday, January 01, 2018.
---

I was doing a lot of testing at the time and causing carbs-required events to occur by giving manual boluses and delaying carb entry. However, 100 means an average of 3 per day. I believe that I should tighten up the parameters a bit for a carbs-required maker event and avoid repeat notifications if possible. Not everyone will use this to trigger the phone action, so maybe these need to be parameters. I think the # of carbs-required before notification is already a parameter (I will check).

sulkaharo:
FYI I asked pushover and they’re apparently not kidding about the 1 message / 20 minutes limitation when using this API on iOS, so any functionality should be planned accordingly.

efidoman:
@sulkaharo - Thanks for the heads up. I'm OK with considering a change to the default pushover glance to once every 20 minutes. However, I've been testing the pushover glances notification to my apple watch with 10 minute updates for several weeks now and I get all the updates. It doesn't seem to throttle me at the 10 minute rate.

Earlier in the testing cycle, when I tried 5 minute updates, I reached my limit sometimes around mid-day, so it seems 10 minutes may be the real quota. I understand Apple setting limits, but vague limits that aren't clearly understood are a little frustrating. ;->

efidoman:
I've been running this for the last 3 full weeks using the pushover glances at 10 minute intervals and it hasn't missed an update or ran out of quota for the day. I'd say the 10 minute interval is most likely OK for this use case.

scottleibrand:
Cool. Do we need any other tweaks for multi-rig support, or are we good to merge?

efidoman:
Is there an quick solution to support multiple rigs or would it require porting it to work in openaps loop?

If it requires porting it to run in the loop, can you give me an overview of what needs to be done and in which files?

PR #862, opened 2017-12-07
Fixed the dev oref0-pump-loop script to drop x12-incompaitble reports…

dcacklam:
… IF an x12 pump is detected

In previous versions of oref0, this action was done by manually implementing an alias-change (as documented in past versions of the oref0 code).

Since the code here is not done via an alias anymore, but rather spelled out in the script, an if/then/else block is required to remove the offending reports when an incompatible pump is detected. 

Tested and verified against my 712.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -596,7 +596,8 @@ function get_settings {
         # On all other supported pumps, these reports work. 
         NON_X12_ITEMS="settings/bg_targets_raw.json settings/basal_profile.json settings/settings.json"
     fi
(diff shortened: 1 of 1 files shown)


bin/oref0-pump-loop.sh
```diff
@@ -587,8 +587,16 @@ function refresh_old_profile {
 
 # get-settings report invoke settings/model.json settings/bg_targets_raw.json settings/bg_targets.json settings/insulin_sensitivities_raw.json settings/insulin_sensitivities.json settings/basal_profile.json settings/settings.json settings/carb_ratios.json settings/pumpprofile.json settings/profile.json
 function get_settings {
-    retry_return openaps report invoke settings/model.json settings/bg_targets_raw.json settings/bg_targets.json settings/insulin_sensitivities_raw.json settings/insulin_sensitivities.json settings/basal_profile.json settings/settings.json settings/carb_ratios.json 2>&3 >&4 | tail -1 || return 1
-    # generate settings/pumpprofile.json without autotune
+    if grep 12 settings/model.json
```
alimhassam: (minor) should be grep -q unless you want to print the model in the logs everytime.

dcacklam: Done.

PR #865, opened 2017-12-13
Ignore records with 0 sgv

cjo20:
It seems that there are occasionally records from MDT cgm with a sgv of 0. This causes the trend arrows to think that BG is rising rapdily. 

Written without my rig nearby (oops), so needs to be tested by me or another MDT CGM user before merge.


cjo20:
Been running with this for several hours, looks like it's working ok.

PR #872, opened 2017-12-25
Speed up autotune and make it work in NS

scottleibrand:
This significantly speeds up autotune by only passing the last 6 hours of insulin history when calculating IOB.  In addition to running faster on rigs etc, this makes it practical to run autotune in NS (see https://github.com/nightscout/cgm-remote-monitor/pull/3191).  To support that, this also changes lib references from absolute to relative for compatibility.


lib/autotune-prep/categorize.js
```diff
@@ -152,6 +163,26 @@ function categorizeBGDatums(opts) {
...
+        // trim down IOBInputs.history to just the data for 6h prior to BGDate
+        //console.error(IOBInputs.history[0].created_at);
+        var newHistory = [];
+        for (h=0; h<fullHistory.length; h++) {
+            var hDate = new Date(fullHistory[h].created_at)
+            //console.error(fullHistory[i].created_at, hDate, BGDate, BGDate-hDate);
+            //if (h == 0 || h == fullHistory.length - 1) {
+                //console.error(hDate, BGDate, hDate-BGDate)
+            //}
+            if (BGDate-hDate < 6*60*60*1000 && BGDate-hDate > 0) {
```
alimhassam: i'm not sure how likely it is, or if it affects the result, but should this timeout be max(dia, 6h)?

alimhassam:
Just a suggestion, but i think it would be a good idea (and probably easier) to create a small openaps/autotune project for the autotune stuff and oref0 and cgm-remote-monitor can then import it, rather than copying the file and having to maintain both copies.

scottleibrand:
That sounds like a good approach.  Do you know how to actually do the import if we move the autotune code to its own project?

lib/autotune-prep/categorize.js
scottleibrand: To do that would require some re-plumbing to get DIA into this part of the code.  Even for people with 7h or 8h DIA, I'm not sure that including the older insulin treatments would be any more accurate, since we stop considering carbs after 6h.  And either way, it'd be a tiny effect, so I'm fine leaving it as is with 6h hard-coded, unless someone else wants to do the work to make it use DIA.

alimhassam:
I'm not an expert in nodejs, but i think for the "import" part, you just need to add the new project to "package.json" 
after moving the files over you'll also need to create a new package.json in that new project to export the module.

I think there are npm commands for both these things, i'm hoping it should be straightforward.

scottleibrand:
Opened https://github.com/openaps/oref0/issues/874 to track that.  If no other objections, I'll go ahead and merge this to get the speedup into oref0 in the meantime.

PR #873, opened 2017-12-27
Pushover Priority to Urgent

jpcunningh:
Set Pushover Priority to urgent to ensure user sees the message.

### DIFF
--- a/bin/oref0-pushover.sh
+++ b/bin/oref0-pushover.sh
@@ -26,5 +26,5 @@ elif [[ $ONLYFOR =~ "carb" ]] && ! cat $FILE | egrep "add'l"; then
 elif [[ $ONLYFOR =~ "insulin" ]] && ! cat $FILE | egrep "maxBolus"; then
     echo "No additional insulin required."
 else
-    curl -s -F "token=$TOKEN" -F "user=$USER" -F "sound=$SOUND" -F "message=$(jq -c "{bg, tick, carbsReq, insulinReq, reason}|del(.[] | nulls)" $FILE) - $(hostname)" https://api.pushover.net/1/messages.json && touch monitor/pushover-sent
+    curl -s -F "priority=2" -F "retry=60" -F "expire=600" -F "token=$TOKEN" -F "user=$USER" -F "sound=$SOUND" -F "message=$(jq -c "{bg, tick, carbsReq, insulinReq, reason}|del(.[] | nulls)" $FILE) - $(hostname)" https://api.pushover.net/1/messages.json && touch monitor/pushover-sent
 fi
--- a/bin/oref0-pushover.sh
+++ b/bin/oref0-pushover.sh
@@ -6,6 +6,11 @@ FILE=${3-enact/smb-suggested.json}
 SOUND=${4-none}
 SNOOZE=${5-15}
 ONLYFOR=${6-carbs}
+PRIORITY=0
+RETRY=60
+EXPIRE=600
+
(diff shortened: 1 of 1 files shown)


scottleibrand:
This would cause problematic alarm fatigue for users running multiple rigs. We probably shouldn’t do this by default in oref0, but rather do so once we’ve moved the oref0 pushover stuff over to NS, where we can guarantee only one notification per event. In the mean time, we could create a preference to trigger the urgent priority for users who want that...

jpcunningh:
Thanks! I'll work on adding a preference for people running only one rig, and add it into this pull request. I'll add preferences for overriding the default priority (0), retry, expiration, and sound.

jpcunningh:
Is it acceptable to put a preference in preferences.json and have oref0-pushover extract it with jq? Or should oref0-setup ask the user for the configuration parameter and pass it to oref0-pushover on the crontab command line? The script would need the location of preferences, but oref0-setup has it available to pass in on the crontab line.

I prefer preferences.json so it is easier to change as the user's needs change rather than running oref0-setup again or manually editing the crontab. I'm new, though, so I don't know the established conventions.

jpcunningh:
And change the ONLYFOR command line option to be a preference as well?

jpcunningh:
In general is this the correct branch to submit pull request to? What's the process for keeping the docs in sync? If this request is merged, when and how to I get the docs updated to document the new feature?

scottleibrand:
That looks reasonable.  What situations have you tested so far?

scottleibrand:
If you want to open a PR against the docs repo, we can then decide whether to merge it once this is merged to dev, or wait until it's released to master.

scottleibrand:
Oh, and yes, new stuff like this gets merged to oref0 dev first (unless we need to retarget to a shared feature branch first for collaboration), and then oref0 dev eventually gets merged to master when it's well tested and ready to release.

jpcunningh:
I made a first pass through all of the options to make sure they worked. I'm going to make another pass through this evening before going to bed, and I'll reply here when I'm through.

jpcunningh:
They all seem to work. With nothing put in the preferences file, it's behavior is unchanged. I also tested each of the new preferences, and they worked as well.

scottleibrand:
Ok, unless anyone else has any comments or wants to test anything, I’ll merge it soon. Thanks!

bin/oref0-pushover.sh
```diff
@@ -14,7 +19,61 @@ if [ -z $TOKEN ] || [ -z $USER ]; then
...
+
+PREF_VALUE=$(cat $PREF_FILE | jq .pushover_retry 2>/dev/null)
+
+if [ ! -z $PREF_VALUE ] && [ $PREF_VALUE != "null" ] && [ "$PREF_VALUE" -eq "$PREF_VALUE" ]; then
+    RETRY=$PREF_VALUE
+fi
+
+PREF_VALUE=$(cat $PREF_FILE | jq .pushover_expire 2>/dev/null)
+
+if [ ! -z $PREF_VALUE ] && [ $PREF_VALUE != "null" ] && [ "$PREF_VALUE" -eq "$PREF_VALUE" ]; then
```
efidoman: Just curious, what does the last comparison in this if statement do? It seems to be comparing a variable value to itself which should always match?

jpcunningh:
The last comparison tests to ensure the value is a number. Non-numbers will
result in an error which will be implicitly considered as false for any
numeric value passed to evaluation operations.

Issue #847, opened 2017-12-01
bluez 5.47/5.48 and openxshareable issue


rkresha:
```
File "/usr/local/lib/python2.7/dist-packages/openxshareble/ble/uart.py", line 78, in pair_auth_code
    self._auth = self._uart.find_characteristic(self.AUTH_UUID)
  File "/home/root/src/Adafruit_Python_BluefruitLE/Adafruit_BluefruitLE/interfaces/gatt.py", line 48, in find_characteristic
    for char in self.list_characteristics():
  File "/home/root/src/Adafruit_Python_BluefruitLE/Adafruit_BluefruitLE/bluez_dbus/gatt.py", line 56, in list_characteristics
    paths = self._props.Get(_SERVICE_INTERFACE, 'Characteristics')
  File "/usr/lib/python2.7/dist-packages/dbus/proxies.py", line 145, in __call__
    **keywords)
  File "/usr/lib/python2.7/dist-packages/dbus/connection.py", line 651, in call_blocking
    message, timeout)
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.InvalidArgs: No such property 'Characteristics'
```

Fixed by rolling back to 5.37 of bluez... probably need to investigate a fix so we can stay current with bluez. Note: in the code we jumped from 5.37 directly to 5.47 due to #724 ... so I'm vulnerable now, but functionality over security.

cduarte:
Do you use openxshareable? I read about the openxshareable this week and saw that the last update is about 2 yesars ago.

rkresha:
Yep, G4 w/ share, openxshareable is my offline method

cduarte:
Thanks for the answer :)

do you use this: https://github.com/openaps/openxshareble ?

cduarte:
So with this you don't need to use a phone with xdrip? correct? if yes... is perfect :)

rkresha:
> Thanks for the answer :)
> 
> do you use this: https://github.com/openaps/openxshareble ?

Yes, if you set your CGM when running `oref0-setup.sh` to "ShareBLE" it will clone that repo. Or I've run into a dependency issue where it needs to make the bluetooth driver first before it calls the script so I have the clone of that repo before I run oref0-setup.sh (I don't remember the specific error and maybe they've fixed it by now I just still continue to do it the way I used to get it to work)... Do you have a G4 w/ share or a different CGM?

rkresha:
[...]

PR #876, opened 2017-12-30
Upgrade bluez to 5.48, start bluetoothd if it's not running, remove experimental

PieterGit:
- bugfix: also start bluetoothd in oref0-bluetoothup if it's not running
- upgrade bluez to 5.48, changelog http://www.bluez.org/release-of-bluez-5-48/
- remove `--experimental` (finally) on bluetoothd startup and change it in /etc/rc.local if it's still there

Tested with pebble and works. Will test bluetooth tethering later on.

### DIFF
--- a/bin/oref0-bluetoothup.sh
+++ b/bin/oref0-bluetoothup.sh
@@ -1,13 +1,18 @@
 #!/bin/bash
 
+# start bluetoothd if bluetoothd is not running
+if ! ( ps -fC bluetoothd ) ; then
+   sudo /usr/local/bin/bluetoothd &
+fi
+
 if ! ( hciconfig -a | grep -q "PSCAN" ) ; then
    sudo killall bluetoothd
-   sudo /usr/local/bin/bluetoothd --experimental &
+   sudo /usr/local/bin/bluetoothd &
 fi
 
 if ( hciconfig -a | grep -q "DOWN" ) ; then
    sudo hciconfig hci0 up
-   sudo /usr/local/bin/bluetoothd --experimental &
+   sudo /usr/local/bin/bluetoothd &
 fi
 
 if !( hciconfig -a | grep -q $HOSTNAME ) ; then
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -624,19 +624,20 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
         if getent passwd pi > /dev/null; then
             bluetoothdminversion=5.43
         else
-            bluetoothdminversion=5.47
+            bluetoothdminversion=5.48
         fi
         bluetoothdversioncompare=$(awk 'BEGIN{ print "'$bluetoothdversion'"<"'$bluetoothdminversion'" }')
         if [ "$bluetoothdversioncompare" -eq 1 ]; then
-            cd $HOME/src/ && wget -4 https://www.kernel.org/pub/linux/bluetooth/bluez-5.47.tar.gz && tar xvfz bluez-5.47.tar.gz || die "Couldn't download bluez"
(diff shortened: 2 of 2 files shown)


PieterGit:
Tethering also works. Please review and merge to dev when ok.

PieterGit:
Running stable for 6 days now. The new bluetoothd versions seems more stable (less bluetooth messages in /var/log, didn't experience kernel problems caused by bluetooth with this bluez version).

rkresha:
Running into issues with ShareBLE at first glance.. Will report back

scottleibrand:
Ok, LMK if it's fixable with 5.48, or if we need to revert.

rkresha:
Not easily fixable based on the work I've done, would prefer you to revert but if you want to forge ahead then I'll just have to make modifications to the setup script to install 5.47 and keep the experimental flag.

rkresha:
@scottleibrand please advise what you are thinking in terms of revert vs forge ahead

scottleibrand:
I’ve been holding off to see if anyone comes up with a fix, but if not I’ll probably want to revert before 0.6.1 release.

PieterGit:
@rkresha if you are hitting https://github.com/openaps/oref0/issues/847 than revering to 5.47 won't help and according to the  Adafruit_Python_BluefruitLE issue we should revert bluez to 5.37 to make  openxshareble work. 

I would suggest that we stick to 5.48 by default (and drop the experimental). Perhaps we can make the oref0-setup use bluez 5.37 temporary (and warn for the security issues) for openxshareble users. I can't create a fix for the openxshareble because EU Dexcom G4 does not support bluetooth.

PieterGit:
If openxshareble also doesn't work with 5.47 (which is installed with oref0 0.6.0), there should be no point of holding back the upgrade to bluez 5.48 for oref0 0.6.1. If no fix for openxshareble  comes before 0.6.1 then openxshareble  be marked as broken for oref0 0.6.0 and oref0 0.6.1. 

Unless there are other issues with  bluez 5.48 that I am not aware of.

rkresha:
@PieterGit I ended up getting 5.47 working with your reference to the other person having an issue and their fix. I updated the gatt.py and got it going just fine.

scottleibrand:
Do we need to update oref0-setup to reflect what you did? If so, could you PR that?

tepidjuice:
To prevent an edge-case issue I had when updating my rig. Could `&& rm -r bluez-5.48.tar.gz &&` be added to this line before the `wget` and after the `cd /$HOME/src/` 

Without this if `bluez-5.48.tar.gz` already exists wget will save to `bluez-5.48.tar.gz.1` causing the tar command to fail if `bluez-5.48.tar.gz` is incomplete/corrupted. This occurred when updating due to poor internet the wget command in line 640 of oref0-setup failed.

scottleibrand:
Sounds like a good improvement.  Can you open a PR against dev to do so?

tepidjuice:
#909

rkresha:
@scottleibrand You want to fork bewest's (https://github.com/bewest/Adafruit_Python_BluefruitLE) under openaps like the ultimate goal (along with openxshareable) and then we can update the oref0-setup and PR it and then I can fork and update the gatt.py in Adafruit and PR it?

PieterGit:
@rkresha: did you also try bluez 5.48 with your Adafruit_Python_BluefruitLE patch? Can you also update https://github.com/openaps/oref0/issues/847

I still would like to remove the `--experimental` flag which does not seem to be needed anymore (except possibly openxshareble, but then we need to fix that problem). If a serious bluetooth issues comes up we should be able to upgrade to bluez 5.48+.

rkresha:
@PieterGit 
https://github.com/openaps/oref0/issues/847#issuecomment-356758858
> "Get" with signature "ss" on interface "org.freedesktop.DBus.Properties" doesn't exist
> 
> Get this error now with bluez5.48 and openxshareable #876

PR #878, opened 2018-01-01
Fix iob calculation error

jpcunningh:
The IOB calculation was forcing a minimum of 5 hour DIA by calculate.js; however, the total.js script was only using the history up to the user's pump profile DIA setting, not the full 5 hours if their pump was set to less than 5 hour DIA.

This change makes total.js and calculate.js force a 5 hour minimum consistently.

### DIFF
--- a/lib/iob/calculate.js
+++ b/lib/iob/calculate.js
@@ -59,7 +59,7 @@ function iobCalcBiLinear(treatment, time, dia) {
 }
 
 
-function iobCalc(treatment, time, dia, peak, profile, curve) {
+function iobCalc(treatment, time, curve, dia, peak, profile) {
 
     if (!treatment.insulin) return {};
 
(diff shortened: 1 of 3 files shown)


jpcunningh:
I tested the new code to ensure that a pump DIA setting of 5 or 6 yield the same results with the old code and the new code. Of course, a DIA of 3 or 4 setting in the pump yield significant differences as the code now consistently enforces a minimum DIA of 5.

sulkaharo:
Anu chance for an unit test fix that’d notice this?

jpcunningh:
I have a set of test inputs that could be used as a starting point of a unit test. Examining the output for each of the curves to ensure the outputs match the expected output could be written fairly easily.

jpcunningh:
The test inputs captured from last night showed as high as a half unit discrepancy in net insulin on board at times if the pump DIA was set to 4.

jpcunningh:
sulkaharo, I added a unit test that will test for the proper behaviour.

scottleibrand:
I think this is ready to merge to dev for further testing if no objections.

PR #879, opened 2018-01-02
Enable IOB calc to recognize suspends

jpcunningh:
This pull request updates history.js to recognize pump suspensions when calculating IOB.

To enable the new feature, set the preference suspend_zeros_iob to true.

### DIFF
--- a/lib/iob/history.js
+++ b/lib/iob/history.js
@@ -94,7 +94,6 @@ function splitAroundSuspends (currentEvent, pumpSuspends, firstResumeTime, suspe
         } else {
             currentEvent.duration = ((currentEvent.date+currentEvent.duration*60*1000)-firstResumeDate)/60/1000;
 
-            currentEvent.timestamp = firstResumeTime;
             currentEvent.started_at = new Date(tz(firstResumeTime));
             currentEvent.date = firstResumeDate
         }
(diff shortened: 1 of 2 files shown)


drnoname82:
Sounds good. Is there any reason this should not be the default behavior?

jpcunningh:
I think it should be the default after it goes through some confidence
testing. It was a relatively complicated add to how the history is
processed so I wanted to be able to turn it on only when I'm watching it.

I want to create a lot more boundary case checks, too, before I turn it on
and leave it on.

Jeremy

drnoname82:
Makes perfect sencse, thank you!

PieterGit:
@jpcunningh : patch looks good (haven't tested it). I don't use the pump suspend. Can you also add a unit test for `splitAroundSuspends`?

jpcunningh:
I put in several tests and made a note in a comment about additional test I plan to add over the next few days. Let me know if anybody thinks we should have additional tests. Thanks!

jpcunningh:
I believe this is ready to go. The method used to account for pump suspensions was find all of the suspensions, delete temp basals overlapping with the suspensions, and add 0 temp basals for the suspended time periods. It also accounts for the pump being suspended coming into the maximum DIA time (8 hours), and the pump being currently suspended at the time iob is calculated.

Several unit tests are implemented to test boundary cases. The method used was to setup scenarios using 0 temp basals, then recreate the scenarios with suspends and ensure the IOB results match.

Travis seems to be having a problem again, so I'll push a comment update in the morning to force it to run again.

scottleibrand:
I think we're ready to merge this to dev for further testing if no objections.

alimhassam:
I haven't tested it live, but from reading the code, it seems good. Nice set of Unit tests also!! 
Just putting this thought out there, to be even more correct, this same logic could be used to split around prime events, and when the reservoir is 0 and reuse the same logic.

PR #886, opened 2018-01-04
Consistent use of `$HOME` instead of `~`. Make it possible to install oref0 in another directory than `myopenaps`  and make it possible to start it from any directory

PieterGit:
Fixes:
- stop installing openaps in the current directory, use $HOME/myopenaps by default
- replace ~ for $HOME because that's our standard way of writing it and it's better,
"The tilde is part of a shell expansion (like in bash, csh, zsh, etc). The $HOME variable is exportable and can be used independent of a specific shell." (source: https://stackoverflow.com/questions/11587343/difference-between-home-and-tilde ) 

Tested on Explorer HAT with PI and it fixes the problem. 
@scottleibrand : can you please review and merge.
I think failed test is caused by Travis and not by this PR.

### DIFF
--- a/bin/oref0-log-shortcuts.sh
+++ b/bin/oref0-log-shortcuts.sh
@@ -1,32 +1,32 @@
 #!/bin/bash
 
-myopenaps=${OPENAPS_DIR:-"~/myopenaps"}
+myopenaps=${OPENAPS_DIR:-"$HOME/myopenaps"}
 
 # add crontab entries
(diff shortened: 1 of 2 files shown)


PieterGit:
@scottleibrand any reason why this PR has not been merged? This fixes a missed `$directory` and makes it possible to start `oref0-setup.sh` or `oref0-runagain.sh` from every directory (instead of only letting it work for `$HOME` and `myopenaps`.

scottleibrand:
Sorry, not sure why this dropped off my radar.  Let me close and reopen it to see if travis-ci fixes itself.

PR #894, opened 2018-01-07
Fix profile bootstrapping

scottleibrand:
When settings/bg_targets_raw.json got moved into NON_X12_ITEMS, it caused settings/bg_targets.json to be run before settings/bg_targets_raw.json, which caused it to fail on a new install unless the settings/bg_targets_raw.json was created manually.  This moves settings/bg_targets.json into NON_X12_ITEMS as well, so it gets run after settings/bg_targets_raw.json.

We'll need to make sure this still works will x12 pumps (that they don't need us to run the settings/bg_targets.json report).

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -581,9 +581,9 @@ function get_settings {
         NON_X12_ITEMS=""
     else
         # On all other supported pumps, these reports work. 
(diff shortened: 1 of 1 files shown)


scottleibrand:
This change fixed my newly installed rig.

@dacklam or anyone with an x12 pump: can you confirm whether this fix will work for x12s as well?

PR #900, opened 2018-01-09
Format Autotune Output Profile

jpcunningh:
Format the profile when copying it to settings to aid visually comparing to pump profile.

### DIFF
--- a/bin/oref0-autotune.sh
+++ b/bin/oref0-autotune.sh
@@ -217,8 +217,8 @@ do
     
     # Autotune  (required args, <autotune/glucose.json> <autotune/autotune.json> <settings/profile.json>), 
     # output autotuned profile or what will be used as <autotune/autotune.json> in the next iteration
-    echo "oref0-autotune-core autotune.$i.json profile.json profile.pump.json | jq . > newprofile.$i.json"
-    if ! oref0-autotune-core autotune.$i.json profile.json profile.pump.json | jq . > newprofile.$i.json; then
+    echo "oref0-autotune-core autotune.$i.json profile.json profile.pump.json > newprofile.$i.json"
+    if ! oref0-autotune-core autotune.$i.json profile.json profile.pump.json > newprofile.$i.json; then
         if cat profile.json | jq --exit-status .carb_ratio==null; then
             echo "ERROR: profile.json contains null carb_ratio: using profile.pump.json"
             cp profile.pump.json profile.json
--- a/bin/oref0-autotune.sh
+++ b/bin/oref0-autotune.sh
@@ -217,8 +217,8 @@ do
     
     # Autotune  (required args, <autotune/glucose.json> <autotune/autotune.json> <settings/profile.json>), 
     # output autotuned profile or what will be used as <autotune/autotune.json> in the next iteration
-    echo "oref0-autotune-core autotune.$i.json profile.json profile.pump.json > newprofile.$i.json"
-    if ! oref0-autotune-core autotune.$i.json profile.json profile.pump.json > newprofile.$i.json; then
+    echo "oref0-autotune-core autotune.$i.json profile.json profile.pump.json | jq . -M > newprofile.$i.json"
+    if ! oref0-autotune-core autotune.$i.json profile.json profile.pump.json | jq . -M > newprofile.$i.json; then
         if cat profile.json | jq --exit-status .carb_ratio==null; then
             echo "ERROR: profile.json contains null carb_ratio: using profile.pump.json"
             cp profile.pump.json profile.json
--- a/bin/oref0-autotune.sh
+++ b/bin/oref0-autotune.sh
@@ -217,8 +217,8 @@ do
     
     # Autotune  (required args, <autotune/glucose.json> <autotune/autotune.json> <settings/profile.json>), 
     # output autotuned profile or what will be used as <autotune/autotune.json> in the next iteration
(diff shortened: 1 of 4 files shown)


jpcunningh:
Absolutely. I'll update it later this evening to use jq.

Thanks!

jpcunningh:
All of the json calls in the module have been changed to equivalent jq calls.

jpcunningh:
I tested each command to verify the new command provided identical results
as the old command,  but another set of eyes would be good.

Thanks!

bin/oref0-setup.sh
```diff
@@ -1042,7 +1042,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
         (crontab -l; crontab -l | grep -q "cd $directory && ps aux | grep -v grep | grep -q 'oref0-autosens-loop' || oref0-autosens-loop 2>&1" || echo "* * * * * cd $directory && ps aux | grep -v grep | grep -q 'oref0-autosens-loop' || oref0-autosens-loop 2>&1 | tee -a /var/log/openaps/autosens-loop.log") | crontab -
         if [[ $ENABLE =~ autotune ]]; then
             # autotune nightly at 4:05am using data from NS
-            (crontab -l; crontab -l | grep -q "oref0-autotune -d=$directory -n=$NIGHTSCOUT_HOST" || echo "5 4 * * * ( oref0-autotune -d=$directory -n=$NIGHTSCOUT_HOST && cat $directory/autotune/profile.json | json | grep -q start && cp $directory/autotune/profile.json $directory/settings/autotune.json) 2>&1 | tee -a /var/log/openaps/autotune.log") | crontab -
+            (crontab -l; crontab -l | grep -q "oref0-autotune -d=$directory -n=$NIGHTSCOUT_HOST" || echo "5 4 * * * ( oref0-autotune -d=$directory -n=$NIGHTSCOUT_HOST && cat $directory/autotune/profile.json | jq . -M | grep -q start && cat $directory/autotune/profile.json > $directory/settings/autotune.json) 2>&1 | tee -a /var/log/openaps/autotune.log") | crontab -
```
scottleibrand: What is the advantage of using cat file > newfile here instead of cp file newfile?

jpcunningh:
There isn't one. It's left over from having the output piped through jq on
the previous commit. After studying the autotune code more yesterday
evening, it made much more sense to format it inside autotune rather than
externally through the cron command.

I'll put it back to cp like it is in the current dev as soon as I get home
this afternoon.

bin/oref0-setup.sh
```diff
@@ -1042,7 +1042,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
         (crontab -l; crontab -l | grep -q "cd $directory && ps aux | grep -v grep | grep -q 'oref0-autosens-loop' || oref0-autosens-loop 2>&1" || echo "* * * * * cd $directory && ps aux | grep -v grep | grep -q 'oref0-autosens-loop' || oref0-autosens-loop 2>&1 | tee -a /var/log/openaps/autosens-loop.log") | crontab -
         if [[ $ENABLE =~ autotune ]]; then
             # autotune nightly at 4:05am using data from NS
-            (crontab -l; crontab -l | grep -q "oref0-autotune -d=$directory -n=$NIGHTSCOUT_HOST" || echo "5 4 * * * ( oref0-autotune -d=$directory -n=$NIGHTSCOUT_HOST && cat $directory/autotune/profile.json | json | grep -q start && cp $directory/autotune/profile.json $directory/settings/autotune.json) 2>&1 | tee -a /var/log/openaps/autotune.log") | crontab -
+            (crontab -l; crontab -l | grep -q "oref0-autotune -d=$directory -n=$NIGHTSCOUT_HOST" || echo "5 4 * * * ( oref0-autotune -d=$directory -n=$NIGHTSCOUT_HOST && cat $directory/autotune/profile.json | jq . -M | grep -q start && cp $directory/autotune/profile.json $directory/settings/autotune.json) 2>&1 | tee -a /var/log/openaps/autotune.log") | crontab -
```
scottleibrand: It doesn't hurt anything, but there's no need for -M here: since it's writing to a pipe, monochrome is the default behavior.

bin/oref0-autotune.sh
```diff
@@ -217,8 +217,8 @@ do
     
     # Autotune  (required args, <autotune/glucose.json> <autotune/autotune.json> <settings/profile.json>), 
     # output autotuned profile or what will be used as <autotune/autotune.json> in the next iteration
-    echo "oref0-autotune-core autotune.$i.json profile.json profile.pump.json > newprofile.$i.json"
-    if ! oref0-autotune-core autotune.$i.json profile.json profile.pump.json > newprofile.$i.json; then
+    echo "oref0-autotune-core autotune.$i.json profile.json profile.pump.json | jq . -M > newprofile.$i.json"
+    if ! oref0-autotune-core autotune.$i.json profile.json profile.pump.json | jq . -M > newprofile.$i.json; then
```
scottleibrand: I believe -M is unnecessary here as well, as jq only outputs colored json by default if writing to a terminal.

jpcunningh:
👍

jpcunningh:
The order of the keys in the json stringify output was non-deterministic, making quick comparisons between days impossible.  This commit changed it to use the json-stable-stringify to both format to a readable format and produce it with keys in a deterministic order.

scottleibrand:
Ready to merge this to dev for further testing there?

jpcunningh:
Yes. It should be good to go. I'm getting consistent file order and format now.

PR #901, opened 2018-01-10
Delete non-OpenAPS log files earlier to save disk space

drnoname82:
Recently, several users have reported problems due to missing disk space on their rigs.  With the current configuration, log files can easily consume several hundred megabytes.

This small change lets the rig delete non-OpenAPS logs (system, cron, etc.) much earlier. I have been using this configuration for a while and it has kept the size of my non-OpenAPS log files below 100MB.

This PR does not touch the treatment of OpenAPS log files.

### DIFF
--- a/logrotate.rsyslog
+++ b/logrotate.rsyslog
@@ -14,9 +14,8 @@
 
 /var/log/syslog
 {
-	rotate 24
-	daily
-	size 10M
+	rotate 3
+	size 5M
 	missingok
 	notifempty
 	delaycompress
@@ -39,9 +38,8 @@
 /var/log/debug
 /var/log/messages
 {
-	rotate 24
-	daily
-	size 10M
+	rotate 3
+	size 5M
 	missingok
 	notifempty
 	compress


danamlewis:
Nice, thanks! Side note - if this works well, we might want to get this out as a release and do 0.6.1 soon-ish. I know there are some other things that need cleaning and fixing and testing, but may be a good idea to target this release since it's a growing problem for folks.

PieterGit:
Thanks for this PR. Just found out my production rig ran out of diskspace.

sulkaharo:
I’ve set the number of logs a bit higher and set the interval to hourly. Not sure what the interval is if it’s not defined?

drnoname82:
Interval definition and size are not meant to be used together ( https://askubuntu.com/questions/548632/logrotate-using-multiple-criteria-for-rotating ) which is why I chose only size. As far as I understand, logrotate is currently only run daily since it is in /etc/cron.daily

The different log files grow at very different rates, which is why I thought only rotating them by size is the most universal solution. However, I think most OpenAPS users will never ever look at the system logs anyway :)

scottleibrand:
As a future enhancement, I wouldn't mind running logrotate in cron.hourly instead of cron.daily.  Some error conditions generate a problematic level of logs in less than 24h.

PR #904, opened 2018-01-13
Autotune Command Line Option to Classify UAM as Basal

jpcunningh:
This adds an option to autotune for a user to manually add a --categorize-uam-as-basal option to the command line to add all UAM categorizes BG data to basal data. This is useful if the user is confident all carb intake events were logged and the user's basal rates are far enough from reality that time periods are being incorrectly categorized as UAM events instead of basal events.

### DIFF
--- a/bin/oref0-autotune-prep.js
+++ b/bin/oref0-autotune-prep.js
@@ -46,9 +46,12 @@ if (!module.parent) {
         process.exit(1);
     }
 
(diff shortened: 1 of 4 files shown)


drnoname82:
Great idea! Tested and works as expected, results are very useful for us.

drnoname82:
We have been running this branch for several days now without any issues, so thumbs up for merging to dev.

PR #909, opened 2018-01-24
Update oref0-setup.sh for the edgecase of failed wget

tepidjuice:
remove existing `bluez-5.48.tar.gz` in case of failed wget so subsequent commands run.

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -637,7 +637,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
         fi
         bluetoothdversioncompare=$(awk 'BEGIN{ print "'$bluetoothdversion'"<"'$bluetoothdminversion'" }')
         if [ "$bluetoothdversioncompare" -eq 1 ]; then
-            cd $HOME/src/ && wget -4 https://www.kernel.org/pub/linux/bluetooth/bluez-5.48.tar.gz && tar xvfz bluez-5.48.tar.gz || die "Couldn't download bluez"
+            cd $HOME/src/ && rm -r bluez-5.48.tar.gz && wget -4 https://www.kernel.org/pub/linux/bluetooth/bluez-5.48.tar.gz && tar xvfz bluez-5.48.tar.gz || die "Couldn't download bluez"
             killall bluetoothd &>/dev/null #Kill current running version if its out of date and we are updating it
             cd $HOME/src/bluez-5.48 && ./configure --disable-systemd && make || die "Couldn't make bluez"
             killall bluetoothd &>/dev/null #Kill current running version if its out of date and we are updating it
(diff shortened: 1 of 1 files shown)


bin/oref0-setup.sh
```diff
@@ -637,7 +637,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
         fi
         bluetoothdversioncompare=$(awk 'BEGIN{ print "'$bluetoothdversion'"<"'$bluetoothdminversion'" }')
         if [ "$bluetoothdversioncompare" -eq 1 ]; then
-            cd $HOME/src/ && wget -4 https://www.kernel.org/pub/linux/bluetooth/bluez-5.48.tar.gz && tar xvfz bluez-5.48.tar.gz || die "Couldn't download bluez"
+            cd $HOME/src/ && rm -rf bluez-5.48.tar.gz && wget -4 https://www.kernel.org/pub/linux/bluetooth/bluez-5.48.tar.gz && tar xvfz bluez-5.48.tar.gz || die "Couldn't download bluez"
```
scottleibrand: Does `rm -rf` return `true` if the file is not found?  May want to make `cd $HOME/src/ && rm -rf bluez-5.48.tar.gz` its own line before `cd $HOME/src/ && wget` etc. to make sure the `wget` doesn't incorrectly depend on the `rm`?

tepidjuice: I think both options are fine.

rm -rf should fail silently.

For example `rm -rf this-dir-does-not-exist && echo success` returns `success`
However `rm -rf this-dir-does-not-exist && echo success` returns `rm: cannot remove 'this-dir-does-not-exist': No such file or directory`

PieterGit: If you specify the output file using the -O option it will overwrite any existing file.
Or we could use `curl http://server/folder/file1.html > file1.html`

Source: https://serverfault.com/questions/171369/how-to-allow-wget-to-overwrite-files or https://stackoverflow.com/questions/30418188/how-to-force-wget-to-overwrite-an-existing-file-
ignoring-timestamp

I think that's cleaner than removing a file that is non existing for most users.

tepidjuice: I agree overwriting is a cleaner option.

scottleibrand: I fixed this for golang (in the go-explorer-hat branch) by adding `-c` to wget, if you'd like to do the same for bluez.

tepidjuice: Updated with the `-c` flag. It's definitely a cleaner option. However I'm not familiar with the behaviour of this flag. To prevent silent failures like I experienced would it be better to use `-O` as suggested by @PieterGit?

scottleibrand:
The -c flag continues a partial download. The -O flag forces it to output to a named file, but would still re-download the file every time.

PR #912, opened 2018-01-29
Use mdt for button command

alimhassam:


bin/oref0-pump-loop.sh
```diff
@@ -311,8 +311,9 @@ function smb_bolus {
     && if (grep -q '"units":' enact/smb-suggested.json 2>&3); then
         # press ESC three times on the pump to exit Bolus Wizard before SMBing, to help prevent A52 errors
         echo -n "Sending ESC ESC ESC to exit any open menus before SMBing: "
-        try_return timerun openaps use pump press_keys esc esc esc | jq .completed | grep true \
-	&& ( try_return timerun mdt bolus enact/smb-suggested.json && jq '.  + {"received": true}' enact/smb-suggested.json > enact/bolused.json ) | tail -1 \
+        #try_return timerun openaps use pump press_keys esc esc esc | jq .completed | grep true \
+        mdt -f internal button esc esc esc \
```
alimhassam: should have piped stderr to null here.

PR #917, opened 2018-02-03
Ignore time travel

scottleibrand:
When traveling westward and moving pump clock back in time, the pumphistory-merge method causes oref0 to use "future" boluses in calculating predBGs, which can result in persistent misdosing of insulin.  To prevent that, this PR changes oref0-calculate-iob to parse the pumphistory-zoned and pumphistory-24h-zoned records directly, and ignore out-of-order records. As a result, a timezone change can still cause dosing history to be ignored, (which oref0 fairly quickly adjusts to based on deviations), but prevents any persistent misdosing beyond that.

There are some TODOs in this code for further work here.  Eventually it'd be best if we actually parse pump time change records to understand the actual time of all events in the history, but this should prevent any serious issues from timezone changes until someone takes the time to do so.

### DIFF
--- a/bin/oref0-calculate-iob.js
+++ b/bin/oref0-calculate-iob.js
@@ -20,7 +20,7 @@
 
 var generate = require('oref0/lib/iob');
 function usage ( ) {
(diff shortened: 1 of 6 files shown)


scottleibrand:
There's now a more comprehensive solution in place in the go-explorer-hat branch, so I'm going to close this in favor of waiting to merge that.

PR #919, opened 2018-02-05
Fix clock sync so that it actually uses 90s

dcacklam:
The clock sync 'says' it is checking to see if the pump is within 90s of the rig.

However it actually uses 55s, and the error if it is out-of-sync says 55s as well. 

This causes issues on slower rigs (Pi Zero W).

Fixed.


dcacklam:
Further re-done. Closing.

PR #922, opened 2018-02-10
Changed suspend_zeros_iob default to true.

jpcunningh:
After successfully using this feature for a while, I recommend we make it the default behavior.


jpcunningh:
Before merging this, we need to get the suspends handled all the way through nightscout db and autotune so autotune doesn't account for more IOB thanks really there due to a suspend.

jpcunningh:
I verified it makes it all the way through NS and is handled correctly by Autotune. No changes were necessary.

inventor96:
@jpcunningh Should this show up in Nightscout as a 0 temp basal?

jpcunningh:
No, Nightscout's basal rendering code does not recognize the pump suspend events as a 0 temp basal.

PR #924, opened 2018-02-11
Fix handling of dates so that we are not comparing the time the pump-…

dcacklam:
Fix date-comparison routine so that it captures the date to a variable immediately after pulling the device-clock date, rather than comparing what the clock *was* set to when the json file was written to disk, to the time it is when the test executes. 

This reduces the number of erroneous pump-set calls when CPU load is high. 

Also, fix set-device-clocks so that it doesn't try and set the 'clock' on xdrip CGM.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -156,12 +156,12 @@ function smb_reservoir_before {
     try_fail timerun openaps report invoke monitor/clock.json monitor/clock-zoned.json 2>&3 >&4 | tail -1
     echo -n "Checking pump clock: "
     (cat monitor/clock-zoned.json; echo) | tr -d '\n'
-    echo -n " is within 90s of current time: " && date
+    echo -n " is within 55s of current time: " && date
     let DTG_DIFFERENCE=$CURRENT_DTG-$(date +%s -d $(cat monitor/clock-zoned.json | sed 's/"//g'))
     DTG_DIFFERENCE=${DTG_DIFFERENCE/#-/}
-    if [ "$DTG_DIFFERENCE" -gt "90" ]
+    if [ "$DTG_DIFFERENCE" -gt "55" ]
     then
-	echo Pump clock is more than 90s off: attempting to reset it
+	echo Pump clock is more than 55s off: attempting to reset it
         timerun oref0-set-device-clocks
         # Refresh current-date, since we just reset device clock.
         $CURRENT_DTG=$(date +%s)
@@ -172,6 +172,8 @@ function smb_reservoir_before {
 	then
 		echo "Error: pump clock refresh error / mismatch"
 		fail "$@"
+	else
+		echo "Pump clock is now within 90s of current time. Proceeding."
 	fi
     fi
     find monitor/ -mmin -1 -size +5c | grep -q pumphistory || { echo "Error: pumphistory too old"; fail "$@"; }
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -152,16 +152,28 @@ function smb_reservoir_before {
     # Refresh reservoir.json and pumphistory.json
     try_fail refresh_pumphistory_and_meal
     try_fail cp monitor/reservoir.json monitor/lastreservoir.json
+    CURRENT_DTG=$(date +%s)
     try_fail timerun openaps report invoke monitor/clock.json monitor/clock-zoned.json 2>&3 >&4 | tail -1
     echo -n "Checking pump clock: "
     (cat monitor/clock-zoned.json; echo) | tr -d '\n'
     echo -n " is within 90s of current time: " && date
-    if (( $(bc <<< "$(date +%s -d $(cat monitor/clock-zoned.json | sed 's/"//g')) - $(date +%s)") < -55 )) || (( $(bc <<< "$(date +%s -d $(cat monitor/clock-zoned.json | sed 's/"//g')) - $(date +%s)") > 55 )); then
-        echo Pump clock is more than 55s off: attempting to reset it
+    let DTG_DIFFERENCE=$CURRENT_DTG-$(date +%s -d $(cat monitor/clock-zoned.json | sed 's/"//g'))
+    DTG_DIFFERENCE=${DTG_DIFFERENCE/#-/}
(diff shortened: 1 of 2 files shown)


bin/oref0-pump-loop.sh
```diff
@@ -152,16 +152,28 @@ function smb_reservoir_before {
...
+    CURRENT_DTG=$(date +%s)
     try_fail timerun openaps report invoke monitor/clock.json monitor/clock-zoned.json 2>&3 >&4 | tail -1
     echo -n "Checking pump clock: "
     (cat monitor/clock-zoned.json; echo) | tr -d '\n'
     echo -n " is within 90s of current time: " && date
-    if (( $(bc <<< "$(date +%s -d $(cat monitor/clock-zoned.json | sed 's/"//g')) - $(date +%s)") < -55 )) || (( $(bc <<< "$(date +%s -d $(cat monitor/clock-zoned.json | sed 's/"//g')) - $(date +%s)") > 55 )); then
-        echo Pump clock is more than 55s off: attempting to reset it
+    let DTG_DIFFERENCE=$CURRENT_DTG-$(date +%s -d $(cat monitor/clock-zoned.json | sed 's/"//g'))
+    DTG_DIFFERENCE=${DTG_DIFFERENCE/#-/}
+    if [ "$DTG_DIFFERENCE" -gt "90" ]
```
scottleibrand: This has to be less than 90 to avoid a boundary condition. I chose 55.

scottleibrand: Specifically, we ran into a situation where when it was set to 90 seconds it would not reset the clock but would also feel the following check.

scottleibrand: So instead I had it try to fix the clock any time it is 55s off but keep looping if that failed, up to 90s.

dcacklam: I can't be sure, but I believe that the difference in check-design might fix that...

Because you were obtaining pump-time at the beginning (openaps call) then calling openaps again to zone it, then getting the 'current' time as part of a series of sub-shells in the calculate routine, it was possible for enough time to pass to get a false-fail...

I would put forward that the issue I was experincing with the neutered PiZero was simply that at 500mhz, it was running so slowly that I slipped outside the 55s-vs-90s window on a continuous basis. 

The changes above should eliminate that edge-case unless the system is so incredibly bogged down that it takes 90 seconds between 'date +%s' populating CURRENT_DTG, & openaps-invoke & monitor/clock.json being written to disk.  

Which is logically harder to have happen, than it is for 90s to pass between clock.json being written, and the process of writing out clock-zoned & executing the bc statement with it's sub-shells.

scottleibrand: Ok, that's good, but it gets us back to the original reason for the 55s pump clock set: the time of the last bolus in pumphistory is what's used to determine if it's been long enough to SMB again.  If we relax the enforcement to allow the pump clock 90s off of the rig time (instead of requiring that the offset minus the time it takes to do the check doesn't exceed 90s), then we can be in a situation where we're SMB'ing every 2 minutes, which is probably too fast.  So it's better to try to reset the pump clock if it has drifted more than a minute, even if we still allow the loop to proceed up to a longer limit.

dcacklam: Got it. So check 55s on the first one, 90 on the 2nd is what you would prefer.

scottleibrand:
Thanks. Ready to merge to dev, or so we need any specific testing first?

dcacklam:
Ready to merge to dev.

scottleibrand:
This completely failed testing, so I've reverted it after my first attempt at fixing syntax errors revealed even more syntax errors.

```
Pump clock is more than 55s off: attempting to reset it
Waiting for ntpd to synchronize... OK!
Setting pump time to Wed Feb 14 19:40:45 PST 2018
mmeowlink.exceptions.CommsException: No acknowledgement from pump on wakeup. Is it out of range or is the battery too low?
/usr/local/bin/oref0-pump-loop: line 167: 1518666018=1518666072: command not found
sed: -e expression #1, char 7: unknown option to `s'
date: option requires an argument -- 'd'
Try 'date --help' for more information.
/usr/local/bin/oref0-pump-loop: line 169: let: DTG_DIFFERENCE=1518666018-: syntax error: operand expected (error token is "-")
Error: pump clock refresh error / mismatch
```

```
Pump clock is more than 55s off: attempting to reset it
Waiting for ntpd to synchronize... OK!
Setting pump time to Wed Feb 14 19:48:12 PST 2018
mmeowlink.exceptions.CommsException: No acknowledgement from pump on wakeup. Is it out of range or is the battery too low?
sed: -e expression #1, char 7: unknown option to `s'
date: option requires an argument -- 'd'
Try 'date --help' for more information.
/usr/local/bin/oref0-pump-loop: line 171: [: 1518666519-: integer expression expected
Pump clock is now within 90s of current time. Proceeding.
```

dcacklam:
Those errors are caused by invalid data in monitor/clock-zoned.json (eg, if the pump was out of range & something other than an epoch DTG - or nothing at all - is in that file).

That resulted in no value being presented after the '-' sign in the subtraction operation, and thus a syntax error. 

I've updated the code to detect & handle that. 

Will let you know when I've confirmed it catches the errors & prevents them.

scottleibrand:
Here's the contents of those files at the time.  They were way stale, but not invalid.
```
root@edison-eb5 ~/myopenaps # cat monitor/clock.json
"2018-02-14T17:43:11"root@edison-eb5 ~/myopenaps # cat monitor/clock-zoned.json
"2018-02-14T17:43:11-08:00"root@edison-eb5 ~/myopenaps # ll monitor/clock*
-rw-r--r-- 1 root root 21 Feb 14 17:43 monitor/clock.json
-rw-r--r-- 1 root root 27 Feb 14 17:43 monitor/clock-zoned.json
```

dcacklam:
Got it.

Will test that too...

The actual syntax error was the missing second operand in the subtraction problem, which would result from a falure of something in the $(cat monitor/clock-zoned.json ...) statement...

Under your original code (I copied & pasted the file-extraction code, so it should be the same cat/sed/etc statement), that would just fail the entire section (due to the && s) and I don't know how BC would handle it, under mine... not so much, plus a syntax error...

PR #927, opened 2018-02-15
Store radio_locale in pump.ini

cluckj:
This will allow locale checks in other parts of openAPS to determine which locale (US or WW) the radio is using.


bin/oref0-setup.sh
```diff
@@ -1080,8 +1080,10 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
         cp $HOME/go/src/github.com/ecc1/medtronic/cmd/pumphistory/openaps.jq $HOME/myopenaps/ || die "Couldn't cp openaps.jq"
         #Necessary to "bootstrap" Go commands...
         if [[ $radio_locale =~ ^WW$ ]]; then
+          grep -q radio_locale pump.ini ||  echo "$(< pump.ini)" > pump.ini ; echo "radio_locale=$radio_locale" >> pump.ini
```
scottleibrand: What does `echo "$(< pump.ini)" > pump.ini` do?

cluckj: Nothing much? I lazily copied it from where this was originally done, at https://github.com/openaps/oref0/blob/f89ff2ac15ec739016c9fb389a7e999fc45a0250/bin/oref0-setup.sh#L859

The latter part of the line, `echo "radio_locale=$radio_locale" >> pump.ini` should be enough.

PR #929, opened 2018-02-22
Don't reverse pump history

alimhassam:
Currently the carb history calculation doesn't
work when the pump history is in reverse order.
This should work but doesn't. Remove the reversing until
this is fixed correctly.

Note, i'm not sure if this could have other negative effects, but the bolus wizard carbs work for me with this change.


scottleibrand:
Do we still need this in light of #932 etc?

alimhassam:
No this should be closed.

PR #931, opened 2018-02-23
refactor calculate.js

parenthetic:
* remove unused code
* create separate function for iobCalcExponential
* use consistent variable names in iobCalcBilinear and iobCalcExponential
* add comments and references 
* provide more intuition for bilinear activity calculation
* (deleted a few blank lines in iob.test.js, too)

### DIFF
--- a/lib/iob/calculate.js
+++ b/lib/iob/calculate.js
@@ -26,10 +26,14 @@ function iobCalc(treatment, time, curve, dia, peak, profile) {
 
 function iobCalcBiLinear(treatment, minsAgo, dia) {
     
-    var timeScalar = 3.0 / dia;
-    var peak = 75;  // peak insulin activity, in minutes
-    var end = 180;  // end of insulin activity, in minutes
+    const default_dia = 3.0 // assumed duration of insulin activity, in hours
+    const peak = 75;        // assumed peak insulin activity, in minutes
+    const end = 180;        // assumed end of insulin activity, in minutes
 
+    // scale minsAgo by the ratio of the default dia / the user's dia 
+    // so the calculations for activityContrib and iobContrib work for 
+    // other dia values (and using the constants specified above)
+    var timeScalar = default_dia / dia; 
     var scaled_minsAgo = timeScalar * minsAgo;
 
     var activityContrib = 0;  // insulin used in last minute
--- a/lib/iob/calculate.js
+++ b/lib/iob/calculate.js
@@ -2,7 +2,7 @@ var _ = require('lodash');
 
 function iobCalc(treatment, time, curve, dia, peak, profile) {
 
-    if (treatment.insulin) return {
+    if (treatment.insulin) {
 
         // Calc minutes since bolus (minsAgo)
         if (typeof time === 'undefined') {
--- a/lib/iob/calculate.js
+++ b/lib/iob/calculate.js
@@ -61,15 +61,15 @@ function iobCalcBilinear(treatment, minsAgo, dia) {
 
         activityContrib = treatment.insulin * (slopeUp * scaled_minsAgo);
 
-        let x1 = (scaled_minsAgo / 5) + 1;  // scaled minutes since bolus, pre-peak; divided by 5 to work with coefficients estimated based on 5 minute increments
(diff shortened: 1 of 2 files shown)


lib/iob/calculate.js
```diff
@@ -61,28 +61,26 @@ function iobCalcBiLinear(treatment, time, dia) {
         time = new Date();
     }
 
-    if (treatment.insulin) {
```
scottleibrand: Why do you think this is unnecessary?  Is it being checked elsewhere?

lib/iob/calculate.js
```diff
@@ -16,11 +16,12 @@ function iobCalcExponential(treatment, time, dia, peak, profile) {
 
     var end = dia * 60;  // end of insulin activity, in minutes
  
+    // Use custom peak time if values are valid
     if (profile.useCustomPeakTime && profile.insulinPeakTime !== undefined) {
-        if (profile.insulinPeakTime < 35 || profile.insulinPeakTime > 120) {
-            console.error('Insulin Peak Time is only supported for values between 35 to 120 minutes');
-        } else {
+        if (profile.insulinPeakTime > 35 && profile.insulinPeakTime < 120) {
```
scottleibrand: You've effectively changed the minimum allowed value from 35 to 36.  I doubt anyone is using it, but probably better for consistency to make this a >= 35?

lib/iob/calculate.js
```diff
@@ -73,7 +73,7 @@ function iobCalcBiLinear(treatment, minsAgo, dia) {
         iobContrib = treatment.insulin * ( (-0.001852*x1*x1) + (0.001852*x1) + 1.000000 );
 
     } else if (scaled_minsAgo < end) {
-        activityContrib = treatment.insulin * (2 / dia / 60 - (scaled_minsAgo - peak) * 2 / dia / 60 / (60 * 3 - peak));
+        activityContrib = treatment.insulin * (2 / dia / 60 - (scaled_minsAgo - peak) * 2 / dia / 60 / (end - peak));
```
scottleibrand: Why this change?

scottleibrand:
Added 3 comments on individual commits, since there were too many changes to grok all at once.  GitHub is showing them all as outdated, but they're all still valid AFAICT.

scottleibrand:
Also, the unit tests are failing on your new `let` statements.  Might want to try reverting those and see if that's enough to make them pass?

lib/iob/calculate.js
parenthetic: Previously, the first line in the iobCalc function was: `if (!treatment.insulin) return {}`. At commit c2b5382 I flipped that logic around, though, and placed the entire logic of `iobCalc` inside a `if (treatment.insulin) {}` block.

An alternate could be to add a similar condition block inside the `treatments.forEach(funtion(treatment) {}` block in `total.js`, or we could even place a similar condition within each of the `iobCalcBilinear` and `iobCalcExponential`  functions. Seems like the latter would then be duplicating some logic and the former would be a little too far removed from the actual logic for someone to hold in their head that that condition applies while reading the `iobCalc` logic.

lib/iob/calculate.js
parenthetic: Good point. Should do `<= 120` for consistency, too.

lib/iob/calculate.js
parenthetic: It seems risky to hard code values in a formula, and less easy to understand what the logic is. If `end` were to be changed, this calculation for `activityContrib` would no longer be correct.

In 071bc9f, I set `activityPeak = 2 / dia / 60` and 
`slopeDown = activityPeak / (end - peak)`.

I think  those changes are helpful because then `2 / dia / 60` isn't duplicated in the code several times, and the overall formula is easier to understand.

parenthetic:
Ah, I see that `let` wasn't fully supported until ([node 4.8.7](http://node.green). It seems like node 0.10.29 is what's installed on my rig, but I have a much more recent version installed on my computer (8.9.4). That is to say: my changes passed my unit tests because of the differences in node environments.

I'll change those back to `var` declarations, and hopefully it should all work.

In 2d5ce77, I used the `const` keyword to declare several fixed constants. Basic support for this syntax appears to be supported in node 0.10.48, but am not certain about 0.10.29...

PR #932, opened 2018-02-24
Fast pump history

alimhassam:
Quicker 24h history refresh.
 - This works by maintaining the settings/pumphistory-24h-zoned.json
   file and only adding new recrords since the last refresh every time.

 - For safety the file is reloaded every 6hours or if for any reason it's
   corrupted (jq can't parse it, or can't get a timestamp on the first array
   element).

 - Logging can still be improved, i tried but failed to make it nice but stil
   keep it detailed enough.

 - Another: improvement would be to move this file back to the monitor directory

 - Also as per the large FIXME comment in the code, there could be an issue when
   clock is modified on the pump.

### DIFF
--- a/bin/oref0-ns-loop.sh
+++ b/bin/oref0-ns-loop.sh
@@ -112,7 +112,7 @@ function ns_meal_carbs {
     #openaps report invoke monitor/carbhistory.json >/dev/null
     nightscout ns $NIGHTSCOUT_HOST $API_SECRET carb_history > monitor/carbhistory.json.new
     cat monitor/carbhistory.json.new | jq .[0].carbs | egrep -q [0-9] && mv monitor/carbhistory.json.new monitor/carbhistory.json
-    oref0-meal monitor/pumphistory-merged.json settings/profile.json monitor/clock-zoned.json monitor/glucose.json settings/basal_profile.json monitor/carbhistory.json > monitor/meal.json.new
+    oref0-meal settings/pumphistory-24h-zoned.json settings/profile.json monitor/clock-zoned.json monitor/glucose.json settings/basal_profile.json monitor/carbhistory.json > monitor/meal.json.new
     grep -q COB monitor/meal.json.new && mv monitor/meal.json.new monitor/meal.json
     echo -n "Refreshed carbhistory; COB: "
     grep COB monitor/meal.json | jq .mealCOB
@@ -169,7 +169,7 @@ function upload_recent_treatments {
 
 #nightscout cull-latest-openaps-treatments monitor/pumphistory-zoned.json settings/model.json $(openaps latest-ns-treatment-time) > upload/latest-treatments.json
 function format_latest_nightscout_treatments {
-    nightscout cull-latest-openaps-treatments monitor/pumphistory-zoned.json settings/model.json $(openaps latest-ns-treatment-time) > upload/latest-treatments.json
+    nightscout cull-latest-openaps-treatments settings/pumphistory-24h-zoned.json settings/model.json $(openaps latest-ns-treatment-time) > upload/latest-treatments.json
 }
 
 function check_mdt_upload {
--- a/bin/oref0-ns-loop.sh
+++ b/bin/oref0-ns-loop.sh
@@ -112,7 +112,7 @@ function ns_meal_carbs {
     #openaps report invoke monitor/carbhistory.json >/dev/null
     nightscout ns $NIGHTSCOUT_HOST $API_SECRET carb_history > monitor/carbhistory.json.new
     cat monitor/carbhistory.json.new | jq .[0].carbs | egrep -q [0-9] && mv monitor/carbhistory.json.new monitor/carbhistory.json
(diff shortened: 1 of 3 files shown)


scottleibrand:
Could we simply rm the pumphistory file when pump clock is more than 55s off and we run oref0-set-device-clocks?  Would that fix the FIXME issue, or only if there's been a pumphistory entry since the time change?

alimhassam:
interesting idea, i guess my suggestion in that comment is still a better approach, but your that idea would definitely aliviate the need for it. the downside is that it might cause slowdown for pumps that have regular clock skews, but probably a small price to pay.
I would be okay with either (or both) approaches, depending if how hard it is to get my suggestion to work on the go code.

bin/oref0-pump-loop.sh
```diff
@@ -588,7 +585,7 @@ function calculate_iob {
 
 function invoke_pumphistory_etc {
     check_clock 2>&3 >&4 || return 1
-    read_pumphistory 2>&3 >&4 || return 1
+    read_pumphistory 2>&3  || return 1
```
alimhassam: I wanted to be able to show in the logs when a full refresh is done vs just a pumphistory update, but this change kinda messes up the log a little, not sure what's the right approach.

bin/oref0-pump-loop.sh
```diff
@@ -39,13 +39,13 @@ old_main() {
...
             && refresh_old_pumphistory_enact \
-            && refresh_old_pumphistory_24h \
             && refresh_old_profile \
             && touch /tmp/pump_loop_enacted -r monitor/glucose.json \
             && ( refresh_temp_and_enact || ( smb_verify_status && refresh_temp_and_enact ) ) \
             && refresh_pumphistory_and_enact \
             && refresh_profile \
-            && refresh_pumphistory_24h \
+            && pumphistory_daily_refresh \
+            && touch /tmp/pump_loop_success \
```
alimhassam: oops, is this line needed? i think it was a merge conflict i corrected wrong.

bin/oref0-pump-loop.sh
```diff
@@ -84,7 +84,8 @@ main() {
                     && refresh_temp_and_enact \
                     && refresh_pumphistory_and_enact \
                     && refresh_profile \
-                    && refresh_pumphistory_24h \
+		    && pumphistory_daily_refresh \
+                    && touch /tmp/pump_loop_success \
```
alimhassam: same here.

bin/oref0-ns-loop.sh
```diff
@@ -169,7 +169,7 @@ function upload_recent_treatments {
 
 #nightscout cull-latest-openaps-treatments monitor/pumphistory-zoned.json settings/model.json $(openaps latest-ns-treatment-time) > upload/latest-treatments.json
 function format_latest_nightscout_treatments {
-    nightscout cull-latest-openaps-treatments monitor/pumphistory-zoned.json settings/model.json $(openaps latest-ns-treatment-time) > upload/latest-treatments.json
+    nightscout cull-latest-openaps-treatments settings/pumphistory-24h-zoned.json settings/model.json $(openaps latest-ns-treatment-time) > upload/latest-treatments.json
```
alimhassam: This change is also needed in the openaps ns-loop, but i cannot test it myself as i don't use it.

alimhassam:
Been running this for 2 days as it is on this pr and it works perfectly. I'm updating this with a a fix for the merge issue and small additional improvements.

The suggestion to refresh the history when clock is updated is an improvement, but will only avoid the issue on the rig that updates the clock.

scottleibrand:
We could also delete and refresh the pumphistory whenever there is a clock change record within the last N records...

alimhassam:
i guess we should detect when the last update contains a clock change record then re-do a full update. It will require parsing the history though which brings some unwanted complexity back. 

Maybe we look at this as a follow up, it's not something likely to occur regularly. What do you think?

scottleibrand:
I can do a quick jq check to delete the file and revert to full refresh every loop (which I'm doing now and works fine) until we have something like 5 records after the clock change.  Then we can further optimize from there if needed.

alimhassam:
the openaps.jq will need updating, because it curretnly ignores clock records.

alimhassam:
updates:
 - moved pumphistory-24h-zoned to monitor/
 - removed line commited by error
 - read_full_pumphistory after set-device-clocks

scottleibrand:
per https://github.com/openaps/oref0/pull/936#issuecomment-368388235 @alimhassam  tested #932 on different scenarios the last 2 days, and says it seems to behave correctly:
 * no pumphistory file: triggers a full reload
 * invalid or empty pump history file: triggers a full reload
 * a full reload is triggered approximately every 6h +/- 15 min.

ecc1:
It's easy to include the ChangeTime and NewTime history records in the openaps.jq script; is that sufficient?

scottleibrand:
Yes, we’ll need those regardless of how we end up solving the time change problem. Thanks.

alimhassam:
This pull request is superseeded by https://github.com/openaps/oref0/pull/937 closing it.

PR #934, opened 2018-02-24
Fix bluetoothup

dcacklam:
PR to reduce bluetooth flapping on Pi by not doing the kill bluetoothd if no PSCAN on arm archetectures.

### DIFF
--- a/bin/oref0-bluetoothup.sh
+++ b/bin/oref0-bluetoothup.sh
@@ -6,7 +6,8 @@ if ! ( ps -fC bluetoothd ) ; then
 fi
 
 #Raspberry Pi doesn't keep PSCAN up the way Edison does. Check for ARM CPU (vs x86 on Edison) before executing this bloc
-if ! ( uname -m | grep arm > /dev/null)
+sys_arch=$(uname -m)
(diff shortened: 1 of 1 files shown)


scottleibrand:
What I did in the go-explorer-hat branch is https://github.com/openaps/oref0/commit/cb41fb7b4b9565f5902a763ea02085cce52b15f7

Can you think of any good reasons to use one approach or the other?  I'm fairly agnostic.

dcacklam:
If someone deletes or renames the 'edison' account (or installs a different Linux base that has a different (or no) default account) the passwd-based hardware-ID will break. 

Unless we introduce a different HW platform than Edison & Pi, looking for ARM arch is a sure-fire way to check for Pi hardware.

scottleibrand:
I’d rather support new hardware types (we’ve already started looking into one that might be coming to market soon) than try to support people doing arbitrary unnecessary customization like changing directory or user names.

dcacklam:
In that case your way makes sense...

scottleibrand:
Ok, I just cherry-picked cb41fb7b4b9565f5902a763ea02085cce52b15f7 and pushed it to dev.  Thanks.

PR #935, opened 2018-02-24
Improved preflight

dcacklam:
Efficiency & error handling improvements for preflight function & get-settings.

Reduces number of times we call model.json based on a time period.
Handles errors other-than 512-pump-detected properly.
Attempts to correct invalid or missing settings.json separate from other logic.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -418,12 +418,48 @@ function mdt_get_bg {
 }
 # make sure we can talk to the pump and get a valid model number
 function preflight {
+    # More efficient preflight function.
+    # Checks for invalid/nonexistent model.json & attempts to correct.
+    # Only refreshes model.json on 2hr intervals, reducing the number of openaps calls. 
+    # Returns different errors for '512 pump' vs 'invalid model file', etc
+    # Less use of egrep & shell-outs vs shell-internals. 
+    PF_FAILED=0
     echo -n "Preflight "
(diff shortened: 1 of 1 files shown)


scottleibrand:
One non-obvious reason for the model check in preflight is to give us a single-packet command to use to see if the pump is nearby.  If the pump doesn't respond to that (with one retry), then we shouldn't proceed to do more intensive things like pumphistory pulls with multiple retries, or we might interfere with other rigs that are closer.

I like the code to avoid refreshing model again during get-settings, but I think that only refreshing model every 2 hours would cause more problems than it solves at least in multiple-rig environments, unless there's some other single-packet request/response we could substitute for model in the preflight.

dcacklam:
My logic behind trying to avoid the model call, was trying to cut loop times down in a single-rig environment, where out-of-range just means the loop will fail either way - but one less 'openaps invoke' (of any kind) can chop useful time off the loop length... Which is my main effort (other than bluetooth) as I try to get as stable as I can running 0.61-dev on the pi zero. 

Didn't think about multi-rig. I can see that posing more of an issue than it does on single (eg, on single it fails for out-of-range either way. 

Feel free to cherrypick the get-settings code & leave the preflight mods behind... 

One other thought that I just had, is that the 'get time' routine (Which DOES have to be run every-single-time, to verify the pump being within 90s) could serve the 'are we in range' function too, if done right...

scottleibrand:
I think the underlying speed-up goal has been well addressed in 0.7.0.  If you'd like to PR any of these changes against 0.7.0-dev, feel free, but otherwise I think we can close this.

PR #936, opened 2018-02-26
use just pumphistory-24h-zoned for COB calc; refresh it every 10-30m

scottleibrand:
This is an interim solution to get bolus wizard carb entry working, until we can test out and merge https://github.com/openaps/oref0/pull/932

It's not super efficient, but I've tested it and it works.


alimhassam:
i believe this breaks ns treatment upload, which also was fixed here: https://github.com/openaps/oref0/pull/932 and fixes the "efficiency" as well.

scottleibrand:
Yeah, I just noticed that and pushed a fix.  Planning to test #932 next, but maybe not tonight.

alimhassam:
thanks! FYI i have tested #932 932 on different scenarios the last 2 days, and it seems to behave correctly for me.
- no pumphistory file: triggers a full reload
- invalid or empty pump history file: triggers a full reload
- a full reload is triggered approximatively every 6h  +/- 15 min.

PR #938, opened 2018-02-28
move xdrip cgm to oref0-ns-loop

alimhassam:
pr #752 reverted the change where xdrip cgm users were moved back to use the
openaps ns loop. This doesn't seem needed anymore, and has been working for me for couple of weeks.

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -942,7 +942,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
         # Add module needed for EdisonVoltage to work on jubilinux 0.2.0
         grep iio_basincove_gpadc /etc/modules-load.d/modules.conf || echo iio_basincove_gpadc >> /etc/modules-load.d/modules.conf
     fi
-    if [[ ${CGM,,} =~ "mdt" ]] || [[ ${CGM,,} =~ "xdrip" ]]; then # still need this for the old ns-loop for now
+    if [[ ${CGM,,} =~ "mdt" ]]; then # still need this for the old ns-loop for now
         cd $directory || die "Can't cd $directory"
         for type in edisonbattery; do
             echo importing $type file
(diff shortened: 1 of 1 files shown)


efidoman:
I tested this by changing Logger to only send to xdripAPS (commented out the NS post). The bg data did not get forwarded on to NS by the new oref0-ns-loop. This change cannot be merged until that issue is resolved. I think it means porting the bg updload code from the old ns-loop to the new one.

alimhassam:
The reason this couldn't be merged was because the new `oref0-ns-loop` doesn't upload to NS, whereas it seems like the `openaps ns-loop` did.

We had some discussion about this on gitter in the last couple of weeks on different channels, and i think the best approach is to have each cgm app do it's own uploading. This is already supported by xdrip+ xdrip-js-logger, Lookout, so i think this pr could be merged. 

If another cgm needs openaps to do its uploading it'll have to be re-implemented, but we could just create a new cgm type or a new option in the setup for it, so i think this could be merged in the meantime.

scottleibrand:
Just put out a call for input from anyone who would be affected by deprecating `xdripaps` upload support: https://gitter.im/nightscout/intend-to-bolus?at=5abc3962270d7d37086d8c40

alimhassam:
as discussed on gitter here, this is okay for Lookout https://gitter.im/nightscout/intend-to-bolus?at=5abc4002bb1018b37a3775d2 we'll try to implement backfilling in Lookout.

I'm pretty sure this is also fine for xdrip+ and xdrip-js-logger.  Does that covers most of users who would select xdrip as a cgm type in oref0-setup? Comment here if there are any other concerns.

davidwagaps:
We use G4 share with the Xdrip+ app. Xdrip+ is configured to send data to the rig and to NS. Would this setup be affected?

scottleibrand:
@davidwagaps sounds like xDrip+ is doing all your uploading for you, so you don't need the rig to also do so.  I'm looking for people (if they exist) for whom their CGM's uploader is insufficient and they actually need oref0 to do the uploading.

alimhassam:
@scottleibrand FYI, in the code change in this pr only affect people who select `xdrip` as their cgm in the setup.
everyone else is already on oref0-ns-loop.

alimhassam:
@davidwagaps you wouldn't be affected. Xdrip+ is uploading to ns for you already in that situation. You would only be affected if xdrip+ was configured to upload only to the rig.

drnoname82:
We use xdripaps and I think it is best to let XDrip+ upload the CGM values.

ToreBj:
I use the xdrip option and was wondering what happens when you go offline(loose internet connection). During that time Xdrio+ sends BG data to the rig and all works fine. When internet connection is back, what happens to the NS update(backfill). Will this be the same as before? i.e. Xdrip+ backfilling or Xdripaps to NS? Just want to make sure how this is covered. Probably not a problem?

daveewall:
Scott, isn’t the case of a pure Medtronic Enlite setup requiring oref0 to do the uploading?  That’s what I’m using.

drnoname82:
@ToreBj You can simply enter two REST-API adresses in XDrip+. The CGM data that is uploaded to the rig via xdripaps stays on the rig, while XDrip+ will backfill CGM data to your Nightscout instance.

scottleibrand:
@daveewall MDT CGM will need some attention after the 0.7.0 release, but this PR only affects `xdrip` CGM users for now.

dcacklam:
@drnoname82 

Should also be noted that you may need more than 2 addresses, if you use xdripAPS for offline looping.

Eg, you may need one-each for the various IPs your rig may have while offline (BT tether, offline wifi, etc) because DNS won't work & xdrip doesn't reliably do Bonjour/mDNS

I usually end up with 3-4 addresses (on for direct to NS, one for BT tether, one for wifi tether, and a DNS name for the rig for cases where I am 'soft offline' due to a firewall)....

Notably, I've never used the rig-as-NS-uploader method of xdripjs setup, so I've been doing multi-IP the whole time I've been using OpenAPS

scottleibrand:
Since no one using xDripAPS who requires this functionality has responded, I'm going to go ahead and merge this if no objections.

PR #940, opened 2018-02-28
Fix pumphistory exit code ignored.

alimhassam:
Fix issue where pumphistory exit code wasn't causing the if statement to fail, and loop was continuing with a stale histroy file


alimhassam:
only tested in the normal flow, but i need to reproduce an error somehow.

alimhassam:
I believe this is ready to review and merge

scottleibrand:
Confirmed this fix works in real life as well:
```
Continuing oref0-pump-loop at Thu  1 Mar 19:05:10 PST 2018
Preflight OK. Profile less than 60m old; Profile valid. Pump history update failed. Last record 2018-03-01T18:36:05-08:00
Retrying invoke_pumphistory_etc
Pump history updated through 2018-03-01T19:02:54-08:00; meal.json refreshed
```

PR #941, opened 2018-02-28
Disable upload of BGCapture pump records to NS.

alimhassam:
THis might not be needed and creates unnecessary noise when used with fakeMeter.

### DIFF
--- a/lib/pump.js
+++ b/lib/pump.js
@@ -14,6 +14,7 @@ function translate (treatments) {
       case 'BasalProfileStart':
       case 'ResultDailyTotal':
       case 'BGReceived':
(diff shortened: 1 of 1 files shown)


scottleibrand:
We should figure out how this works for people with real meters connected to their pumps, and people who enter BG readings into the bolus wizard: many of them prefer having those uploaded to NS.

alimhassam:
makes sense, i thought that might be the case. will need to think of a way to make this work on all use cases.

alimhassam:
This has been resolved differently for fakemeter. As per @scottleibrand comment some people like to see bolus wizard bg entry in ns.
Closing the pull request.

PR #943, opened 2018-03-04
Go precompiled setup

alimhassam:
- Give the option to download the go precompiled binaries  …
instead of recompiling and installing all the extra go stuff.

This requires precompiled binaries to be released, but it's an
improvement specially on the pizero where the setup is so long otherwise.

- Run apt-get clean at the end of setup

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -1114,6 +1114,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
           echo "Downloading Go pump binaries from:" $downloadUrl
           wget -qO- $downloadUrl | tar xJv -C $HOME/go/bin || die "Couldn't download and extract Go pump binaries"
           echo "Installing Go pump binaries ..."
+          cp $HOME/go/bin/openaps.jq $HOME/myopenaps/ || die "Couldn't cp openaps.jq"
           rsync -rtuv $HOME/go/bin/ /usr/local/bin/ || die "Couldn't rsync go/bin"
           mv /usr/local/bin/mmtune /usr/local/bin/Go-mmtune || die "Couldn't mv mmtune"
         fi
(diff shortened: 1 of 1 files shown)


bin/oref0-setup.sh
```diff
@@ -232,6 +232,13 @@ if [[ -z "$DIR" || -z "$serial" ]]; then
             echo
         fi
     fi
+    read -p "Would you like to [D]ownload precompiled go pump communication library or build them from [S]ource? [D]/S " -r
+    buildgofromsource=false
+    if [[ $REPLY =~ ^[Ss]$ ]]; then
+      buildgofromsource=true
+      echo "Building go libs from source"
+    fi
```
scottleibrand: Should probably have an else clause here acknowledging when someone selects Download.

bin/oref0-setup.sh
```diff
@@ -232,6 +232,13 @@ if [[ -z "$DIR" || -z "$serial" ]]; then
             echo
         fi
     fi
+    read -p "Would you like to [D]ownload precompiled go pump communication library or build them from [S]ource? [D]/S " -r
```
scottleibrand: The Go language should be capitalized any time we’re not referring to an actual lowercase command / file.

alimhassam:
As you prefer, if you want to do it, go ahead, otherwise i'll let you finish reviewing and fix these in ~24h. Ideally, would be good to have  @ecc1 to generate a new set of binaries

alimhassam:
I'm testing this, i think what i did here might break the dexcom stuff, that i don't use, @scottleibrand will you be able to look at that part?

alimhassam:
not a perfect test, but i tested on pi, by running the setup one an already running aps. It correctly downloaded and copied the binaries to `/usr/local/bin` and is running fine even after i deleted `/usr/local/go`  will need to test a fresh install also.

PR #944, opened 2018-03-04
openaps-menu status needs profile.json

alimhassam:
openaps-menu needs profile.json to be able to determine if
it needs to convert bgs to mmol, pass it as argument.

also see: https://github.com/openaps/openaps-menu/pull/9

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -118,7 +118,7 @@ main() {
 function update_display {
     # TODO: install this globally
     if [ -e /root/src/openaps-menu/scripts/status.js ]; then
-        node /root/src/openaps-menu/scripts/status.js
+        node /root/src/openaps-menu/scripts/status.js ~/myopenaps/settings/profile.json
     fi
 }
 


bin/oref0-pump-loop.sh
```diff
@@ -118,7 +118,7 @@ main() {
 function update_display {
     # TODO: install this globally
     if [ -e /root/src/openaps-menu/scripts/status.js ]; then
-        node /root/src/openaps-menu/scripts/status.js
+        node /root/src/openaps-menu/scripts/status.js ~/myopenaps/settings/profile.json
```
scottleibrand: All other references to stuff in ~/myopenaps/ use relative paths, as oref0-pump-loop is run from inside ~/myopenaps/ - so I think we can change this to just be `settings/profile.json`.  Want to give that a shot?

alimhassam: you're right, i think i did this wrong in the menu code, but i also noticed that the menu goes to get the files directly instead of taking them as parameters, so i'll do the change that way instead. Will close this pr.

alimhassam:
This is not needed for now, i will follow the existing approach in openaps-menu to keep the consistent style.

PR #945, opened 2018-03-05
Remove i, x, and start from isf guard check

tmecklem:
Since only `offset` is used in calculating which isf is effective for the given
`timestamp`, these other fields add unnecessary confusion to the profile data.

Removing them here allows future removal in decocare and other profile.json
sources. Cleaner (and less) code, cleaner data, easier mental model :)

### DIFF
--- a/lib/oref0-setup/bg_targets_raw.json
+++ b/lib/oref0-setup/bg_targets_raw.json
@@ -1,30 +1,24 @@
 {
-  "units": "mg/dL", 
+  "units": "mg/dL",
   "targets": [
     {
-      "high": 120, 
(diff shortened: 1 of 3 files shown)


scottleibrand:
@tmecklem we're in the process of entirely deprecating decocare for oref0 0.7.0, to support the Raspberry Pi Zero (with the Explorer HAT).  Can you check the contents of the go-explorer-hat branch and see if this cleanup is still needed there?  If so, that's probably the place to do it.

tmecklem:
Changed the PR to `go-explorer-hat` branch.

PR #947, opened 2018-03-07
offline monitor - use xdrip bg data if more recent than monitor

zjedr:
glucose file check to use xdrip glucose data if newer than monitor glucose data

### DIFF
--- a/www/app.py
+++ b/www/app.py
@@ -44,7 +44,7 @@ def enacted():
 
 @app.route("/glucose")
 def glucose():
-    if os.path.isfile("/root/myopenaps/xdrip/glucose.json"):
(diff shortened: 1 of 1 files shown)


www/app.py
```diff
@@ -44,7 +44,7 @@ def enacted():
 
 @app.route("/glucose")
 def glucose():
-    if os.path.isfile("/root/myopenaps/xdrip/glucose.json"):
+    if os.path.getmtime("/root/myopenaps/xdrip/glucose.json") > os.path.getmtime("/root/myopenaps/monitor/glucose.json"):
         json_url = os.path.join("/root/myopenaps/xdrip/glucose.json")
```
scottleibrand: Why is this even necessary?  Doesn't monitor/glucose.json always have the latest CGM data from xdrip or NS?

scottleibrand: Ok, that's a pretty reasonable one as workarounds go.  :-)

PR #955, opened 2018-03-19
Use Current Time Input Consistently in IOB Lib

jpcunningh:
A couple of places in the IOB library used the current clock time instead of the time provided by the inputs. This PR changes those places to use the input clock and updates the tests iob module to account for the change.

### DIFF
--- a/tests/iob.test.js
+++ b/tests/iob.test.js
@@ -608,9 +608,6 @@ describe('IOB', function() {
 
     it('should calculate IOB with Temp Basals that overlap each other', function() {
 
-        var nowDate = new Date();
-        var now = Date.now();
-
         var basalprofile = [{
             'i': 0,
             'start': '00:00:00',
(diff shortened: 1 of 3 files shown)


scottleibrand:
In addition to fixing the unit tests, this is going to require some pretty extensive real-world testing to make sure it doesn’t introduce any change in behavior. 

Is there any current misbehavior you’re trying to fix?

jpcunningh:
We had an event that over a 5 minute period, IOB changed by almost a unit
when it should not have. IOB went negative which caused more insulin to be
delivered than needed. It appeared that something time based had happened.

Running customized tests to attempt to recreate the event revealed the time
passed in is not always used for current time which was creating unexpected
tests results.

I don't think the event we saw is related to this PR change because I
haven't recreated it, yet.

scottleibrand:
Ok, please LMK once you (and preferably someone else) have been looping successfully on this branch for a week or two without issue.

jpcunningh:
We've been on it now for 2 weeks with no issues to note. These have probably been our best 2 weeks so far, but I doubt that has anything to do with this PR.

scottleibrand:
1 failing

  1) IOB should calculate IOB with Temp Basals that overlap each other:
     AssertionError: expected 0.484 to be above 0.484

tests/iob.test.js
```diff
@@ -677,18 +700,15 @@ describe('IOB', function() {
...
 
         var hourLaterInputs = inputs;
-        hourLaterInputs.clock = moment('2016-06-13 00:30:00.000'); //new Date(now + (30 * 60 * 1000)).toISOString();
-        var hourLater = require('../lib/iob')(hourLaterInputs)[0];
+        hourLaterInputs.clock = moment('2016-06-13 00:30:00.000');
+        var hourLaterWithOverlap = require('../lib/iob')(hourLaterInputs)[0];
 
-        hourLater.iob.should.be.lessThan(0.5);
-        hourLater.iob.should.be.greaterThan(0.45);
+        hourLater.iob.should.be.greaterThan(hourLaterWithOverlap.iob);
```
scottleibrand: Is this correct?  Or should it be `>=` rather than `>`?

jpcunningh: It passes on my VM, so after digging into the iob code, I changed it to test greaterThan the original IOB - 0.5 to account for variations.

PR #959, opened 2018-03-24
fix jq -e usage with empty files.

alimhassam:
jq --exit-status always returns a 0 exit code when the
input is empty. This breaks a couple of usage.
Verify that the file exists and is not empty first.
Also remove unnecessary calls to cat, jq could take the input
file as an argument.


alimhassam:
What do you mean by always? 
I'm not sure how you want to verify this?
Is there any concern that jq is misbehaving? 
Or did I misunderstand the doc? 
From my understanding both forms are equivalent but this way avoids a call to cat and avoid having to worry about exit codes and pipes etc..

alimhassam:
I had tested the jq command from the command line and it gives the right result so hopefully that might be enough?
The main fix here is to avoid passing empty file to jq do I could revert the rest if it's controversial.

Issue #961, opened 2018-03-26
Log rotate needs to be set to Daily vs. weekly


Sophieinthecloud:
Devices are filling up because automated change of log rotation is not being set to daily instead of default weekly.  Need to automate this in setup script.  Thanks!

Kdisimone:
Opps...was logged in as someone else there...actually is my issue.  Left over from the Looping get-together yesterday.  Sorry!

danamlewis:
Resolved and will be pushed to master with 0.6.1 release

PR #962, opened 2018-03-29
Fix log rotate for those who don't use boostrap

danamlewis:
Did some sleuthing and found that the log rotate change happens in bootstrap script. But, not everyone is using bootstrap when updating or fixing rigs, and thus more people are reporting issues with logrotate being at weekly, and filling up logs in the course of days in some cases (https://github.com/openaps/oref0/issues/961) 

This adds the logrotate changes automatically in oref0-setup. 

(once patch-released to master, we can remove the manual notes from the docs)

This PR also currently adds a change to one of the shortcut commands; and also removes `update_config=1` from being added to the wpa_supplicant.conf as that is reported to cause several people's problems with network switching.


danamlewis:
This needs testing before we can approve for a merge and an 0.6.1 patch release. Please test by pulling the "fix-log-rotate" branch, and comment here on your testing setup & any details, including verifying by checking `vi /etc/logrotate.conf` that it is correctly set to daily instead of weekly after your setup. You should test a setup with both from scratch using bootstrap to setup the rig; and also from scratch but not using bootstrap to do wifi, and test having oref0-setup change the logrotate automatically. Thanks!

Kdisimone:
The last five rigs I have used bootstrap script and the log rotate file was not updated from weekly. So it’s not just tied to not-using-bootstrap. 

If the fix can be in setup script, I suppose that would address whether or not someone did bootstrap though.

danamlewis:
Ok. This should resolve it for any type of setup, though, since everyone should be using oref0-setup. Thanks for helping test!

BergFulton:
Pulled branch down & installed on 4/8/2018. After running setup script and checking preferences and confirming I was running the correct branch, I checked `logrotate.conf` and log rotate is set to `weekly`.

Kdisimone:
@bergfulton so it didn’t turn to daily automatically?

BergFulton:
@Kdisimone sorry for the late reply (and my bird brain misunderstanding)- no it looks like it was still set to `weekly`. 

I ended up getting out of loop today again for the same reason (device full). I went back in to the file and edited it manually.

Kdisimone:
can we edit the code comment above Scott's recent commit line there?  The bootstrap script is also letting the weekly logrotate stick around...so this isn't just an issue for people who skip bootstrap.  (but if the setup-script catches it, should fix for all).  Thanks

scottleibrand:
I left it to allow oref0-setup to fix anyone who had changed it to `daily` manually.

Kdisimone:
sorry, line 591 is the one I'm talking about

Kdisimone:
Totally!  thanks!

danamlewis:
Ok. Fixed via @scottleibrand's addition, and mine, so ready for more testers to check this again. @scottleibrand, can you make sure to make the same change to dev as well?

scottleibrand:
Merged via a18e84c4ee027b201bb6985d9b51f458ac374a7b

danamlewis:
^ we need additional testing on this (fix-log-rotate branch). 1) ensuring logrotate is changed correctly automatically after setup 2) ensuring update_config=1 does not appear in wpa_supplicant

PR #964, opened 2018-04-10
backport carelink support by supressing mmtune execution

juehv:
added two if statements to supress mmtune execution if carelink stick is used. 
Users have to add the "line radio_type = carelink" to pump.ini file.
Modification is tested by d3xr and myself.


scottleibrand:
Will need to retarget this to dev, and then I think it can be merged.

juehv:
Ok did that :)

PR #974, opened 2018-04-16
make maxUAMSMBBasalMinutes a preference

scottleibrand:
Currently, SMB is limited to delivering 30m worth of basal at a time unless IOB is insufficient to cover COB (generally after meal carbs have been entered, but only a partial bolus or no manual bolus has been given).  Some people have found it necessary to make SMB more aggressive even in the absence of meal announcements, so this provides an optional preference to control SMB's maximum bolus in the UAM case similarly to how maxSMBBasalMinutes controls it in the announced-meal case.

Special thanks to @dcacklam for identifying this as a problem and testing a similar fix.


danamlewis:
Code-wise this looks fine to me. Will merge for part of 0.6.1 testing tomorrow if there are no objections.

PR #976, opened 2018-04-17
avoid filling logs with DIA warning, since we override to 5h now anyway

scottleibrand:


danamlewis:
LGTM: let's merge to dev and include in 0.6.1 testing

PR #978, opened 2018-04-22
Make logrotate in cron run hourly instead of daily

jimrandomh:
The frequency of log-file rotation is controlled by config files, but
also by when logrotate is run by cron. Log-rotation is important to
prevent running out of disk space; since the log files are extremely
compressible, but can grow quickly, a day or two's worth of logs can
potentially deplete the disk space on an Intel Edison, causing problems.
Past commits tried to address this by changing the log-rotation schedule
in logrotate's config files, but it didn't work because the process that
*reads* the config files still only ran once per day. This changes that
from daily to hourly.


scottleibrand:
We should also cherry-pick this commute over to dev and PR to there too.

PR #980, opened 2018-04-22
Symlink instead of copy openaps.jq

cluckj:
Will ensure openaps.jq will be updated when the medtronic repository is updated.


scottleibrand:
Any idea what the Travis failure is, or should I dig in?

PR #981, opened 2018-04-22
Try three times to get a Bluetooth IP address

jaylagorio:
When my phone is in and out of service several times in as many minutes (between subway stations, for example) it doesn't assign an IP address to the device when it tries to connect via Bluetooth. Once back in a more stable environment it takes a while before it tries to get on the Internet via Bluetooth again. This change, tested over several weeks, significantly reduces the time offline during and after using the subway.


GlennCoco:
Great idea @jaylagorio, I have the same issue on airplanes.

jaylagorio:
Thanks, guys!

PR #983, opened 2018-04-22
Bash refactoring: a file for common functions

jimrandomh:
Currently there is a fair amount of redundancy between the shell scripts in oref0, because if a function could be shared between scripts but isn't part of the public interface and isn't worth the overhead of its
own separate script, there wasn't anywhere to put it. This introduces a file oref0-bash-common-functions.sh, for stuff that's shared between scripts. This commit a couple existing functions there, "die", "highload" and "overtemp", which were previously copy-pasted into multiple places. ("Overtemp" had gotten out of sync, with some of the places updates for Pi and others not.)

This also introduces "usage" and "print_usage", which handle detecting the --help flag, and rewrites the help-text handling of most of the scripts to use them; and introduces functions is_edison and is_pi, which handle the idiom of checking /etc/passwd for "edison" and "pi" accounts to find out what type of rig we're running on.

### DIFF
--- a/bin/oref0-bash-common-functions.sh
+++ b/bin/oref0-bash-common-functions.sh
@@ -1,0 +1,76 @@
+#!/bin/echo This file should be source'd from another script, not run directly:
+#
+# Common functions for shell script components of oref0.
+
+# Set $self to the name the currently-executing script was run as. This is usually
+# used in help messages.
+self=$(basename $0)
+
+PREFERENCES_FILE="preferences.json"
+
+
+function overtemp {
+    # check for CPU temperature above 85°C
+    if is_pi; then
+        TEMPERATURE=`cat /sys/class/thermal/thermal_zone0/temp`
+        TEMPERATURE=`echo -n ${TEMPERATURE:0:2}; echo -n .; echo -n ${TEMPERATURE:2}`
+        echo $TEMPERATURE | awk '$NF > 70' | grep input \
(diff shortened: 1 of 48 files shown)


bin/oref0-bash-common-functions.sh
```diff
@@ -0,0 +1,98 @@
...
+# Print the program's help text, as previously set by usage(). This would
+# typically be used after detecting invalid arguments, and followed by "exit 1".
+print_usage () {
+    echo "$HELP_TEXT"
+}
+
+# Check that the current working directory is the myopenaps directory; if it
+# isn't, print a message to stderr and exit with status 1 (failure). We assume
+# we're in the right directory if there's a file named "openaps.ini" here.
+assert_pwd_is_myopenaps () {
```
scottleibrand: Should we rename this to be `assert_cwd_contains_ini` or similar to reflect what we're actually checking?

jimrandomh: Either seems fine. My original thinking was that semantically we're going for ~/myopenaps, and that the presence of openaps.ini is an implementation detail.  The existence of ~/myopenaps-cgm-loop means the difference might sometimes be relevant; but mainly I was thinking of this as a defensive-programming check for the various things in crontab etc that should start with a cd.

scottleibrand:
This is fully incorporated in #994 and #1030, right?

PR #984, opened 2018-04-22
 Unit test to syntax-check script and source files

jimrandomh:
This adds a unit test which goes through the source (everything in
oref0/bin and oref0/lib) and attempts to compile/parse each Python,
Javascript and shell file, failing the test if there are syntax errors.
For Python, this is done with Python 3, regardless of whether the file
specifies Python 2 or Python 3 (currently there are some files using
each). Since Python 3 is stricter, some files were changed to pass
Python 3's stricter syntax rules, in particular, converting mixed spaces
and tabs to all spaces, and putting parentheses on print() statements.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: node_js
 sudo: false
 node_js:
-  - "0.10"
-  - "0.12"
+  - "8.11.1"
 script: make travis
 after_script: make report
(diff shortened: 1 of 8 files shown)


alimhassam:
Seems like some checks are failing on the script itself

jimrandomh:
The problem appears to be that Travis is running a very old version of node (0.12), which doesn't have the string.endsWith function.

jimrandomh:
It looks like this branch is already depending on nodejs 8 (see openaps-packages.sh), so this makes more sense to fix as a Travis CI configuration issue rather than as a backport-to-old-nodejs issue.

PR #990, opened 2018-04-27
Fix typo in filename of /tmp/autosens-completed

jimrandomh:
Fix a typo that would cause completed_recently in oref0-autosens-loop.sh to always be false. This effectively disabled a feature that was supposed to make autosens recalculation happen less frequently when under high load.


scottleibrand:
Thanks.  Merged, and cherry-picked to dev as well.

PR #992, opened 2018-05-02
autoconfirm apt-get

dramageek:
Add -y flags to the rest of the apt-get commands to bypass y/n prompts.

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -195,7 +195,7 @@ if [[ -z "$DIR" || -z "$serial" ]]; then
         echo
     fi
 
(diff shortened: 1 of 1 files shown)


bin/oref0-setup.sh
```diff
@@ -195,7 +195,7 @@ if [[ -z "$DIR" || -z "$serial" ]]; then
         echo
     fi
 
-    if grep -qa "Explorer HAT" /proc/device-tree/hat/product 2>/dev/null ; then
```
dramageek: Bah, looks like I fatfingered this out. I didn't intend to change this line.

PR #993, opened 2018-05-02
Don't error when HAT file doesn't exist

cluckj:
Different location this time. Not sure why it wasn't using -q in the first place, either?


scottleibrand:
Probably for the same reason I only added /dev/null to the first instance: I didn't think to check for a second one when I added -q either.  :-)

PR #994, opened 2018-05-03
Refactor lots of things

jimrandomh:
This branch has lots of general refactoring that I've been doing. One tip for reading the diff: rather than reading it merged, read the diff for each commit separately; wherever there's a find/replace (or something similar in spirit to a find/replace), it's separated into its own commit. I've made liberal use of "git rebase" to edit bug/bugfix pairs out of the history, so the commit log should be quite readable.

Major changes in this branch:

 * Crontab now executes only oref0-cron-[interval] scripts, which don't vary based on config
 * Much less use of "openaps alias"
 * Generally more readable shell scripts

This is not the end of the refactoring I'm doing, but further changes (other than fixes to things within this branch) will be in a new branch.


bin/oref0-bash-common-functions.sh
```diff
@@ -0,0 +1,431 @@
...
+
+# Usage: file_is_recent_and_min_size <path> <minutes> <bytes>
+# Return success if the modification time of the file at the given path has a
+# modification time newer than the given number of minutes (default 5) *and* is
+# at least the given number of bytes (default 5). The minimum size is chosen
+# to exclude files like "{}\n" but include just about everything else. Returns
+# failure if the file doesn't exist, is too small, or is old.
+file_is_recent_and_min_size () {
+    BASE_NAME="$(basename $1)"
+    find "$(dirname $1)" -mmin -"${2-5}" -name "$BASE_NAME" -size +${3-5}c |grep -q "$BASE_NAME" >/dev/null 2>&1
```
alimhassam: This is existing, but i'm not sure why we need to search for the BASE_NAME here instead of just look for the file by full name? Also that extra grep seems unnecessary.

bin/oref0-pump-loop.sh
```diff
@@ -262,15 +254,15 @@ function smb_verify_reservoir {
...
     && echo -n ", suggested: " \
-    && jq -r -C -c .reservoir enact/smb-suggested.json | tr -d '\n' \
+    && jq -r -C -c .reservoir enact/smb-suggested.json | nonl \
     && echo -n " and after: " \
     && cat monitor/reservoir.json \
     && (( $(bc <<< "$(< monitor/lastreservoir.json) - $(< monitor/reservoir.json) <= 0.1") )) \
     && (( $(bc <<< "$(< monitor/lastreservoir.json) - $(< monitor/reservoir.json) >= 0") )) \
-    && (( $(bc <<< "$(jq -r .reservoir enact/smb-suggested.json | tr -d '\n') - $(< monitor/reservoir.json) <= 0.1") )) \
-    && (( $(bc <<< "$(jq -r .reservoir enact/smb-suggested.json | tr -d '\n') - $(< monitor/reservoir.json) >= 0") ))
+    && (( $(bc <<< "$(jq -r .reservoir enact/smb-suggested.json | nonl) - $(< monitor/reservoir.json) <= 0.1") )) \
```
alimhassam: I might be wrong, but i think we need `bc -l` for decimal calculations?

alimhassam:
Nice work! Lot's of good changes here IMO! and the commit structure makes it very easy to review. This is touching a lot of files, so could potentially cause conflicts with other changes. Should try to get this tested and merged relatively shortly to minimise the risk?
I only had a partial look at the changes, but i might merge this pr to my rig for live testing.

bin/oref0-bash-common-functions.sh
jimrandomh: `find` is being used in a slightly weird way here; what it does is it recursively enumerates files in `$(dirname $1)` and outputs to stdout the name of every file which matches the `-mmin` and `-name` filters. `find` always returns true, so the `grep` checks whether the file we're looking for was in the output.

bin/oref0-pump-loop.sh
jimrandomh: `bc` uses whatever precision/number of decimal places are in its input. This is actually pretty sketchy; for example `bc <<< "0.5 * 0.5` outputs `0.2`. But in this case it seems to work out okay.

The `-l` option is defined in bc's manpage as "Define the standard math library." So I think we only need this for functions like cosine.

I'm open to suggestions for things to replace `bc` with for simple floating point arithmetic that are usable from shell. Unfortunately `node` won't work, because its interpreter startup time is too slow (about 800ms on my Edison, vs 10ms for bc).

bin/oref0-pump-loop.sh
```diff
@@ -733,7 +733,7 @@ function refresh_profile {
     else
         profileage=$1
     fi
-    find settings/ -mmin -$profileage -size +5c | grep -q settings.json && echo -n "Settings less than $profileage minutes old. " \
+    file_is_recent_and_min_size settings/settings.json $profileage | grep -q settings.json && echo -n "Settings less than $profileage minutes old. " \
```
jimrandomh: This line was incorrect - there's a stray "|grep -q settings.json" left over, which causes it to think the settings always need refreshing regardless of their actual age. Fixed in later commit.

bin/oref0-pump-loop.sh
```diff
@@ -155,7 +155,7 @@ function smb_reservoir_before {
 
 # check if the temp was read more than 5m ago, or has been running more than 10m
 function smb_old_temp {
-    (find monitor/ -mmin +5 -size +5c | grep -q temp_basal && echo temp_basal.json more than 5m old) \
+    (file_is_recent_and_min_size monitor/temp_basal.json && echo temp_basal.json more than 5m old) \
```
jimrandomh: This line was incorrect - it inverts the age part of the test. Reverted in later commit.

bin/oref0-pump-loop.sh
mattkasa: If the version of jq you have includes slurpfile or argfile support, you could theoretically cut the overhead of two or three of the three or four subshells being invoked per line here by replacing the bc calls with jq calls similar to:
```
jq -r "if (.reservoir - \$monitor[] <= 0.1) then 1 else 0 end" --slurpfile monitor monitor/reservoir.json enact/smb-suggested.json
```

scottleibrand:
Ouch, this got away from us.  There have been a lot of changes since this refactoring (although admittedly many of them already existed in the go-dexcom branch at the time of this refactoring), so it was a lot of work to reconcile them all, but I think I finally managed it.  @jimrandomh would you mind giving https://github.com/openaps/oref0/tree/0.7.0-refactor a review and make sure it captures everything correctly, doesn't have any improperly resolved merge conflicts, etc.?

scottleibrand:
What I ended up doing was cherry-picking all the commits from this PR one by one, and resolving the merge conflicts on each one individually before proceeding.

jimrandomh:
Yeah, the changes on this branch were particularly merge-conflict-heavy. Hopefully the partition into legible commits made it easier.

I've... got another one of these (giant sequences of refactoring commits), actually, which I've been holding back to keep the merge-conflicts from getting overwhelming. Now that this one's landed, I'm going to rebase/merge that one, and open another PR for it.

scottleibrand:
Please base that off the new misc-stuff branch, to minimize conflicts with that. Gonna need some help testing all these and getting them merged relatively quickly. Just switched our rigs to the misc-stuff branch and it’s looking good so far, but I haven’t tested oref0-setup yet.

PR #995, opened 2018-05-05
if targets are < 20, assume that they're intended to be mmol/L

scottleibrand:


Tornado-Tim:
Would be good to get this in the 0.6.1 version before its released.

GlennCoco:
Brilliant idea. Get a little frustrated converting mmol to mg/dl every time I want to put a temp target into NS. Are you committing this to dev branch? @scottleibrand

Tornado-Tim:
Patch PR applied to Nightscout also, so you can enter a temp target in mmol: https://github.com/nightscout/cgm-remote-monitor/pull/3549

GlennCoco:
Brilliant Scott. Works! NS doesn’t show it but I can see it working on the rig.

PR #1001, opened 2018-05-13
add sgv.json?count=# route

tynbendad:
support similar to xdrip+, for offline use with watchfaces, etc.

### DIFF
--- a/www/app.py
+++ b/www/app.py
@@ -1,7 +1,7 @@
 import os
 import socket
 
-from flask import Flask, render_template, url_for, json, jsonify
+from flask import Flask, render_template, url_for, json, jsonify, request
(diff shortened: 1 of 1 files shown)


tynbendad:
fyi there are 2 things i didn't yet include since it wasn't obvious how, from https://github.com/NightscoutFoundation/xDrip/blob/master/Documentation/technical/Local_Web_Services.md

1) /pebble endpoint - probably only needed for some pebble watchfaces.
2) units_hint ("The first record will contain a hint as to the units being used, units_hint will be mgdl or mmol to indicate the user's preference but the values will always be sent as mgdl").   i wasn't sure where to get this info.

PR #1011, opened 2018-05-18
Add carelink to pump.ini automatically for carelink installs

danamlewis:
Add carelink to pump.ini automatically for carelink installs instead of forcing people to manually add it.


danamlewis:
Merging based on verbal confirmation from Gitter that this was all that was needed to enable carelink setup

PR #1013, opened 2018-05-18
activate exercise_mode / high_temptarget_raises_sensitivity at 101

scottleibrand:
The current cutoff for activating exercise_mode / high_temptarget_raises_sensitivity is > 110 ( >= 111 ).  This seems to cause confusion over the behavior of temp targets in the 101-110 range.  To simplify things, this PR proposes removing the + 10 threshold, so that exercise_mode / high_temptarget_raises_sensitivity apply proportionally to any temp target > 100 ( >= 101 ).


danamlewis:
Note to clarify: for people with regular targets >100 mg/dL, this would also apply the sensitivity adjustment when a temp target is >100. So: for someone who normally targets 150 but targets 130, for example, prior to a meal, that will actually cause opposite impact. Flagging this to note in the docs that we possibly want to discourage people with such high targets not to use exercise mode or variations of settings if they're not comfortable with targets closer to 100.

scottleibrand:
Any objections to merging this to 0.7.0-dev for testing?

PR #1019, opened 2018-05-22
Fixed currentRate calculation when using autosens.

jpcunningh:
Using the profile_data.current_basal rather than the basal rate for the time of the insulin delivery event creates a scenario where the basal rate used for all of the insulin delivery events is only the current rate rather than the scheduled rate.


scottleibrand:
Looking through the history (https://github.com/openaps/oref0/pull/684/files) it looks like that was simply an ordering error I made when adding the new sensitivityRatio code. I'll go ahead and merge this to dev and 0.7.0-dev for testing

PR #1025, opened 2018-06-01
adding duty cycle feature for carelink + mdt users

juehv:
This update adds a function to enable an optional duty cycle check, to slow down the loop execution.

The intention is two fold: 
First the battery consumtion is reduced drastically (Pump and Pi)  if the loop runs less often.
Second if carelink is used, the loop produces to much communication, 
pretty much killing the communication between the pump and enlite sensors.

In default configuration this feature is off (DUTY_CYCLE=0) and doesn't change the loop execution.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -29,6 +29,7 @@ fi
 
 # main pump-loop
 main() {
+    ckeck_duty_cycle
     prep
     if ! overtemp; then
         echo && echo "Starting oref0-pump-loop at $(date) with $upto30s second wait_for_silence:"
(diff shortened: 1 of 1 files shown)


juehv:
Sure. I replaced the \t with 4 space. Hope that makes the indentation clear (since it looks correct in notepad++ )

Regarding emergency action: If a configurable ammound of time passes without a successful loop, the function calls oref-reset-usb (if enabled) and if that doesn't help, it reboots the system (if enabled). This helped me a lot, since the carelink stick sometimes stops working for no reason...

scottleibrand:
This LGTM.  I'll merge it soon for testing in dev, unless anyone has any comments or wants to test anything standalone first.

PR #973, opened 2018-04-16
0.7.0-dev (Go-based pump comms)

scottleibrand:
Prep for merge of 0.7.0-dev to dev. 

This is a fairly major change, which deprecates the Python-based `openaps` toolkit in favor of @ecc1's new Go-based pump communications libraries.  This dramatically improves the speed and reliability of pump communications, allowing OpenAPS ro run on a Pi Zero W (for example with the new Explorer HATs), but is also a big improvement on any other hardware as well.  Such a foundational change affects a lot of things, so we'll need some widespread testing of this branch by quite a few more folks, in addition to the testing that folks in the https://gitter.im/openaps/hardware-dev channel have been doing in the go-explorer-hat branch.  (Special thanks to @cluckj @alimhassam and @ecc1 for their extensive work on this, and to @zjedr @jpcunningh and @tmecklem for their contributions as well.)

Checkout and install the 0.7.0-dev branch per the directions at http://openaps.readthedocs.io/en/latest/docs/Customize-Iterate/update-your-rig.html#alternative-step-1b-test-a-feature-branch-not-recommended-for-initial-setup

Once you've confirmed everything is still working, we'll also need people to re-run oref0-setup, and even re-flash your rig and set up from scratch, to test all of the new Go-based pump comms stuff (replacing the python-based `openaps` toolkit).

As always, but **especially importantly for this update**, please make sure that you first test on a backup/dev rig, and that you keep running an unmodified 0.6.x rig while testing out all the new stuff until you're completely comfortable with it.

PR #1027, opened 2018-06-06
Added force npm install argument and fixed dexcom go complied binaries not being downloaded

rkresha:
-npm or --npm_install = force will force `npm run global-install` as long as you aren't on the master branch

removed `if` statement around the download of GoLang from google because buildgofromsource shouldn't have any bearing on whether Go is downloaded... was causing issues on a fresh rig/build
#973

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -131,6 +131,10 @@ case $i in
     PUSHOVER_USER="${i#*=}"
     shift # past argument=value
     ;;
+	-npm=*|--npm_install=*)
+	npm_option="${i#*=}"
+	shift
+	;;
     *)
             # unknown option
     echo "Option ${i#*=} unknown"
@@ -568,7 +572,9 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
     cd $HOME/src/oref0
     if git branch | grep "* master"; then
         npm list -g --depth=0 | egrep oref0@0.6.[0] || (echo Installing latest oref0 package && sudo npm install -g oref0)
-    else
+	elif [[ ${npm_option,,} == "force" ]]; then
+		echo Installing latest oref0 from $HOME/src/oref0/ && cd $HOME/src/oref0/ && npm run global-install
+	else
         npm list -g --depth=0 | egrep oref0@0.6.[1-9] || (echo Installing latest oref0 from $HOME/src/oref0/ && cd $HOME/src/oref0/ && npm run global-install)
     fi
 
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -692,6 +692,16 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
             sudo make install || die "Couldn't make install bluez"
             killall bluetoothd &>/dev/null #Kill current running version if its out of date and we are updating it
             sudo cp ./src/bluetoothd /usr/local/bin/ || die "Couldn't install bluez"
+            
+            # Replace all other instances of bluetoothd and bluetoothctl to make sure we are always using the self-compiled version
+            while IFS= read -r bt_location; do 
+                if [[ $($bt_location -v|awk -F': ' '{print $NF}') != "5.48" ]]; then
+                    killall bluetoothd &>/dev/null #Kill current running version if its out of date and we are updating it
+                    sudo cp -p $(find $HOME/src/bluez-5.48 -name $(basename $bt_location)) $bt_location || die "Couldn't replace $(basename $bt_location) in $(dirname $bt_location)"
+                    touch /tmp/reboot-required
+                fi
+            done < <(find / -name "bluetoothctl" -o -name "bluetoothd")
+            
             oref0-bluetoothup
         else
             echo bluez version ${bluetoothdversion} already installed
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -1161,6 +1161,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
             wget -qO- $downloadUrl | tar xJv -C $HOME/go/bin || die "Couldn't download and extract Go dexcom binaries"
         fi
         cp -pruv $HOME/go/bin/* /usr/local/bin/ || die "Couldn't copy go/bin"
+        mv /usr/local/bin/mmtune /usr/local/bin/Go-mmtune || die "Couldn't mv mmtune"
     fi
 
     #if [[ "$ttyport" =~ "spi" ]]; then
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -1052,7 +1052,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
     echo
 
     #Check to see if Explorer HAT is present, and install all necessary stuff
-    if grep -qa "Explorer HAT" /proc/device-tree/hat/product || [[ "$ttyport" =~ "spidev0.0" ]]; then
+    if grep -qa "Explorer HAT" /proc/device-tree/hat/product &> /dev/null || [[ "$ttyport" =~ "spidev0.0" ]]; then
         echo "Looks like you're using an Explorer HAT!"
         echo "Making sure SPI is enabled..."
         sed -i.bak -e "s/#dtparam=spi=on/dtparam=spi=on/" /boot/config.txt
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -1032,6 +1032,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
     echo "export API_SECRET" >> $HOME/.bash_profile
     echo DEXCOM_CGM_ID="$BLE_SERIAL" >> $HOME/.bash_profile
     echo "export DEXCOM_CGM_ID" >> $HOME/.bash_profile
+    echo 
 
     echo
 
@@ -1058,7 +1059,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
 
     # install Go for Explorer Board/HAT
     if [[ "$ttyport" =~ "spidev" ]] || [[ ${CGM,,} =~ "g4-go" ]]; then
-      if $buildgofromsource; then
+      # if $buildgofromsource; then  # Is preventing go from being downloaded, I think the check is invalid in this position since it's not related to building custom go binaries
         source $HOME/.bash_profile
         if go version | grep go1.9.; then
             echo Go already installed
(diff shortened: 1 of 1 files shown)


bin/oref0-setup.sh
```diff
@@ -1052,7 +1059,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
 
     # install Go for Explorer Board/HAT
     if [[ "$ttyport" =~ "spidev" ]] || [[ ${CGM,,} =~ "g4-go" ]]; then
-      if $buildgofromsource; then
+      # if $buildgofromsource; then  # Is preventing go from being downloaded, I think the check is invalid in this position since it's not related to building custom go binaries
```
scottleibrand: No, I think this was correct.  If you're not building anything from source, you don't want to download and install the (huge) go toolchain: you can just download the precompiled binaries.

rkresha: Ok, yeah that makes sense, we just need to adjust how it is downloading the g4 binaries, since it is currently trying to use `go get` which doesn't work if you don't have go installed

rkresha:
Doing an end-to-end test of a fresh flash all the way through and will report back

rkresha:
Looks really good, added some additional bluetoothd and bluetoothctl replacements for issues I was seeing with it using the BT binaries, specifically the g4-loop was picking 5.23 for whatever reason instead of 5.48 .. this last commit makes that a thing of the past.

bin/oref0-setup.sh
```diff
@@ -200,7 +204,7 @@ if [[ -z "$DIR" || -z "$serial" ]]; then
         echo
     fi
 
-    if grep -qa "Explorer HAT" /proc/device-tree/hat/product 2>/dev/null ; then
+    if grep -qa "Explorer HAT" /proc/device-tree/hat/product &>/dev/null ; then
```
scottleibrand: The -q flag already squashes stdout.  The 2>/dev/null was squashing stderr (file does not exist) as well.  Any reason to change that?

bin/oref0-setup.sh
```diff
@@ -686,6 +692,16 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
...
             sudo cp ./src/bluetoothd /usr/local/bin/ || die "Couldn't install bluez"
+            
+            # Replace all other instances of bluetoothd and bluetoothctl to make sure we are always using the self-compiled version
+            while IFS= read -r bt_location; do 
+                if [[ $($bt_location -v|awk -F': ' '{print $NF}') != "5.48" ]]; then
+                    killall bluetoothd &>/dev/null #Kill current running version if its out of date and we are updating it
+                    sudo cp -p $(find $HOME/src/bluez-5.48 -name $(basename $bt_location)) $bt_location || die "Couldn't replace $(basename $bt_location) in $(dirname $bt_location)"
+                    touch /tmp/reboot-required
+                fi
+            done < <(find / -name "bluetoothctl" -o -name "bluetoothd")
```
scottleibrand: This is some advanced bash.  Can you explain what it does?

bin/oref0-setup.sh
```diff
@@ -754,7 +770,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
     # TODO: deprecate g4-upload and g4-local-only
     if [[ ${CGM,,} =~ "g4-upload" ]]; then
         mkdir -p $directory-cgm-loop
-        if ( cd $directory-cgm-loop && ls openaps.ini 2>/dev/null >/dev/null && openaps use -h >/dev/null ); then
+        if ( cd $directory-cgm-loop && ls openaps.ini &>/dev/null && openaps use -h >/dev/null ); then
```
scottleibrand: As above, the intention here is to squash "file not found" errors.

bin/oref0-setup.sh
```diff
@@ -928,7 +944,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
     sudo sysctl -p
 
     # Install EdisonVoltage
-    if egrep -i "edison" /etc/passwd 2>/dev/null; then
+    if egrep -i "edison" /etc/passwd &>/dev/null; then
```
scottleibrand: Why change this?

bin/oref0-setup.sh
```diff
@@ -965,7 +981,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
     echo Running: openaps report add enact/suggested.json text determine-basal shell monitor/iob.json monitor/temp_basal.json monitor/glucose.json settings/profile.json settings/autosens.json monitor/meal.json
     openaps report add enact/suggested.json text determine-basal shell monitor/iob.json monitor/temp_basal.json monitor/glucose.json settings/profile.json settings/autosens.json monitor/meal.json
 
-    if egrep -qi "edison" /etc/passwd 2>/dev/null; then
+    if egrep -qi "edison" /etc/passwd &>/dev/null; then
```
scottleibrand: Why change this?

rkresha:
@scottleibrand I'm good with the current state of this PR. Let me know if you have any questions. -Rob

bin/oref0-setup.sh
```diff
@@ -1100,12 +1117,12 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
           #cd ../mmtune && go install -tags cc111x || die "Couldn't go install mmtune"
           #cd ../pumphistory && go install -tags cc111x || die "Couldn't go install pumphistory"
           #cd ../listen && go install -tags cc111x || die "Couldn't go install listen"
-          rsync -rtuv $HOME/go/bin/ /usr/local/bin/ || die "Couldn't rsync go/bin"
+          cp -pruv $HOME/go/bin/* /usr/local/bin/ || die "Couldn't copy go/bin"
```
scottleibrand: What's the advantage of using cp over rsync?

bin/oref0-setup.sh
```diff
@@ -1100,12 +1117,12 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
...
           mv /usr/local/bin/mmtune /usr/local/bin/Go-mmtune || die "Couldn't mv mmtune"
           ln -sf $HOME/go/src/github.com/ecc1/medtronic/cmd/pumphistory/openaps.jq $directory/ || die "Couldn't softlink openaps.jq"
         else
           arch=arm-spi
-          if egrep -i "edison" /etc/passwd 2>/dev/null; then
+          if egrep -i "edison" /etc/passwd &>/dev/null; then
```
scottleibrand: Why change this?

bin/oref0-setup.sh
```diff
@@ -1115,17 +1132,30 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
           wget -qO- $downloadUrl | tar xJv -C $HOME/go/bin || die "Couldn't download and extract Go pump binaries"
           echo "Installing Go pump binaries ..."
           ln -sf $HOME/go/bin/openaps.jq $directory/ || die "Couldn't softlink openaps.jq"
-          rsync -rtuv $HOME/go/bin/ /usr/local/bin/ || die "Couldn't rsync go/bin"
+          cp -pruv $HOME/go/bin/* /usr/local/bin/ || die "Couldn't copy go/bin"
```
scottleibrand: What's the advantage of using cp over rsync?

bin/oref0-setup.sh
```diff
@@ -1115,17 +1132,30 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
...
+            if egrep -i "edison" /etc/passwd &>/dev/null; then
+                arch=386
+            fi
+            downloadUrl=$(curl -s https://api.github.com/repos/ecc1/dexcom/releases/latest | \
+            jq --raw-output '.assets[] | select(.name | contains("'$arch'")) | .browser_download_url')
+            echo "Downloading Go dexcom binaries from:" $downloadUrl
+            wget -qO- $downloadUrl | tar xJv -C $HOME/go/bin || die "Couldn't download and extract Go dexcom binaries"
         fi
-        rsync -rtuv $HOME/go/bin/ /usr/local/bin/ || die "Couldn't rsync go/bin"
+        cp -pruv $HOME/go/bin/* /usr/local/bin/ || die "Couldn't copy go/bin"
```
scottleibrand: What's the advantage of using cp over rsync?

bin/oref0-setup.sh
```diff
@@ -1197,7 +1227,7 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
         (crontab -l; crontab -l | grep -q "oref0-bluetoothup" || echo '* * * * * ps aux | grep -v grep | grep bash | grep -q "oref0-bluetoothup" || oref0-bluetoothup >> /var/log/openaps/network.log' ) | crontab -
         fi
         # proper shutdown once the EdisonVoltage very low (< 3050mV; 2950 is dead)
-        if egrep -i "edison" /etc/passwd 2>/dev/null; then
+        if egrep -i "edison" /etc/passwd &>/dev/null; then
```
scottleibrand: Why change this?

scottleibrand:
The main change looks good.  Lots of questions about the ancillary stuff inline.

bin/oref0-setup.sh
rkresha: It's dumping it to stdout during the run of the setup, not needed... I originally noticed /etc/passwd coming up a bunch through the run which made me wonder why

bin/oref0-setup.sh
rkresha: consistency, cp is more universal, I had a commit a long while back and we changed other rsync's over to cp

bin/oref0-setup.sh
rkresha: consistency, cp is more universal, I had a commit a long while back and we changed other rsync's over to cp

bin/oref0-setup.sh
rkresha: yep so instead of writing `2>/dev/null >/dev/null`, can be simplified to `&> /dev/null` in bash

scottleibrand: Ah, ok.  I wasn't familiar with that syntax.  These are all good then, thx.

bin/oref0-setup.sh
rkresha: looks for all files named bluetoothd OR bluetoothctl throughtout the filesystem and then runs a `-v` on them to get the version... if they don't match 5.48, it will go and find the same named file under $HOME/src/bluez-5.48 and copy to the location. 

> Example: It finds a bluetoothd file in /home/.rootfs/usr/local/bin/ for instance that reports it is version 5.23, it will go find bluetoothd in $HOME/src/bluez-5.48 and copy it to /home/.rootfs/usr/local/bin/

scottleibrand: Nice.  :-)

scottleibrand: Would it be possible to change the != to a "less than" so that any new version of bluethooth > 5.49 is also allowed?

rkresha:
We'll probably want to put the docs with the update that you don't need to run `npm run global-install` when changing over to the feature branch... just added on `-npm_install=force` ... not sure how to keep tabs on that since this is still a way out

bin/oref0-setup.sh
rkresha: looking into how to accomplish this, bash apparently doesn't like decimals

scottleibrand: How about just a string based sort comparison?

rkresha: I'm commiting the change, I need to run through it again to verify it works as intended, works good as a I spot checked the logic/syntax. Will update with my findings

rkresha: Works like a dream

rkresha:
Hold on ... pump comms aren't working

rkresha:
Alright we're good, I'm pretty sure it wasn't the last two commits, but those two are legitimate adjustments

PR #1029, opened 2018-06-09
post-0.7.0-refactor stuff

scottleibrand:
This PR contains a number of minor changes for 0.7.0:

- Ability to (manually) tune insulinPeakTime and insulinEndTime.  This is not run automatically in cron at this time.
- Re-enable oref0-radio-reboot, to automatically reboot the rig for various health check failures that can be resolved that way
- De-duplicate the PATH variable
- Verify that meal.json is valid, and don't replace it if not.
- Retry SMB pump-loop on failure instead of falling back to basal-only pump-loop
- Retry mmtune if needed, and display output for easier troubleshooting
- Raise autotune's IOB threshold for UAM exclusion from currentBasal to 2 * currentBasal
- Only use the lowest UAM of deviations for autotuning ISF when there's not enough ISF data
- Allow customization of minimum interval between SMBs (hidden, default is still 3 minutes)
- Blinks the cc1110 LEDs once every 10s during wait_for_bg.  It already blinked once per `.` during wait_for_silence, and once per pump comms command otherwise, so you can now tell if the rig is dead if it doesn't blink for more than about a minute.


jaylagorio:
Independent of the other PR, so far I think the only issues I've run into with this branch are the following:

 - I use a meter that sends its result to the pump, which automatically shows up in Nightscout as a red dot. I've noticed those aren't showing up anymore. Full disclosure: I'm not sure if that started with the `0.7.0-dev` branch or with this one based on it.
 - When I get home from using Bluetooth all day I expect Wi-Fi to get picked back up at some point and in previous releases it does reliably, but on this branch unless I reboot the device it doesn't happen. After the reboot it joins up immediately just fine.

scottleibrand:
To debug why the rig isn't switching back to wifi, it'd be helpful to see the network.log output from after you get home when it should be on wifi but isn't releasing Bluetooth.

Regarding the meter BG, which CGM upload method are you using?  Where is the MBG recorded: pumphistory.json or elsewhere?  Where is it normally uploaded to NS: the entries.json or treatments.json collection?

jaylagorio:
I'll get on the network.log output tomorrow when I get home.

The CGM upload right now has switched between `g5-upload` and `xdrip-js`, but the MBG must be have been coming from pumphistory.json. Although my CGM method has changed from `g5-upload` the red dot would get sent up even when I didn't enter the MBG into the Dexcom receiver (and the meter doesn't communicate with that device). The MBGs had to get there from the pump because that was/is the only place they're getting sent to 100% of the time, right?

As for the collection it was dropped into - I can run some queries tomorrow while I'm pulling the network.log file.

jaylagorio:
Filed here: https://github.com/ecc1/medtronic/issues/4

scottleibrand:
Added some comments above.  I think this is ready for review, and could be merged to dev soon if no one sees any objections.

scottleibrand:
Merging this to dev shortly if no objections.

PR #1040, opened 2018-06-24
fix mdt-cgm.json

juehv:
Remove wrong trailing comma in mdt-cgm.json to make setup working again for MDT users.

### DIFF
--- a/lib/oref0-setup/mdt-cgm.json
+++ b/lib/oref0-setup/mdt-cgm.json
@@ -101,5 +101,5 @@
       "json_default": "True"
     },
     "name": "nightscout/glucose.json"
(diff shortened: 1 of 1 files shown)


thecubic:
Could we add some build-integrated testing for situations like this?  If something checked all the `.json` files for well-formedness (pipe to jq, check rc) project-wide there might be an early signal to catch pre-merge.  Even if that can't be integrated, a diagnostic standalone shell script would generate good signal for supportability

scottleibrand:
Please do. 

There is a new bash unit tester that checks all the bash scripts for syntax errors. You could probably add it to that.

PR #1042, opened 2018-06-24
Add xdrip-js as a CGM system option for oref0-setup

jaylagorio:
Adds xdrip-js as a CGM system target.

1. When the CGM option is set to `xdrip-js`, during installation all of the `xdrip` stuff gets executed as well.
1. Added `xdrip-js` to the list of valid CGM targets
1. Causes BlueZ to be installed
1. `-dtx=` or `--dexcom_tx_sn=` assigns `$DEXCOM_CGM_TX_ID` using the command line
1. Asks for the Dexcom transmitter serial during interactive setup only if `xdrip-js` is the CGM system
1. Clones `Logger` (https://github.com/xdrip-js/Logger) into ~/src/ and install only when `xdrip-js` is the CGM system
1. Adds `Logger` directly to `cron` to execute each minute
1. To clarify Dexcom receiver vs. transmitter IDs in the setup script, `$DEXCOM_CGM_ID` is renamed to `$DEXCOM_CGM_RECV_ID` (which still maps to `$BLE_SERIAL`), all of which is split out from `$DEXCOM_CGM_TX_ID`

Sample command line to use xdrip-js as the CGM system with a Pi HAT:

`oref0-setup --dir=/root/myopenaps --serial=###### --cgm=xdrip-js --ns-host=http://nightscout.com --api-secret=xxxx --tty=/dev/spidev0.0 --enable=' autosens meal autotune ' --radio_locale='US' --dexcom_tx_sn=12A34B`

When changing to a new transmitter the easiest way would be for users to rerun setup but with the new transmitter ID, otherwise they could edit `preferences.json` and their `crontab`.

When attempting to upgrade Node 8 using https://github.com/openaps/oref0/commit/9006a6d831095beddeb0774cdc074ce37573ec98 the following error is produced:

`## You appear to be running on ARMv6 hardware. Unfortunately this is not currently supported by the NodeSource Linux distributions. Please use the 'linux-armv6l' binary tarballs available directly from nodejs.org for Node.js 4 and later.`

To get around this `openaps-packages.sh` was modified to check whether an Explorer HAT was present, indicating that installation is on a Pi, and to install via manual download:

```
#if ! sudo apt-get install -y npm; then
# install/upgrade to node 8
if ! nodejs --version | grep 'v8.'; then
    if grep -qa "Explorer HAT" /proc/device-tree/hat/product &>/dev/null ; then
        mkdir $HOME/src/node && cd $HOME/src/node
        wget https://nodejs.org/dist/v8.10.0/node-v8.10.0-linux-armv6l.tar.xz
        tar -xf node-v8.10.0-linux-armv6l.tar.xz
        cd *6l
        sudo cp -R * /usr/local/
    else
        sudo bash -c "curl -sL https://deb.nodesource.com/setup_8.x | bash -" || die "Couldn't setup node 8" 
        sudo apt-get install -y nodejs || die "Couldn't install nodejs" 
    fi
fi
```


bin/oref0-setup.sh
```diff
@@ -18,7 +18,7 @@ source $(dirname $0)/oref0-bash-common-functions.sh || (echo "ERROR: Failed to r
 
 # TODO: deprecate g4-upload and g4-local-only
 usage "$@" <<EOT
-Usage: $self <--dir=directory> <--serial=pump_serial_#> [--tty=/dev/ttySOMETHING] [--max_iob=0] [--ns-host=https://mynightscout.herokuapp.com] [--api-secret=[myplaintextapisecret|token=subjectname-plaintexthashsecret] [--cgm=(G4-upload|G4-local-only|G4-go|G5|MDT|xdrip)] [--bleserial=SM123456] [--blemac=FE:DC:BA:98:76:54] [--btmac=AB:CD:EF:01:23:45] [--enable='autotune'] [--radio_locale=(WW|US)] [--ww_ti_usb_reset=(yes|no)]
+Usage: $self <--dir=directory> <--serial=pump_serial_#> [--tty=/dev/ttySOMETHING] [--max_iob=0] [--ns-host=https://mynightscout.herokuapp.com] [--api-secret=[myplaintextapisecret|token=subjectname-plaintexthashsecret] [--cgm=(G4-upload|G4-local-only|G4-go|G5|MDT|xdrip|xdrip-js)] [--bleserial=SM123456] [--blemac=FE:DC:BA:98:76:54] [--dexcom_tx_sn=12A34B] [--btmac=AB:CD:EF:01:23:45] [--enable='autotune'] [--radio_locale=(WW|US)] [--ww_ti_usb_reset=(yes|no)]
```
jpcunningh: I would much rather this option be `logger` rather than `xdrip-js` so it doesn't get confused with lookout that is also xdrip-js based and make it difficult to integrated Lookout into the install script later.

jaylagorio: I do see where you're coming from, but I'm concerned "logger" isn't very descript. Certainly nothing against the project or the name itself, but xdrip-js is much more immediately recognizable.

efidoman: Will these changes work for both Logger or Lookout? If not, then we probably should consider supporting both if xdrip-js is used. Thoughts?

jpcunningh:
Lookout is primarily geared to using the official calibrated values and
provides GUI-based transmitter status and control.

Logger is primarily geared to using expired transmitters, though I believe
it now supports unexpired usage now, and provides command line status
control.

jaylagorio:
Ah, I didn't understand that difference! Maybe I should have been focussing on Lookout all along anyway. Let me poke around, but the troubled transmitter I'm working through is definitely going to hold things up until I figure out why it's being weird.

scottleibrand:
Let us know when you've decided whether to use xdrip-js Lookout and/or xdrip-js Logger, and are ready for code review and testing.  Looks like there's also now a merge conflict: let us know if you need help resolving that.

jaylagorio:
I think for this first go around I'd like to stick with Logger over Lookout, but in a next iteration I would be open to adding Lookout. I haven't had a chance to work on Lookout like I did with Logger and would want to test at least as rigerously.

As for the conflict, it looks like I can't resolve it myself. But these lines:

```
        if validate_g4share_serial $BLE_SERIAL; then
            (crontab -l; crontab -l | grep -q "DEXCOM_CGM_ID=" || echo DEXCOM_CGM_ID=$BLE_SERIAL) | crontab -
        fi
        (crontab -l; crontab -l | grep -q "PATH=" || echo "PATH=$PATH" ) | crontab -
```

Should be this:

```
        if validate_g4share_serial; then
            (crontab -l; crontab -l | grep -q "DEXCOM_CGM_RECV_ID=" || echo DEXCOM_CGM_RECV_ID=$BLE_SERIAL) | crontab -
        fi
        if validate_g5transmitter_serial; then
            (crontab -l; crontab -l | grep -q "DEXCOM_CGM_TX_ID=" || echo DEXCOM_CGM_TX_ID=$DEXCOM_CGM_TX_ID) | crontab -
        fi
```

jaylagorio:
Rebasing this PR to `misc-stuff` to hopefully make it easier to merge. It looks like `openaps-packages.sh` and `oref0-setup.sh` have 1 conflict each now, but both can be resolved by choosing the `patch-1` branch.

bin/openaps-packages.sh
```diff
@@ -15,9 +15,19 @@ echo 'Acquire::ForceIPv4 "true";' | sudo tee /etc/apt/apt.conf.d/99force-ipv4
...
-    sudo apt-get install -y nodejs || die "Couldn't install nodejs"
+#if ! sudo apt-get install -y npm; then
+# install/upgrade to node 8
+if ! nodejs --version | grep 'v8.'; then
+    if grep -qa "Explorer HAT" /proc/device-tree/hat/product &>/dev/null ; then
+        mkdir $HOME/src/node && cd $HOME/src/node
+        wget https://nodejs.org/dist/v8.10.0/node-v8.10.0-linux-armv6l.tar.xz
+        tar -xf node-v8.10.0-linux-armv6l.tar.xz
+        cd *6l
+        sudo cp -R * /usr/local/
```
scottleibrand: This could do something unintended if one of the above commands fail.  Probably want to chain some of them together with && or similar, or use absolute paths here.

scottleibrand:
@jaylagorio you want to resolve the comments above and the merge conflicts?  Or do you need assistance with that?

jaylagorio:
Can you see my comments under your comment about the `openaps-packages.sh` changes? I can see my comments when I'm logged in but not when I'm not logged in...

scottleibrand:
Nope: https://www.dropbox.com/s/j9tlnodd3jiexnb/Screenshot%202018-08-01%2014.57.34.png?dl=0

jaylagorio:
Huh, not sure I understand what's going on. But since I identified which changes in the setup script should be prefered in the conflict in that file (and Github won't let me resolve the conflict myself or it isn't clear to me how to get around the un-gray the button) I asked was if this was an acceptable alternative to the originally proposed additions to `openaps-packages.sh`:

```
( mkdir $HOME/src/node && cd $HOME/src/node && \
( ( wget https://nodejs.org/dist/v8.10.0/node-v8.10.0-linux-armv6l.tar.xz && echo "Extracting Node..."  && tar -xf node-v8.10.0-linux-armv6l.tar.xz ) || die "Couldn't download and extract Node" ) && \
( ( cd *6l && sudo cp -R * /usr/local/ ) || die "Couldn't copy Node to /usr/local" ) )
```

What do you think?

scottleibrand:
Yeah, that approach looks good.  I just resolved the merge conflicts, added (a slightly simplified version) of the `&&` and `|| die` stuff, and applied a validation fix to the new `validate_g5transmitter_serial` function like we did with `validate_g4share_serial`.  See `git show c05916ee5c4e0315e126ea91797f0f04e9f3938d` for details.

jaylagorio:
Thank you for those mods!

PR #1043, opened 2018-06-24
add prefs for phones that can support advanced bt tethering

tynbendad:
bt_with_wifi: allow bt tether while rig also has a wifi connection.
bt_offline: keep bt tether steady even when there is no internet

not all phones will support these, but with them our samsung galaxy s7 will stay connected to the rig regardless of wifi, so we can always use the offline flask served pages (using android firefox, not chrome, due to its offline support), with or without internet, and the rig will still use wifi (either via its own wifi or the phone's) when available, even when bt tethered.

### DIFF
--- a/bin/oref0-online.sh
+++ b/bin/oref0-online.sh
@@ -40,20 +40,45 @@ main() {
                 ifdown wlan0; ifup wlan0
             fi
         fi
-        # if online via wifi, disconnect BT
-        if has_ip wlan0 && ifconfig | egrep -q "bnep0" >/dev/null; then
-            bt_disconnect $MACs
-            #wifi_dhcp_renew
-        fi
+	    if ! ls preferences.json 2>/dev/null >/dev/null \
+	        || ! cat preferences.json | jq -e .bt_with_wifi >/dev/null; then
+            # if online via wifi, disconnect BT
+            if has_ip wlan0 && ifconfig | egrep -q "bnep0" >/dev/null; then
+                bt_disconnect $MACs
+                #wifi_dhcp_renew
+            fi
+		else
+            if ifconfig | egrep -q "bnep0" >/dev/null; then
+				echo "Preference bt_with_wifi is enabled - have bt iface; leaving enabled."
+        		if ! has_ip wlan0; then
+					echo "Preference bt_with_wifi is enabled - enabling wifi."
+            		wifi_dhcp_renew
+        		fi
+				ifconfig | egrep -A6 "wlan|bnep"
+			else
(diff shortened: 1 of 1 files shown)


jaylagorio:
At least part of this seems similar to https://github.com/openaps/oref0/pull/855, which I'm also a fan of. Or am I missing something? Is there less overlap here than I think?

tynbendad:
thanks for pointing that out - if i had any clue how to view a pull requests' code i would compare the changes there to mine... any tips?

jaylagorio:
Sure! The easiest way is to click the Files Changed tab on the PR itself. It'll show you each of the files changed (in this case just one) and all the additions/deletions made.

https://github.com/openaps/oref0/pull/855/files

tynbendad:
thanks, sometimes i can't see which text is clickable ;)

yes, it looks similar as far as keeping bt going under a wifi connection, but i've also made it not break the bt connection when there is no ip connection, which is helpful when completely offline.

tynbendad:
Although it says something about how badly a change like these is needed that I came up with about the same one separately. I'd rather see 855 get in and make changes from there if it would help get one if these in sooner.

jaylagorio:
I'm a fan of the option for a full-time BT connection regardless of which PR gets merged, but agree with you that if it takes a crawl-walk-run approach by doing #855 and then follow up changes from you it would get us where we want to go!

scottleibrand:
The reason something like #855 hasn’t been merged is it seems that folks with Android phones suggest lots of things that would work great with Android, but would make for a worse user experience with iOS. I think we need someone with both kinds of phones to test all the potential changes to see which are improvements across both platforms and which help on one but hurt on the other. 

We might end up needing different behaviors configurable for which kind of phone you have.

tynbendad:
Both of the pull requests use preferences to enable the new behavior. And,
at least this pull request, should not change the flow of the script
without the new pref being enabled. So not supported in iPhone should not
block this. I have an iPhone and am pretty sure these prefs won't help it.

scottleibrand:
Ok, I like preference gated non-default options for new stuff. I’ll review this in more detail after ADA.

scottleibrand:
Let's rebase this only 0.7.0-dev so we can use all the fancy new bash function stuff.  :-)

I just created a branch to make that easier: https://github.com/openaps/oref0/compare/0.7.0-dev...1043-to-0.7.0-dev?expand=1

tynbendad:
i'm not on 0.7.0-dev yet, but have been meaning to setup a pi hat for a while. let me know if you need something from me for this & i'll try to catch up in the meantime.

scottleibrand:
#1048 merged: please test 0.7.0-dev to confirm that it works properly

PR #1044, opened 2018-06-24
real fix  if_mdt_get_bg

juehv:
Changed if statement to check preferences.json instead of openaps.ini for used cgm device.


juehv:
sure you're right. That's more readable.

PR #1050, opened 2018-07-05
Reduce onbattery threshold, refresh data faster

jaylagorio:
- Lowers the threshold used in `onbattery` to determine whether the device is plugged in to 90%
 - Removes a call to `onbattery` that causes Nightscout to be updated closer to real-time (and removes two therefore redundant if statements)

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -68,6 +68,8 @@ main() {
                 fi
             fi
             touch /tmp/pump_loop_completed -r /tmp/pump_loop_enacted
+	    # before each of these (optional) refresh checks, make sure we don't have fresh glucose data
+	    # if we do, then skip the optional checks to finish up this loop and start the next one
             if ! glucose-fresh; then
                 if onbattery; then
                     refresh_profile 30
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -74,8 +74,10 @@ main() {
                 else
                     refresh_profile 15
                 fi
-                pumphistory_daily_refresh
-                refresh_after_bolus_or_enact
+		if ! glucose-fresh; then
+		    pumphistory_daily_refresh
+                    refresh_after_bolus_or_enact
+		fi
             fi
             cat /tmp/oref0-updates.txt 2>&3
             touch /tmp/pump_loop_success
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -74,12 +74,12 @@ main() {
                 else
                     refresh_profile 15
                 fi
-		if ! glucose-fresh; then
-		    pumphistory_daily_refresh
-		    if ! glucose-fresh; then
+                if ! glucose-fresh; then
+                    pumphistory_daily_refresh
+                    if ! glucose-fresh; then
                         refresh_after_bolus_or_enact
                     fi
-		fi
+                fi
             fi
             cat /tmp/oref0-updates.txt 2>&3
             touch /tmp/pump_loop_success
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -76,7 +76,9 @@ main() {
                 fi
 		if ! glucose-fresh; then
 		    pumphistory_daily_refresh
-                    refresh_after_bolus_or_enact
+		    if ! glucose-fresh; then
+                        refresh_after_bolus_or_enact
+                    fi
 		fi
             fi
             cat /tmp/oref0-updates.txt 2>&3
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -68,8 +68,8 @@ main() {
                 fi
             fi
             touch /tmp/pump_loop_completed -r /tmp/pump_loop_enacted
-	    # before each of these (optional) refresh checks, make sure we don't have fresh glucose data
-	    # if we do, then skip the optional checks to finish up this loop and start the next one
+            # before each of these (optional) refresh checks, make sure we don't have fresh glucose data
+            # if we do, then skip the optional checks to finish up this loop and start the next one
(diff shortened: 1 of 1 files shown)


jaylagorio:
Unless there's a consequence of `pumphistory_daily_refresh` I missed, I was trying to say the `! glucose-fresh` conditional is redundant if you take the `onbattery` call out because then it becomes:

```
if ! glucose-fresh; then
    pumphistory_daily_refresh 
    if ! glucose-fresh; then 
        refresh_after_bolus_or_enact 
    fi 
fi 
```

Anyway, let me do some tests over the next few days to see how this impacts battery life. What would you want the preference named?

scottleibrand:
You still need one instance of ! glucose-fresh to avoid delaying the next loop if there is not going to be any wait_for_bg. 

Are you using / testing with an Edison and/or Pi rig?

jaylagorio:
Got it. This is a Pi Zero with the HAT.

scottleibrand:
Cool, then you’re in a good position to test. My goal is to get <5m loop times and >16h battery life out of a <3000mAh single-cell Li-ion.

jaylagorio:
My lipo is 2500mAh so I'll estimate the time based on that. I think we heard that with the underclock it's around 14.5 hours? And as for the < 5 minute loop time - so far so good, you're already there!

jaylagorio:
I just went looking for the Twitter thread where the person mentioned they got 14.5 hours and couldn't find it, but I don't recall the battery size they were using.

I tested with the 2500mAh battery yesterday on the `misc-issues` branch and got 10 hours exactly. Today with this PR as-is (rooted in `misc-issues`) on the same hardware and the same battery I got 9.5 hours. Not the biggest hit in the world, but I'm going to see if I can get a confirmation test tomorrow.

scottleibrand:
If you're not underclocking (to 500MHz) that would be worth a try as well.  If we can get the battery life we want with it underclocked and still do the faster data refresh, I think that'd be ideal.

jaylagorio:
Agreed! I am underclocking to 500MHz but I should probably add that Logger/xdrip-js running the Bluetooth stack constantly will impact the battery in addition to anything 0.7.0 is doing.

Do you have a link to a bigger battery you'd like me to try?

alimhassam:
I think xdrip-js is a good realistic test. Some people might also use the g4-loop witch use Bluetooth

jaylagorio:
Follow up question, slightly off topic: when I change the battery in the pump I noticed I don't have to do an mmtune before the loop starts up again, so is there a reason we mmtune at device startup instead of saving previous mmtune data and using it across device reboots?

jaylagorio:
After about a week of testing with this PR I've noticed:

 - Using Bluetooth I can get 12 - 13 hours of use on a 2500mAh battery, so if your goal is 16 hours on a 3000mAh battery that looks realistic.
 - I use a meter that sends its result to the pump, which automatically shows up in Nightscout as a red dot. I've noticed those aren't showing up anymore.

scottleibrand:
We still need one instance of ! glucose-fresh - you want to re-add that @jaylagorio?

Once that's fixed, what are folks' thoughts on battery life and whether we want to re-enable post-loop pumphistory pulls while on battery?

jaylagorio:
@scottleibrand The one on line 71 isn't good enough on its own? What else needs that check around it that I'm missing?

scottleibrand:
Every time we do something that takes time, we want to do another check to see if we’ve gotten new fresh glucose before starting the next time-consuming thing.

jaylagorio:
I swear I'm not intentionally being obtuse, and ordinarily this is what I'd look to code comments to try and figure out., but I don't think I understand the point of adding more `if` statements with calls to `glucose-fresh` in relation to the whole loop when we're already inside an `if` statement because of the return value from `glucose-fresh`.

That said, I returned those lines to essentially what they were before - if I missed the mark please let me know explicitly what you're looking for? Removing the `onbattery` call was my main goal and I'm totally OK making the `glucose-fresh` calls look like whatever you want them to look like.

scottleibrand:
If glucose wasn’t fresh, and you did a thing that takes time, glucose may now be fresh where it wasn’t when the first if check was performed.

jaylagorio:
Right - but then in relation to the overall loop, what happens if you have a fresh one right when you're about to start an operation that takes some time? I think you have a more big-picture view that I haven't had a chance to get to yet.

scottleibrand:
All of these operations are optional, and there's not much point in doing them unless it's going to be awhile until the next loop.  For example, if you just got a fresh glucose value while performing the `refresh_profile`, it makes no sense to do `pumphistory_daily_refresh` when you can instead end the loop, start the next one, and immediately pull a fresh pumphistory (and use it to loop).

jaylagorio:
> it makes no sense to do `pumphistory_daily_refresh` or `refresh_after_bolus_or_enact` when you can instead end the loop, start the next one, and immediately pull a fresh pumphistory (and use it to loop).

And there it is - doing these checks shortcircuits the current loop and instead makes execution fall through to the end so the next loop can start sooner.

That's what I was looking for, thank you!

scottleibrand:
Exactly.  And that's why I'm skeptical of removing the `onbattery` check: the pumphistory refresh is going to happen a couple minutes later regardless: we're just doing extra pump comms to speed it up by a minute or two, which may not be worth it while on battery.

So given that, and the measured battery life on the Pi, what are folks' thoughts on whether we want to re-enable post-loop pumphistory pulls while on battery?

scottleibrand:
Thoughts?

jaylagorio:
I'm consistently getting 13 hours on a Pi 0 + HAT with a 2500 mAh lipo using Bluetooth for data. Reliable for every day during the work week so far.

scottleibrand:
Going to go ahead and merge this, and use it as the basis for doing more frequent full pumphistory refreshes to minimize the impact of any race conditions while incrementally refreshing during bolus wizard runs.

PR #1052, opened 2018-07-09
Pump specific SMB size

MarkMpn:
Current SMB logic assumes a 0.1U multiple for boluses. Different pumps support different bolus increments, so make this configurable in the profile.

I use AndroidAPS rather than OpenAPS so I might well have got the profile setup structure wrong and I don't have an OpenAPS rig to test this on, so please let me know if anything needs changing!


scottleibrand:
LGTM.  Anyone want to test this before I merge it to dev?

PR #1054, opened 2018-07-12
Fix nodejs-legacy breakage on latest raspbian

cluckj:
The latest node-js for raspbian deprecates the use of nodejs-legacy, which breaks this script on new pi installs. This will make it so that nodejs-legacy is only installed on the edison.


scottleibrand:
One side effect of merging this without doing a full release will be that people with existing installs of 0.6.2 using the `master` branch will see a message that they're a commit behind.  Applying this change to existing installs doesn't do anything, except make that message go away, so there's no need to do a re-install or re-run of oref0-setup: anyone who wants to clear the message can just do `cd ~/src/oref0 && git pull`.

PR #1055, opened 2018-07-12
Reintroduce the A52_risk_enable functionality

jaylagorio:
The `smb_bolus` function wasn't observing the `A52_risk_enable` preference. For some reason, my pump will occasionally pause for more than one or two seconds when switching to the 24-hour log screen. I think this happens when there are a great many bolus events over that 24 hours but can't be sure. When it takes too long to get to this screen the ESC ESC ESC sequence also takes too long and the function fails, leading to an inability to SMB. This modification to reintroduce the `A52_risk_enable` functionality is my current workaround to the problem.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -386,12 +386,19 @@ function smb_bolus {
     #mdt bolus does not work on the 723 yet. Only tested on 722 pump
     file_is_recent enact/smb-suggested.json \
     && if (grep -q '"units":' enact/smb-suggested.json 2>&3); then
-        # press ESC four times on the pump to exit Bolus Wizard before SMBing, to help prevent A52 errors
-        echo -n "Sending ESC ESC ESC ESC to exit any open menus before SMBing "
-        try_return mdt -f internal button esc esc esc esc 2>&3 \
-        && echo -n "and bolusing " && jq .units enact/smb-suggested.json | nonl && echo " units" \
-        && ( try_return mdt bolus enact/smb-suggested.json 2>&3 && jq '.  + {"received": true}' enact/smb-suggested.json > enact/bolused.json ) \
-        && rm -rf enact/smb-suggested.json
+        if check_pref_bool .A52_risk_enable true; then
+            # don't press ESC four times on the pump - user has accepted risk of A52 errors
+            echo -n "SMBing and bolusing " && jq .units enact/smb-suggested.json | nonl && echo " units"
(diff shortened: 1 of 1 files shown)


jaylagorio:
When specifically enabled in preferences? As it was in previous versions?

scottleibrand:
The A52 risk preference is not a “please A52 my pump all the time” preference: that would be dangerous. Rather, it’s an acceptance that if we do everything possible to avoid an A52, it may still happen. 

It sounds like you instead want to retry the 4 ESC function?

scottleibrand:
We’ve always tried to do ESCs since we discovered the A52 issue, and should continue to try to do so.

jaylagorio:
After a day or two of the preference being put in the first time (where I believe it didn't do the escapes at all, like in this PR) I got the hang of making sure the loop had just finished or an SMB had just been delivered - I enabled the preference because I knew to be conscientious about it. It's not a preference in the JSON file because we wanted users to have to go hunting for it, but it was documented appropriately. I don't know that retrying the four escapes will work because that wraps the display back around to the 24-hour view.

Instead, could you give the failure timeout an extra two seconds before deciding to continue? That should give it enough time to do the extra processing it feels it needs to do.

scottleibrand:
Heh, we keep talking over each other. Which exact line do you think needs the sleep added?

jaylagorio:
The problem happens unpredictably so I'll need to wait until it happens again to test a `retry_return` change.

I don't think there's a `sleep` that can be added in the script because the failure comes from the `mdt` call. If that had an optional timeout parameter that could be passed which extended the otherwise reasonable default time before failure is declared that would solve the problem.

scottleibrand:
I’m not sure if @ecc1’s `mdt` script supports variable timeouts. 

If you can’t get anything else to work, a retry_return where we continue regardless of the output of the esc attempt should be safer than skipping them entirely, and would allow SMBs to continue even if they fail.

jaylagorio:
I'll give the `retry_return` a shot and let you know next time it happens.

scottleibrand:
@jaylagorio how did retry_return work?  Do you want to keep this PR open and modify it when ready, or close it and do a new one?

jaylagorio:
Hey there! The underlying pump problem has not come up again yet. I haven't been able to figure out how to cause it, only observe it for the 24 hours over which it takes place until it resoles itself the next day. Can we leave it open for a bit?

scottleibrand:
Shall we go ahead and close this, and do a new PR if/when needed?

jaylagorio:
Yup, let's get this off your plate. We can revisit if it starts up again.

PR #1059, opened 2018-07-16
Update duty cycling code

juehv:
* Add SPI reset
* Add logging function
* Enable independend timer for emergency actions

(Tested for around two weeks on my Pi0+ HAT + MDT)


juehv:
I think my update doesn't produce this build error since the report shows in line 843 "ok 60 Syntax checks bin/oref0-pump-loop.sh", which is the only file I changed.

PR #1068, opened 2018-07-26
 update duty cycling code

juehv:
I removed messages when duty cycling is disabled.


bin/oref0-pump-loop.sh
```diff
@@ -217,18 +218,14 @@ function check_duty_cycle {
         
         if [ "$DUTY_CYCLE" -gt "0" ] && [ "$DIFF_SECONDS" -gt "$DUTY_CYCLE" ]; then 
             echo "$DIFF_SECONDS (of $DUTY_CYCLE) since last run --> start new cycle."
-            return 0
-        elif [ "$DUTY_CYCLE" -eq "0" ]; then
```
scottleibrand: Do we want to remove this elif completely? Won’t that make it fall through to the next else, and print an incorrect message instead of no message?

juehv: hm you're right!

juehv:
@scottleibrand Do you have additional concerns with the code?

PR #1070, opened 2018-07-27
Option to select a pre-release go lib version

alimhassam:
Added an option to specify which version of go lib to use. This is useful for different reasons:
1. to manually enter a pre-release version, 
1. To select an older or alternate version if needed.
1. When we move to master, we could easily build upon this functionality to freeze the default version being downloaded. This will be done by changing the default value of the $ecc1medtronicversion variable from `latest` to the version we want to use on master.

Note: i just wrote the code rapidely for this, but didn't have time to test it yet as my dev rig isn't working at the moment. 

If someone can test that it all works it would be appreciated, otherwise i'll try to verify it all works once my spare rig is fixed.

cc: @ecc1

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -29,6 +29,8 @@ directory=""
 EXTRAS=""
 radio_locale="US"
 buildgofromsource=false
+ecc1medtronicversion="latest"
+ecc1dexcomversion="latest"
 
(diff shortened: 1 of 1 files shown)


ecc1:
The dexcom release tag won't in general be the same as the medtronic release tag, so there should be two variables.

scottleibrand:
@alimhassam you want to adjust this to address @ecc1's comment?

PR #1071, opened 2018-07-27
Fix not getting an IP on BT connection

Tornado-Tim:
For some reason the ```&&``` appendage breaks the dhclient command for the `bnep0` interface while connected to wifi with the new wifi and bt-options

Splitting the commands out fixes it.....


scottleibrand:
LGTM. If anyone wants to test this first LMK, otherwise I’ll merge it soon.

PR #1080, opened 2018-08-10
Use endsWith from Lodash

edwingustafson:
Lodash has an endsWith method — use it instead of maintaining project's own

### DIFF
--- a/lib/round-basal.js
+++ b/lib/round-basal.js
@@ -1,6 +1,4 @@
-var endsWith = function endsWith(text, val) {
-	return text.indexOf(val, text.length - val.length) !== -1;
-}
+var endsWith = require('lodash/endsWith');
 
 var round_basal = function round_basal(basal, profile) {
 


scottleibrand:
Is Lodash already a dependency here, or does this introduce any new dependencies?

edwingustafson:
Six other JavaScript files in the project require Lodash. My changeset adds it to `round-basal.js` for the first time. So it expands dependence on Lodash.

Be my guest to reject pull request if that's unwelcome.

```
$ find bin lib -name \*.js | xargs grep -n lodash
bin/oref0-upload-profile.js:24:var _ = require('lodash');
lib/iob/calculate.js:1:var _ = require('lodash');
lib/iob/history.js:4:var _ = require('lodash');
lib/profile/index.js:6:var _ = require('lodash');
lib/profile/isf.js:2:var _ = require('lodash');
lib/profile/basal.js:2:var _ = require('lodash');
lib/round-basal.js:1:var endsWith = require('lodash/endsWith');
```

PR #1081, opened 2018-08-12
Update Enlite CGM to use Go code

juehv:
* Clean up pump loop functions to call an external script
* Wrote an g4update-like script to update cgm values --> oref0-mdt-update.sh
* Script uses go program cgmupdate to update cgm values
* Script converts go output to old openaps format for compatiblity reasons (using jq)
* Script needs *latest* version of cgmupdate (thanks to @ecc1 for updating his go code)

! Many Thanks to @cluckj for his help during development
! Patch was tested for two weeks


scottleibrand:
Nice.  I wonder if it'd be better to source the new functions the way we do oref0-bash-common-functions.sh (or just put them all right into oref0-pump-loop) so we can reuse the existing wait_for_silence functions rather than reinventing that particular wheel (in a slightly different way for reasons I don't fully understand).

juehv:
I wanted to go towards CSMA/CA. That's a sophisticated technology used by WiFi to avoid communication collisions. But that was to much work for that weekend and I'm not sure if it's possible with the used radios. I keep working on that and maybe we can use it everywhere afterwards. The commonly used wait_for_silence was to slow for my taste ;)

Think a separate script is a good way to go, since it makes the loop script less messy and gives a feeling of modulation.

PS: Sorry for the typos .. I should be more careful with that ...

scottleibrand:
Wait for silence was originally based on the general approach of CSMA/CD, but has been modified to take into account the signal strength, as our goal is not (as with wifi) to let every rig get airtime, but rather to make sure that at least one rig is always able to loop, preferably the one that can do so with the fewest retries.

juehv:
Hm ... in the current dev version I can't find where signal strength is taken into account... Additionally I guess that's just a good idea for people using multiple rigs at the same time but a bad idea for groups of looping people (at a conference for example).
However, I can work with the current dev version of wait_for_silence if you want to move it to oref0-bash-common-functions.sh

For future it may be the better idea to include the listen before talk into go programms :)

juehv:
Should we leave it as it is for this iteration and I rather spend my time in implementing the listen before talk within the go code?

scottleibrand:
The signal strength logic is only on the wait for silence call after mmtune. The other calls all have varying wait times based on how long everything else takes.

scottleibrand:
I just pushed https://github.com/openaps/oref0/commit/ecce7655b25bcff4e1bdc690b5c99bb3c9a7213c if you'd like to replace your `listen_before_talk` calls with `wait-for-silence $CS_TIME_WAIT`

juehv:
I'm not happy with wait_for_silence. I get no feedback if radio is free or not. I fixed that in my update. I also removed the dead carelink code (carelink will not work with go code, so no need for that any more; and it was a depricaded version of getting preferences). What I would like to see (but didn't add without talking to you) is a short sleep (~1 second) between the listen commands to give other radios a chance to finish.

scottleibrand:
I would love to fix wait_for_silence to do things better. Totally happy with replacing it with your code, as long as we can pass the time interval like we do now.

juehv:
Ok cool! I'll add a sleep then 😄

juehv:
Ok it's done ... I don't see where is the problem with the automerge ... can you resolve this @scottleibrand  ??

scottleibrand:
Thanks: I’ll take a look in the morning.

juehv:
@Bender1061 Actually I have no idea what's happening then. Think we have to investigate in the Go code. Maybe @ecc1 knows that.

scottleibrand:
Resolved the conflicts.  Can you double-check that this is ready to merge from your perspective?

juehv:
Looks good from my side. I use this branch on my rig (Pi0 + HAT + Enlite + 754) and it works fine :)

PR #1084, opened 2018-08-18
Nightscout Upload Profile Token Auth Support

jpcunningh:


scottleibrand:
Looks good to me. Can you paste any test results confirming it leaves non-tokenauth behavior unchanged?

PR #1086, opened 2018-08-19
Remove Unused oref0 dexusb cgm loop

jpcunningh:
G4-Go script is much better than the old bin/oref0-dexusb-cgm-loop.py script.


jpcunningh:
@PieterGit, does the PR accomplish your suggestion in # ?

PieterGit:
yes, looks good. i'll test it as well.

PieterGit:
Some history/background. This script from me was used in the < 0.6.0 days (which used quite some more Python than nowadays). This used to work similar to the `G4-upload` script. The `G4-Go` CGM is much better, so this should not be used with 0.7.0 anymore and this script can be removed.

PR #1085, opened 2018-08-19
Independent mmtune script

jpcunningh:
openaps mmtune is no longer operative with 0.7.0.

This PR pulls the mmtune function out of oref0-pump-loop into a oref0-radio-tune.sh script so it can be called from the command line similar to how openaps mmtune used to work.

This PR also removes the default openaps mmtune alias that no longer functions.

PR #1087, opened 2018-08-19
various fixes (mainly WW related) from fresh edison/rpi0 install

PieterGit:
- also incorpates @jpcunningh https://github.com/openaps/oref0/pull/1085 and https://github.com/openaps/oref0/pull/1086 (sorry for not seperating my PR and @jpcunningh 's)
- fixes a problem with WW pumps that the `radio_locale` is not in the `pump.ini`. Also reported by @cluckj on https://gitter.im/openaps/oref0?at=5b78b34a802bc42c5f3b4870
- fixes a nasty issue with timeouts with pip3/pip packages, see https://github.com/openaps/oref0/issues/1095
- upgrade to node 8.11.4 and thus npm 5.6.0

Work in progress / do not merge yet.

The uploading of cgm data with `G4-go` does not work for me with tokenbased authentication. Working on that.

@scottleibrand can you do a review so far?


jpcunningh:
@PieterGit, will you incorporate the changes youn requested in this PR such that I should close the other 2 PRs?

Thanks!

bin/oref0-cron-every-minute.sh
```diff
@@ -60,7 +60,8 @@ find /var/log/openaps/pump-loop.log -mmin +5 | grep pump && (
 ) | tee -a /var/log/openaps/pump-loop.log &
 
 # if the rig doesn't recover after that, reboot:
-oref0-radio-reboot &
+# DEPRECATED WITH 0.7.0. See if we can do without
+# oref0-radio-reboot &
```
scottleibrand: No, this wasn't deprecated: I re-enabled it for 0.7.0 with commit de462defb4d0c9a8fa6b173cf29febe2e6a165ca

package.json
```diff
@@ -98,7 +99,7 @@
     "lodash": "^4.15.0",
     "moment": "^2.14.1",
     "request": "^2.79.0",
-    "share2nightscout-bridge": "^0.1.5",
+    "share2nightscout-bridge": "git://github.com/nightscout/share2nightscout-bridge.git#wip/generalize",
```
scottleibrand: What does this do?

npm-shrinkwrap.json
```diff
@@ -0,0 +1,2023 @@
...
+      "resolved": "https://registry.npmjs.org/strip-ansi/-/strip-ansi-3.0.1.tgz",
+      "integrity": "sha1-ajhfuIU9lS1f8F0Oiq+UJ43GPc8=",
+      "requires": {
+        "ansi-regex": "^2.0.0"
+      }
+    },
+    "strip-bom": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/strip-bom/-/strip-bom-2.0.0.tgz",
+      "integrity": "sha1-YhmoVhZSBJHzV4i9vxRHqZx+aw4=",
```
scottleibrand: Was this added intentionally?

bin/oref0-cron-every-minute.sh
PieterGit: i see a lot of radio reboots with the 0.7.0 that seem unnecessary and cause interruption during normal  pump-rig comms.  if it's really needed in some corner cases perhaps move to every 15 minutes? I would like to make it configurable and see if I can do without these reboots.  I think there is still some other issue that makes rebooting without network not boot up the rig without problems.

package.json
PieterGit: This is the format Nightscout uses. We should re publish a new NPM package. I tried contacting Ben West, but didn't get a response.
For a list of changes see https://github.com/nightscout/share2nightscout-bridge/commits/wip/generalize
It fixes a security issue.

bin/oref0-cron-every-minute.sh
PieterGit: only lengthen the delay will keep a lot of false positives (that hopefully will be cancelled in time). previouse mmweowlink and pump comms could really be messed up after a reboot. I rather see if we can minimize the amount of reboots, perhaps only make them as an option.

How about testing it every 15 minutes and upping the reboot time to 10 minutes. That would give a minimum of 10 minutes and max of 25 minutes in case of pump-rig comms failures. Thoughts?

npm-shrinkwrap.json
PieterGit: it's changed with the npm commands. i did not have a lot of merge conflicts with it for nightscout. In case you have merge problems, you always re-create it with `npm shrinkwrap`

bin/oref0-cron-every-minute.sh
scottleibrand: I've never seen any issues caused by a reboot: can you provide more specifics on what problems you've seen as a result of rebooting a 0.7.0 rig?

I'd rather not wait 25 minutes to start looping again if the rig gets into a bad state.  I'm ok with a few extra log messages about scheduled reboots that get canceled before occurring: those don't do any harm, and if someone is looking at their logs closely enough to notice them, they might point to an underlying issue...

bin/openaps-packages.sh
```diff
@@ -20,11 +20,11 @@ if getent passwd edison > /dev/null; then
 fi
 #if ! sudo apt-get install -y npm; then
 # install/upgrade to node 8
-if ! nodejs --version | grep 'v8.'; then
+if ! nodejs --version | grep 'v8.11'; then
```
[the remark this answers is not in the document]
PieterGit: yes, changed to 
```
# install/upgrade to latest node 8 if neither node 8 nor node 10+ LTS are installed
if ! nodejs --version | grep -e 'v8\.' -e 'v1[02468]\.' ; then
```

bin/oref0-cron-every-minute.sh
PieterGit: @scottleibrand 
Issues I now of are:
- older explorer boards have issues with Dexcom G4 USB and the micro usb needs to be inserted after reboot
- edison does not have RTC and during systemstartup internet connection is required in order to set the right time. We should make sure the time from the pump is used in those cases, but that doesn't completely work (yet)

Another reason is that rebooting in case of an issue instead of fixing the root cause is a Microsoft way of doing stuff.

Can we agree on:
- making reboot configurable
- let's try to disable it by default and see if the current 0.7.x codebase has issues that resolve if you reboot the rig
- try to resolve those issues, instead of rebooting
- people who enable the setting should check if the 0.7.0-dev rig reboots in common daily use
- before 0.7.x release decide if we need to use enable reboot by default

npm-shrinkwrap.json
PieterGit: removed ` npm-shrinkwrap.json`. The proper way is not using `npm run global-install` and using a published npm package, but that's quite some work to migrate to that.

scottleibrand:
Fixed merge conflicts and made a few other small changes in https://github.com/openaps/oref0/compare/dev...PieterGit-201808_upgrade?expand=1 to get it to the point where I think it's ready to merge.  @PieterGit is this/that ready to merge from your perspective?

scottleibrand:
Testing that, I get:

npm ERR! invalid: share2nightscout-bridge@0.1.5 /usr/lib/node_modules/oref0/node_modules/share2nightscout-bridge
npm ERR! invalid: yargs@4.3.2 /usr/lib/node_modules/oref0/node_modules/yargs

Investigating...

scottleibrand:
The first time I ran oref0-runagain, it errored out completely, complaining about missing packages and files.  Re-running it, it still got a bunch of complaints about `share2nightscout-bridge` and `yargs`, but it continued past them successfully.  On the third run, it didn't complain at all.  Not sure if there's some order-of-operations bug here or what: we'll have to keep an eye on what happens on additional rigs.

scottleibrand:
Looks like the latest yargs is incompatible with node 0.10:

/root/src/oref0/node_modules/yargs/index.js:5
const yargs = require('./yargs')
^^^^^
SyntaxError: Use of const in strict mode.
    at Module._compile (module.js:439:25)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at Object.<anonymous> (/root/src/oref0/bin/oref0-get-profile.js:42:16)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
Could not run oref0-get-profile

So I just updated #1136 to force upgrade to node 8+

scottleibrand:
Closing in favor of #1136

PR #1088, opened 2018-08-20
Pump read error checking

jpcunningh:
Today, our rig ended up with a zero size file for settings/basal_profile.json. The jq command doesn't return an error code if it doesn't find the value argument, so the get_settings didn't fail.

This PR checks to ensure there are valid values using grep on the output of jq.

It needs to be tested for mmol and carb ratio units other than grams.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -959,7 +959,7 @@ function check_battery() {
 }
 function check_tempbasal() {
   set -o pipefail
-  mdt tempbasal 2>&3 | tee monitor/temp_basal.json >&4 && [ -s monitor/temp_basal.json ] && cat monitor/temp_basal.json | jq .temp >&4 && cp monitor/temp_basal.json monitor/last_temp_basal.json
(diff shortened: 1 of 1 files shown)


scottleibrand:
LGTM.  We'll see if any mmol users want to test it (https://gitter.im/openaps/oref0?at=5b7afa2594f8164c1775fbbc) and then merge to dev for further testing there.

PieterGit:
Did some testing. Initial results look good, but I'll retest later. See Gitterlink above for the discussion of the testing.

PieterGit:
If I interpret https://github.com/ecc1/medtronic/search?p=2&q=grams&unscoped_q=grams correctly the medtronic Go libraries only seem to work with `grams`: 
image

I wouldn't even now what medtronic.Exchanges are at: 
https://github.com/ecc1/medtronic/blob/b39ccbd1825373fc96e9180fe02b432a35edbcbe/cmd/mdt/json.go#L68-L78

scottleibrand:
Ok, sounds like this is ready to merge to dev for further testing there.  If no objections I'll do so soon.

PR #1090, opened 2018-08-21
Improved incremental pump history update.

alimhassam:
Querying by top record timestamp is sometimes racy because of different
reasons:
 - some records have the same timestamp such as Bolus and BolusWizard.
 - some records have timestamps out of order such as DailyTotal records.
 - Timestamps can also become non monotonically increasing whenn clock
has been changed.

To avoid all these problems, a new argument was added to the pumphistory
command to query from a given record id. The record id is the base64
encoded string of the record.

Note this is not available yet in the latest go pump libaries, so this
change depends on https://github.com/ecc1/medtronic/pull/8


alimhassam:
Note, i'm currently running this branch with additional testing code seen here https://github.com/alimhassam/oref0/compare/improvedIncrementalHistory...alimhassam:improvedIncrementalHistoryTesting?expand=1
which is trying to compare the incremental pump history with a full refresh on every loop. I'll try to test and monitor different scerarios, and report back, but let me know any comments meanwhile.

alimhassam:
Sorry I wasn't clear. This can't be merged until the corresponding change on ecc1 repo gets merged

alimhassam:
If anyone wants to test this pull request which removes the race condition in incremental pump history, it's now available when building go lib from source.

alimhassam:
This is now ready to merge

scottleibrand:
What happens after we merge this if someone just does `git pull` and doesn't re-run oref0-setup to get the latest `medtronic`?

alimhassam:
It won't work. They'll need to get the latest binary.

alimhassam:
It'll also be difficult to work around this. Any suggestions? 
Might just be best to merge it and let people know they should re-run oref0-setup? Or oref0-runagain, since this is the dev branch. (Not sure about runagain still working I want to try it out when I get time)

 Knowing this, I'll let you decide when you think its the best time to merge this pr.

scottleibrand:
How exactly will it fail? Will it stop looping until setup/runagain is run? Or something more subtle?

alimhassam:
The pumphistory command will fail with an error saying the parameter is not supported, so I'm pretty sure looping will stop.
I'll test it out this evening to confirm and paste the logs.

alimhassam:
I rolled back my pumphistory binary & openaps.jq binary to a previous version. What actually happens is that it does a full pump history refresh at every loop, but still works. The reason for this is that the old openaps.jq doesn't produce an id field and so the code automatically falls back to a full history refresh.

I think this is acceptable behaviour, though it would be better if people run oref0-runagain after updating their dev branch to get the the benefits.

scottleibrand:
Ok, great. I like backwards compatibility.

scottleibrand:
Looks like there is a merge conflict with valid_pump_settings() - can you resolve that? Is that just a “keep both” thing?

alimhassam:
That's right. Done, I kept both change.

PR #1091, opened 2018-08-22
Fix Logger integration for oref0 interactive install script.

Jonathan-Cole:
This PR should fix a defect with setting the transmitter id and update the script to use the new Logger config file (xdripjs.json).  
Note that the config file enhancement is currently only on the dev branch but should be merged back to master soon.

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -305,7 +305,6 @@ if [[ -z "$DIR" || -z "$serial" ]]; then
     fi
     if [[ ${CGM,,} =~ "xdrip-js" ]]; then
         prompt_and_validate DEXCOM_CGM_TX_ID "What is your current Dexcom Transmitter ID?" validate_g5transmitter_serial
-        DEXCOM_CGM_TX_ID=$REPLY
         echo "$DEXCOM_CGM_TX_ID? Got it."
         echo
     fi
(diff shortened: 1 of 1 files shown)


Jonathan-Cole:
For reference, here is the PR for the Logger enhancement to use a config file: https://github.com/xdrip-js/Logger/pull/113

bin/oref0-setup.sh
```diff
@@ -1068,11 +1067,15 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
...
+        echo xdrip-js selected as CGM, so configuring xdrip-js
         git clone https://github.com/xdrip-js/Logger.git $HOME/src/Logger
         cd $HOME/src/Logger
         sudo npm run global-install
-		touch /tmp/reboot-required
+	touch /tmp/reboot-required
+	#Set transmitter id & generate xdripjs.json config
+	if validate_g5transmitter_serial $DEXCOM_CGM_TX_ID; then
+	  cgm-transmitter "$DEXCOM_CGM_TX_ID"
+	fi
```
scottleibrand: tabs vs. spaces

scottleibrand:
I won't be able to test any of this, so will need other Logger users to do so.

tynbendad:
Scott, i am so git-deficient... if the xdrip-js side of this has already been merged into xdrip-js dev branch, then noone can use the existing openaps dev xdrip-js target, so can you merge this pr based solely on it not harming anything else?

PR #1094, opened 2018-08-27
Remove startsWith and use indexOf instead.

jpcunningh:
This removes an incompatibility with older versions of node.


PieterGit:
I would advice to drop compatibitliy with older node versions than 8.11.x in oref0 0.7.x.
See https://gitter.im/openaps/oref0?at=5b84e316f012380179993268

PR #1098, opened 2018-08-29
Have autotune sync to NS nightly after autotune runs

danamlewis:
This will keep Nightscout in sync with autotune after the nightly run.


danamlewis:
Inspired by @jpcunningh 's additions to the doc with the manual instructions (https://openaps.readthedocs.io/en/latest/docs/While%20You%20Wait%20For%20Gear/nightscout-setup.html#how-to-automatically-sync-your-profile-from-autotune). 

I can't think of any reason why uploading nightly would not be desired behavior, but please chime in if you can think of a use case where that's non-optimal.

bin/oref0-cron-nightly.sh
```diff
@@ -17,5 +17,5 @@ directory="$PWD"
 
 if [[ $ENABLE =~ autotune ]]; then
     # autotune nightly at 4:05am using data from NS
-    (oref0-autotune -d=$directory -n=$NIGHTSCOUT_HOST && cat $directory/autotune/profile.json | jq . | grep -q start && cp $directory/autotune/profile.json $directory/settings/autotune.json) 2>&1 | tee -a /var/log/openaps/autotune.log &
+    (oref0-autotune -d=$directory -n=$NIGHTSCOUT_HOST && cat $directory/autotune/profile.json | jq . | grep -q start && cp $directory/autotune/profile.json $directory/settings/autotune.json && oref0-upload-profile settings/profile.json $NIGHTSCOUT_HOST $API_SECRET) 2>&1 | tee -a /var/log/openaps/autotune.log &
```
jpcunningh: I think you would want to upload either the `$directory/autotune/profile.json` or the `$directory/settings/autotune.json` rather than the `$directory/settings/profile.json` because it won't be updated yet to the new basal profile in the `$directory/settings/autotune.json` by the time it runs. 

The downside is that you won't pick up changes made in the preferences.json file over time.  There is a separate issue opened up on autotune not picking up those changes.  Once that is resolved, I don't think `$directory/settings/autotune.json`  will have a downside.

tim2000s:
I think this should be possible to PR now as the Autotune update issue has been closed and rolled into dev?

efidoman:
I plan to test these changes on my system as soon as I get my install issues resolved.

scottleibrand:
Have been running this for awhile, and it's working properly.  Ready to merge if no objections.

tim2000s:
I’ve been seeing an oddity with this. The uploaded script hasn’t been uploading the autotuned CR, only the pump CR. It also seems to split the last half hour of carb ratio out. Anyone else seen this? (Note that upload currently disabled, hence date in image)
3f05273f-283e-451c-a1d7-73f6ef1331e7

scottleibrand:
Our uploaded CR record isn’t showing a pump value or multiple CR schedules, but it’s also not updating and is now out of date.

viq:
Woudln't it also make sense for the `oref0-upload-profile` to issue a `Profile Switch` even to nightscout?

tim2000s:
@viq No. If you automatically make a profile switch then you don’t give people the option of not profile switching, which some may wish to have. Profile switch events should either be an option in set-up of profile upload or a separate trigger event.

viq:
AFAIK autotune on OpenAPS "just happens", without you getting to review
results first? This would be the same behaviour. And I'm all for adding a
flag to do it, just I think it would make sense for it to be part of same
process. --

viq

tim2000s:
Given the way that people are using AndroidAPS is noticeably different from OpenAPS, I’d disagree. 

As people have now started to use the automated profile switching capabilities off triggers, automating a profile switch in NightScout when this is going on starts to run into conflict with what people are doing to override a profile. This isn’t something that exists in a OpenAPS world so has never been an issue. 

That’s why I’d prefer it to be something that is a deliberate choice for someone wanting to use Autotune. Enough people don’t understand the OpenAPS aspects of AndroidAPS so it would be good to force them to explicitly understand the implications of this.

viq:
Sure, let's not make it the default behaviour. But how about a flag "and
also switch to the profile that you just uploaded"?

-- 

viq

tim2000s:
I don’t think that’s unreasonable. Would be worth updating the docs at the same time with a warning that this would override any automated profile switching using glucose triggers.

scottleibrand:
Updated rounding, and confirmed that a manual run is uploading updated data. Also updated the nightly cron to wait 60s between finishing autotune and uploading profile, to allow cron-every-minute to run ns-loop and update the profile.json. Will let it run on one rig and check it after tonight's autotune run to confirm it gets the new values.

scottleibrand:
This successfully uploaded our newly autotuned profile this morning at 04:08:50, so I think it's good to merge.

st8ass:
> Это успешно загрузило наш недавно настроенный профиль сегодня утром в 04:08:50, так что я думаю, что это хорошо для слияния.

@scottleibrand does not work, if configured authorization with token 
authorization-with-token

scottleibrand:
If you or anyone would like to volunteer to add tokenauth support, I'd be happy to review a PR. We don't use it, though, so I can't really develop or test that easily.

I also wouldn't want to hold up the 0.7.0 release for missing tokenauth support, unless it causes failures of something other than the new profile upload feature. Have you seen any such side effects?

st8ass:
I understand. No problem with that, this is not essential

PR #1100, opened 2018-08-30
Split the pre-bolus ESC key presses in two commands.

alimhassam:
The 4 ESC seems to be done too quickly now that we moved to the go library.
It seems like it isn't as good at avoiding A52 errors as what
was done on master. I think on master the pump is wakened up again between
each key presses, or the python layer adds some latency which isn't
there on the new version.

The compromise here is to send the first ESC, wait for 0.5s before sending the
remaining ESC. From testing on a test pump, this seems to reduce the
chance of getting an A52 error.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -425,7 +425,7 @@ function smb_bolus {
         echo -n "Sending ESC ESC ESC ESC to exit any open menus before SMBing "
         try_return mdt -f internal button esc 2>&3 \
         && sleep 0.5s \  
-        && try_return mdt -f internal button esc esc esc 2>&3 \
+        && try_return mdt -f internal button esc esc esc esc 2>&3 \
         && echo -n "and bolusing " && jq .units enact/smb-suggested.json | nonl && echo " units" \
         && ( try_return mdt bolus enact/smb-suggested.json 2>&3 && jq '.  + {"received": true}' enact/smb-suggested.json > enact/bolused.json ) \
         && rm -rf enact/smb-suggested.json
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -423,8 +423,8 @@ function smb_bolus {
     && if (grep -q '"units":' enact/smb-suggested.json 2>&3); then
         # press ESC four times on the pump to exit Bolus Wizard before SMBing, to help prevent A52 errors
         echo -n "Sending ESC ESC ESC ESC to exit any open menus before SMBing "
-        try_return mdt -f internal button esc 2>&3 \
-        && sleep 0.5s \  
+        try_return mdt -f internal button esc esc 2>&3 \
+        && sleep 0.5s \
         && try_return mdt -f internal button esc esc esc esc 2>&3 \
         && echo -n "and bolusing " && jq .units enact/smb-suggested.json | nonl && echo " units" \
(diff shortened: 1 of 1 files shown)


alimhassam:
yes, the ESC sent while holding the down button are ignored. you're right, might be better to send 1+4 instead.

scottleibrand:
Can we tell from the return code or something if the pump actually accepted the button press?

alimhassam:
not that I'm aware of, unfortunately. The first ESC sent while the button down is pressed does seem to do something as it pauses the scrolling. But since the down or up button is pressed, it doesn't move us back to the previous page of the wizard.

scottleibrand:
Would sending four ESCs as individual commands break out of the scroll menu even if the button is still held down? Or do you have to react and release it before they’ll take?

alimhassam:
You have to react and release before the ESC can actually move the menu to the previous page. 

It seems to be the same as trying to press ESC manually on the pump (which makes sense). I also tried manually, and while keeping the down arrow pressed to scroll, no matter how fast or how many time I press ESC, it doesn't exit the wizard.

scottleibrand:
So we have to have a long enough wait to expect the user to be able to notice the scrolling has stopped and release the button, but not so long that they’ll start scrolling again? Human factors are fun. Can I put in another vote for “let’s stop using the bolus wizard”?

bin/oref0-pump-loop.sh
```diff
@@ -423,7 +423,9 @@ function smb_bolus {
     && if (grep -q '"units":' enact/smb-suggested.json 2>&3); then
         # press ESC four times on the pump to exit Bolus Wizard before SMBing, to help prevent A52 errors
         echo -n "Sending ESC ESC ESC ESC to exit any open menus before SMBing "
-        try_return mdt -f internal button esc esc esc esc 2>&3 \
+        try_return mdt -f internal button esc 2>&3 \
+        && sleep 0.5s \
```
scottleibrand: Is this syntax correct? Or do we drop the s?

bin/oref0-pump-loop.sh
```diff
@@ -423,7 +423,9 @@ function smb_bolus {
     && if (grep -q '"units":' enact/smb-suggested.json 2>&3); then
         # press ESC four times on the pump to exit Bolus Wizard before SMBing, to help prevent A52 errors
         echo -n "Sending ESC ESC ESC ESC to exit any open menus before SMBing "
-        try_return mdt -f internal button esc esc esc esc 2>&3 \
+        try_return mdt -f internal button esc 2>&3 \
```
scottleibrand: Let’s do two here. Having an even total number of ESCs puts the pump back on the home screen, vs the info screen you get with one ESC.

alimhassam:
:+1: I use it occasionally 5-10 times a week, but I only got A52 once earlier this year. Just wanted to follow up and try to make it a bit better. If the additional delay here is not worth it, I'm okay also with dropping this pr completely.

scottleibrand:
I think a half second delay is worth adding.

bin/oref0-pump-loop.sh
alimhassam: I think it's okay, but I introduced an unwanted space in my last change.

bin/oref0-pump-loop.sh
alimhassam: done.

Issue #1093, opened 2018-08-25
Use of autotune.json file overrides changed settings on pump and in preferences.json


tim2000s:
When Autotune creates its new profile from the previous run’s Autotune.json file, it ignores changes made in the pump which are propagated in profile.json or in the preferences.json files. 

This means that if you flip from multiple ISFs to one or if you switch insulin curve, Autotune doesn’t pick up the change without manual intervention. It means we should warn people who have changed to Fiasp to check what their Autotune settings are as they are probably tuning to the rapid-acting rather than ultra-rapid curve. 

To fix it we need to identify when the appropriate settings in profile.json don’t match autotune.json and adjust Autotune accordingly.

scottleibrand:
I think the first step toward troubleshooting this is creating a simple process to reproduce it.  Do you know (or can you figure out) what the exact (minimal) steps are required to do so?

jpcunningh:
@tim2000s, that was my first thought, but then I realized that OpenAPS already keeps the pump profile updated with everything except the autotune basals, so #1101 uses the data that's in the pump profile.

danamlewis:
Closing as this appears to be resolved. Please comment in #1251  if this is still an issue (I have a comment there to see if this is resolved fully).

PR #1101, opened 2018-08-30
Fix Stale Autotune Profile

jpcunningh:
This is an attempt to address #1093.

Having tested Fiasp a couple of weeks ago, we noticed the same behavior.

This uses the pump profile to keep curve, dia, and peak up to date.

### DIFF
--- a/bin/oref0-autotune-prep.js
+++ b/bin/oref0-autotune-prep.js
@@ -84,6 +84,15 @@ if (!module.parent) {
         }
     }
 
+    // get insulin curve from pump profile that is maintained
+    profile_data.curve = pumpprofile_data.curve;
+
+    // if not tuning the insulinEndTime and insulinPeakTime, then use data from pump profile
+    if (!tune_insulin_curve) {
+      profile_data.dia = pumpprofile_data.dia;
+      profile_data.insulinPeakTime = pumpprofile_data.insulinPeakTime;
+    }
+
     try {
         var glucose_data = JSON.parse(fs.readFileSync(glucose_input, 'utf8'));
     } catch (e) {
--- a/bin/oref0-autotune-prep.js
+++ b/bin/oref0-autotune-prep.js
@@ -87,8 +87,10 @@ if (!module.parent) {
     // get insulin curve from pump profile that is maintained
     profile_data.curve = pumpprofile_data.curve;
 
-    // if not tuning the insulinEndTime and insulinPeakTime, then use data from pump profile
-    if (!params['tune-insulin-curve']) {
+    // Pump profile has an up to date copy of useCustomPeakTime from preferences
+    // If the preferences file has useCustomPeakTime use the previous autotune dia and PeakTime.
+    // Otherwise, use data from pump profile.
+    if (!pumpprofile_data.useCustomPeakTime) {
       profile_data.dia = pumpprofile_data.dia;
       profile_data.insulinPeakTime = pumpprofile_data.insulinPeakTime;
     }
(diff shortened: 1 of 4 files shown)


jpcunningh:
@tim2000s, it won't affect overriding the Mandated minimum DIA since that mandate is enforced in the determine IOB calculation code regardless of the settings passed in.

I didn't change how ISF is handled because autotune tunes ISF similar to basals. It looked like the code already respects the boundaries on how far ISF can be moved from the pump profile which would get updated if the pump setting changed.

bin/oref0-autotune-prep.js
```diff
@@ -87,6 +87,15 @@ if (!module.parent) {
         }
     }
 
+    // get insulin curve from pump profile that is maintained
+    profile_data.curve = pumpprofile_data.curve;
+
+    // if not tuning the insulinEndTime and insulinPeakTime, then use data from pump profile
+    if (!tune_insulin_curve) {
```
scottleibrand: This would change those values back to defaults on subsequent runs if you tune insulin curve manually and have autotune running on cron. Do we want that? Or should we stick with the tuned insulin curve?

bin/oref0-autotune-prep.js
```diff
@@ -87,6 +87,15 @@ if (!module.parent) {
         }
     }
 
+    // get insulin curve from pump profile that is maintained
+    profile_data.curve = pumpprofile_data.curve;
```
scottleibrand: Does the pumpprofile get updated when preferences.json is changed?

jpcunningh: It gets updated by oref0-pump-loop get_settings function by calling oref0-get-profile without the autotune profile.

jpcunningh:
Just a note that this has an interaction with #1114. 

Whichever is merged first, the other will need to be updated.

jpcunningh:
Rebased onto #1106 to resolve conflict.

bin/oref0-autotune-prep.js
scottleibrand: I suspect the more common use case is that someone will only have DIA set in the pump, and then will run tune_insulin_curve in a manual autotune run.  I don't want to impose too many hoops: it seems like the default behavior should be "if you've tuned insulinPeakTime or insulinEndTime, keep using the tuned values"...

scottleibrand:
Looks like merging the other PRs introduced another merge conflict.

jpcunningh:
rebased to dev to resolve merge conflict.

bin/oref0-autotune-prep.js
jpcunningh: That makes sense for that to be the more common use case.

Any thoughts on how to make the fact that curve values have been tuned persist so it could test for that case?

jpcunningh:
Not yet. There is a question in the thread above about default behavior.

bin/oref0-autotune-prep.js
scottleibrand: Would it be as simple as checking whether useCustomPeakTime is set (whatever it's called exactly) instead of checking to see if we're doing tune_insulin_curve at the moment?

jpcunningh: Can I add a flag to the autotune output profile it would set to true if tune_insulin_curve had been run? Testing the flag instead of the argument would then be used to determine if it needs to keep the dia and peak from the previous autotune profile vs. pull it in from the pumpprofile.

jpcunningh: Yep, I was overthinking it. Changed to use useCustomPeakTime.

jpcunningh:
Ready to merge.

PR #1109, opened 2018-08-31
Dont reboot after restart

alimhassam:


bin/oref0-radio-reboot.sh
```diff
@@ -15,19 +15,24 @@ EOF
...
-        echo >> $logfile
-    fi
-else
-    if [ -e /run/nologin ]; then
-        echo "No more radio errors; canceling reboot" | tee -a $logfile
-        shutdown -c | tee -a $logfile
-    fi
+SYSTEM_UP=$(date --date="$(uptime -s)" +%s)
+# Start checking only when the system has been up for 15 minutes.
+START_CHECKING=$(date -d"30 minutes ago" +%s)
```
scottleibrand: 30 or 15?

alimhassam: let me know, what's a reasonable number here, I thought the general rule was that the loop might take up to 20 min to start working so I decided to set it to 30 min to give it some time.

scottleibrand: I don't really care, as long as the comment matches.

alimhassam: Ha, oops. I'll take out the number from the comment.

alimhassam: fixed comment.

Issue #1058, opened 2018-07-14
DUTY_CYCLE counts from END of cycle, not start


N3FM:
When setting the duty cycle, in oref0-pump-loop.sh, it seems to be sampling every 60 seconds. However, the elapsed time for the existing cycle is shown at the start of the delay loop, and it exits only when the sampling exceeds the set point (not when it matches). 
For example, set the duty cycle to 240 seconds. I would expect the next cycle to start 4 minutes after the previous cycle started. However, here is what we get: (previous cycle 'Starting oref0-pump-loop at Sat Jul 14 17:37:02 EDT 2018', checked pump clock at 17:38:22)
`
Completed oref0-pump-loop at Sat Jul 14 17:38:34 EDT 2018

28 (of 240) since last run --> stop now.
88 (of 240) since last run --> stop now.
148 (of 240) since last run --> stop now.
208 (of 240) since last run --> stop now.
268 (of 240) since last run --> start new cycle.

Starting oref0-pump-loop at Sat Jul 14 17:43:02 EDT 2018 with 9 second wait_for_silence:
`
This exposes two issues:
1) 28 seconds before 17:38:34 is NOT the time shown by previous 'Starting oref0-pump-loop'; 
2) The duty cycle delayed 240 seconds from the END of the cycle. It shows the 28 seconds (plus preflights) consumed by the previous cycle, and exits 240 seconds after THAT (268 seconds). If we are trying to match up with the 300 second cycle of the Medtronic pumps, this will drift later and later each cycle. (In the example, 'Starting oref0-pump-loop' cycles began 6 minutes (360 seconds) apart when duty cycle was set to 240 seconds.)

For consideration: If we are retrieving the 'actual' MDT sampling/averaging time, add 5 minutes plus a fixed 'processing / error budget' offset to that time, to set the next retrieval time. 
Hope this helps!
N3FM  (Bob Smallwood)

scottleibrand:
PRs welcome.  Also tagging @juehv as he's the only one I know of using the DUTY_CYCLE code.

N3FM:
@juehv Jens, thanks for the info.  The target I'm hoping to achieve is to only run a loop every 5 minutes -- and I think success in that arena would be a BIG DEAL, IF the prior loop was successful -- I'm hoping to NOT pull in any 'No new CGM ... ' messages, as my little rig is prone to get into mischief if it doesn't have new BG to play with (blaming that on 0.6.2 for now! ) 
[...]

PR #1112, opened 2018-09-01
Update duty cycle code

juehv:
Added suggestions from https://github.com/openaps/oref0/issues/1058 and removed emergency actions since a different script is rebooting when radio problems occur.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -131,108 +131,63 @@ function fail {
 
 # The function "check_duty_cycle" checks if the loop has to run and it returns 0 if so.
 # It exits the script with code 0 otherwise.
-# The decision is based on the time since last *successful* loop.
-# !Note duty cycle times are set in seconds.
 #
-# Additionally it may start an "emergency action" if enabled.
-# Possible actions are USB power cycling, SPI reset or reboot the system.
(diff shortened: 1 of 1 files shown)


juehv:
@scottleibrand can you please review the changes? It'll make the core much simpler and removes redundant functionality :)

bin/oref0-pump-loop.sh
```diff
@@ -709,7 +669,7 @@ function get_settings {
         retry_return read_insulin_sensitivities 2>&3 >&4 || return 1
         retry_return read_carb_ratios 2>&3 >&4 || return 1
         retry_return openaps report invoke settings/insulin_sensitivities.json settings/bg_targets.json 2>&3 >&4 || return 1
-	#NON_X12_ITEMS=""
+    #NON_X12_ITEMS=""
```
scottleibrand: Tabs vs. spaces? Maybe just remove this old line.

juehv: This line isn't from me :D I just changed the duty cycle function ;)

scottleibrand: No, but the diff shows you changed the whitespace such that it is now indented correctly. I’m suggesting either fixing that or removing the line.

juehv:
@scottleibrand other issues ?

PR #1113, opened 2018-09-01
Delete oref0-mdt-trend.js

juehv:
oref0-mdt-trend.js was used by python code to add trend information. This is done by the go code now, and thus, oref0-mdt-trend.js is not needed any more.


juehv:
No this is definetly an error (maybe because I pushed the update before mdt-update was in ...)

PR #1114, opened 2018-09-01
oref0-autotune-prep arg processing

jpcunningh:
Refactor argument processing to use yargs for readability and less error prone.


jpcunningh:
Just a note that this has an interaction with #1101.

Whichever is merged first, the other will need to be updated.

PR #1119, opened 2018-09-03
Oref0 determine basal arg parsing cleanup

jpcunningh:


bin/oref0-determine-basal.js
```diff
@@ -20,81 +20,70 @@ var determine_basal = require('oref0/lib/determine-basal/determine-basal');
...
     var argv = require('yargs')
-      .usage("$0 iob.json currenttemp.json glucose.json profile.json [[--auto-sens] autosens.json] [meal.json] [--reservoir reservoir.json] --currentTime [currentTime]")
+      .usage("$0 <iob.json> <currenttemp.json> <glucose.json> <profile.json> [--auto-sens <autosens.json>] [--meal <meal.json>] [--reservoir <reservoir.json>] [--currentTime <currentTime>]")
       .option('auto-sens', {
         alias: 'a',
+        nargs: 1,
         describe: "Auto-sensitivity configuration",
-        default: true
-
+        default: false
```
scottleibrand: What do these default changes do?

bin/oref0-determine-basal.js
```diff
@@ -20,81 +20,70 @@ var determine_basal = require('oref0/lib/determine-basal/determine-basal');
...
-    if (params._.length > 5) {
-      autosens_input = params.autoSens ? params._[4] : false;
-      meal_input = params._[5];
-    }
-    if (params.meal && params.meal !== true && !meal_input) {
-      meal_input = params.meal;
+    if (inputs.length > 4) {
+      argv.showHelp( );
+      console.error('Too many arguments');
+      process.exit(1);
```
scottleibrand: Are you disallowing a formerly allowed arrangement of arguments here?

bin/oref0-determine-basal.js
```diff
@@ -20,81 +20,70 @@ var determine_basal = require('oref0/lib/determine-basal/determine-basal');
...
     }
-    var reservoir_input = params.reservoir;
 
-    if (!iob_input || !currenttemp_input || !glucose_input || !profile_input) {
```
scottleibrand: Is this check preserved somewhere?

jpcunningh: Yes, the check is preserved by the demand for at least 4 non-hyphenated arguments.

bin/oref0-determine-basal.js
jpcunningh: Yes,
--auto-sens and --meal are now required to pass auto-sens and meal input files rather than trying to figure out whether the user entered them as non-hyphenated arguments vs. the --auto-sens <file> or --meal <file> argument method. Now there is no ambiguity that has to be resolved.

oref0-pump-loop.sh and examples.sh are modified to account for the reduced complexity on the way arguments can be provided to the script.

bin/oref0-determine-basal.js
jpcunningh: Defaulting it to true causes determine_basal to crash if no argument was passed in because it was true vs. undefined.

Default it to false allows determine_basal to test for whether it's defined or false to determine if a valid value was provided.

bin/oref0-determine-basal.js
scottleibrand: So if we're going to introduce a change that breaks backwards compatibility (which I'm not opposed to on principle), we need to carefully think through the transition plan, particularly as to whether existing rigs will keep working after a `git pull` without a new run of `oref0-setup`.  Have you tested that, or can you verify that none of the remaining aliases use the old argument format, and that every call is done from within one of the scripts you updated?

jpcunningh: Good point on the aliases. I'll check those.

Any thoughts on how to go about making sure I find them all?

scottleibrand: It might be easiest to start removing the `device`s configured to use these scripts, and any reports configured to use those devices.  Then at some point we'll need to actually test on a real rig running dev that everything still works without the aliases...

scottleibrand:
Thanks.  I think this one is worth testing with both an existing dev rig, and with a clean dev install, to make sure the reports aren't a dependency for anything.

bin/oref0-determine-basal.js
scottleibrand: If you don't have a rig handy that you want to do a clean install on, I think we're ready to do so on at least one of ours soon.

jpcunningh:
I think I can get to that this weekend.

jpcunningh:
@scottleibrand, this is successfully running on a new rig with a clean dev install.

jpcunningh:
although looks like I need to check on why the CI failed.

jpcunningh:
@scottleibrand, I think this is ready.

It's rebased onto dev so the command-behavior-test could be updated to account for the new restrictions on oref0-determine-basal.

PR #1125, opened 2018-09-04
Migrate oref0-set-local-temptarget to yargs argument processing.

jpcunningh:


jpcunningh:
@scottleibrand, this is the last one.

I didn't migrate `oref0-mdt-trend.js` or `oref0-pebble.js` because I couldn't test them.

PR #1130, opened 2018-09-25
Use status.sh instead of status.js, if possible

cluckj:
The next version of openaps-menu will use a socket server interface for external control of the display, so we don't need to call status.js if the user has the new interface script (status.sh).


cluckj:
Yes; it shouldn't break anything while we get new openaps-menu stuff merged.

PR #1131, opened 2018-09-26
Update existing preferences rather than re-create.

alimhassam:


alimhassam:
needs proper testing. I tested the command manually, but not a full runagain or new setup yet.

bin/oref0-setup.sh
```diff
@@ -703,8 +703,8 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
         )
     fi
 
+    cp preferences.json old_preferences.json
```
alimhassam: I moved this line here so we can always have access to old_preferences. I was wondering if i should additionally date the file name?

PR #1134, opened 2018-10-03
Stop local go binaries before overwriting. Fixes #1049.

### LINKED ISSUE #1049: Kill go binaries before upgrading them
Downloading Go pump binaries from: https://github.com/ecc1/medtronic/releases/download/v2018.05.24/medtronic-386-spi.tar.xz
./
./openaps.jq
./cgmpage
./sniff
./historypage
./pumphistory
./listen
./mmtune
./fakemeter
./cgmhistory
./mdt
Installing Go pump binaries ...
‘/root/go/bin/cgmhistory’ -> ‘/usr/local/bin/cgmhistory’
‘/root/go/bin/cgmpage’ -> ‘/usr/local/bin/cgmpage’
‘/root/go/bin/fakemeter’ -> ‘/usr/local/bin/fakemeter’
‘/root/go/bin/historypage’ -> ‘/usr/local/bin/historypage’
‘/root/go/bin/listen’ -> ‘/usr/local/bin/listen’
‘/root/go/bin/mdt’ -> ‘/usr/local/bin/mdt’
‘/root/go/bin/mmtune’ -> ‘/usr/local/bin/mmtune’
‘/root/go/bin/openaps.jq’ -> ‘/usr/local/bin/openaps.jq’
‘/root/go/bin/pumphistory’ -> ‘/usr/local/bin/pumphistory’
cp: cannot create regular file ‘/usr/local/bin/pumphistory’: Text file busy
‘/root/go/bin/sniff’ -> ‘/usr/local/bin/sniff’
Couldn't copy go/bin

larryprice:
If the local go binaries are running, the copy will fail and setup will die. This PR asks the user if it's okay to kill any running binaries and then kills them if requested.

My editor also auto-fixed some inconsistent spacing; lmk if this should be undone.

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -258,8 +258,7 @@ function add_to_crontab () {
 function request_stop_local_binary () {
     if [[ -x /usr/local/bin/$1 ]]; then
         if pgrep -x $1 > /dev/null; then
-            read -p "Need to stop $1 to complete installation. OK? [Y]N" -r
-            if [[ $REPLY =~ ^[^Nn]*$ ]]; then
+            if prompt_yn "Need to stop $1 to complete installation. OK?" y; then
                 pgrep -x $1 | xargs kill
             fi
         fi
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -255,6 +255,32 @@ function add_to_crontab () {
     (crontab -l; crontab -l |grep -q "$1" || echo "$2" "$3") | crontab -
 }
 
+function request_stop_local_binary () {
+    if [[ -x /usr/local/bin/$1 ]]; then
+        if pgrep -x $1 > /dev/null; then
+            read -p "Need to stop $1 to complete installation. OK? [Y]N" -r
+            if [[ $REPLY =~ ^[^Nn]*$ ]]; then
+                pgrep -x $1 | xargs kill
+            fi
+        fi
+    fi
+}
+
+function copy_go_binaries () {
+    for gobinary in $HOME/go/bin/*; do
+        request_stop_local_binary `basename $gobinary`
+    done
+    request_stop_local_binary Go-mmtune
+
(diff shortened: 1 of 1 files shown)


bin/oref0-setup.sh
```diff
@@ -255,6 +255,32 @@ function add_to_crontab () {
     (crontab -l; crontab -l |grep -q "$1" || echo "$2" "$3") | crontab -
 }
 
+function request_stop_local_binary () {
+    if [[ -x /usr/local/bin/$1 ]]; then
+        if pgrep -x $1 > /dev/null; then
+            read -p "Need to stop $1 to complete installation. OK? [Y]N" -r
+            if [[ $REPLY =~ ^[^Nn]*$ ]]; then
```
scottleibrand: Is this logic correct?  I'm not sure what the second ^ does, but I think we want this to be something like `if [[ $REPLY =~ ^[Yy]*$ ]]; then`, right?

bin/oref0-setup.sh
```diff
@@ -255,6 +255,32 @@ function add_to_crontab () {
     (crontab -l; crontab -l |grep -q "$1" || echo "$2" "$3") | crontab -
 }
 
+function request_stop_local_binary () {
+    if [[ -x /usr/local/bin/$1 ]]; then
+        if pgrep -x $1 > /dev/null; then
+            read -p "Need to stop $1 to complete installation. OK? [Y]N" -r
```
alimhassam: Would it be better to use the common prompt_yn functions?

bin/oref0-setup.sh
larryprice: The `^` tells it to check for "not 'N' or 'n'" - my thought was to have "yes" be the default so users could just hit enter. I'll explore the `prompt_yn` function, which looks like a better solution.

bin/oref0-setup.sh
larryprice: Agreed and updated!

PR #1139, opened 2018-10-14
Wip/offline xdrip

sulkaharo:
Realized I hadn't PRd this, so here's a wip pass to replace my extremely hacky earlier integration. This was made against MASTER, causing the issue templates to pop up as well.

Let's discuss how / if we want to proceed


bin/oref0-get-ns-entries.js
```diff
@@ -0,0 +1,209 @@
...
+    	if (data) {
+//			console.error("CGM results loaded from xDrip");
+            processAndOutput(data);
+//	        var fs = require('fs');
+//			fs.writeFileSync(outputPath, JSON.stringify(data));
+		} else {
+//			console.error("Load from xDrip failed"); //", trying Spike");
+//			loadFromSpike();
+		}
+		process.exit(1);
```
scottleibrand: Indentation.

lib/oref0-setup/device.json
```diff
@@ -17,7 +17,7 @@
     "name": "ns-glucose",
     "extra": {
       "fields": "",
-      "cmd": "bash -c \"curl --compressed -m 30 -s $NIGHTSCOUT_HOST/api/v1/entries/sgv.json?count=1000 | json -e \\\"this.glucose = this.sgv\\\"\"",
+      "cmd": "bash -c \"oref0-get-ns-entries cgm/ns-glucose.json $NIGHTSCOUT_HOST $API_SECRET\"",
```
scottleibrand: Can we just delete this “device” instead of updating it?

sulkaharo: Sure if it's not used anywhere

dramageek:
I'm horrible at reading code, but it looks like this does what we're talking about in #1064 ? 
If so, I'm in support of it, and we can close that issue. If this is what I think it is, it's much more graceful than trying to duplicate existing functionality into monitor-xdrip.sh

dramageek:
Haven't tried it, but the 192.168.43.1 address may not always be correct- in my situation it would work on my phone's wifi hotspot, but the bluetooth link goes up on 192.168.44.* .  In my pull request above I ended up using the local gateway address (so it'd work for either), but I never tested if that could have other downsides.

sulkaharo:
@dramageek Yes this seems to do what #1064 discusses. Wonder if using the default gateway IP is the best way to go? I bet it works most of the time but I wonder if there's an Android fork that does something funky

sulkaharo:
Looks like the npm network package is working correctly on our devices

scottleibrand:
Resolved merge conflicts.  Is this still WIP?  Is it something we want to merge to dev for 0.7.0?

scottleibrand:
@sulkaharo @dramageek any comments on the status of this PR?

danamlewis:
@sulkaharo @dramageek can you update on status? trying to decide which remaining open PRs should be targeted toward the upcoming 0.7.0 release...

dramageek:
This should be good to go, but still needs testing. I've had a couple of hardware failures on both my rigs, so I'm unable to test it.

sulkaharo:
@danamlewis We've had it on our production rig and after the latest fixes in the PR in January, we haven't had any issues, including offline looping working in a boat, plane and at school. The only hassle this hasn't fixed is the Phone hotspot not turning on automatically when the phone is rebooted.

scottleibrand:
Closing in favor or #1248

Issue #1022, opened 2018-05-28
Need to add a bolus status check before changing pump clock to sync to rig time


danamlewis:
**An issue occurs if pump time (set clock) is changed during a bolus**: additional bolus record(s) is created in pump history, which would cause OpenAPS to perceive double the IOB from that bolus. (Only a single bolus is delivered to the body, however). 

**Details**: Issues was spotted with a 1u easy bolus (manual bolus) which then showed up in pump as multiple boluses; and in NS as a 2u bolus. OpenAPS recorded IOB prior to bolus as 0.845 and after the 1u bolus (seeing the 2u record) as 2.894. Reservoir levels also confirmed it was actually a single 1u bolus delivered. 

**Proposed solution**: add a bolus status check and do not change clock if a bolus is active.

hilarykoch:
Yes. This happened to my son the other night. He boluses for one unit, but NS and IOB indicated two units were given, and bolus history also indicated multiple boluses. But reservoir levels confirmed only one unit was delivered. However, OpenAPS registered the delivery as two units and continued to make decisions using those two units as part of IOB. Logs showed a time change at the same time of bolus. 
<img width="1218" alt="ns" src="https://user-images.githubusercontent.com/11912289/40637154-2d9f9318-62d1-11e8-9554-7e68851a804a.png">
<img width="1259" alt="iob pre" src="https://user-images.githubusercontent.com/11912289/40637156-30868604-62d1-11e8-96aa-38481d48467b.png">
<img width="1257" alt="iob post" src="https://user-images.githubusercontent.com/11912289/40637157-32d0c2da-62d1-11e8-994d-f3abe6741a75.png">
![img_7400-1](https://user-images.githubusercontent.com/11912289/40637241-aab0c868-62d1-11e8-83ad-e08c53382045.jpeg)
![img_7398-1](https://user-images.githubusercontent.com/11912289/40637242-aac1f638-62d1-11e8-9f96-73635fcc1077.jpeg)
![img_7399-1](https://user-images.githubusercontent.com/11912289/40637243-aad88e2a-62d1-11e8-962a-6f081050c619.jpeg)
![img_7402-1](https://user-images.githubusercontent.com/11912289/40637244-aaecf072-62d1-11e8-8626-a610bc52e083.jpeg)

jmcrawford:
[...]

PR #1140, opened 2018-10-14
#1022 - check status_ESC prior to set_clock

jmcrawford:


bin/oref0-pump-loop.sh
```diff
@@ -212,6 +212,10 @@ function smb_reservoir_before {
     echo -n " is within 90s of current time: " && date +'%Y-%m-%dT%H:%M:%S%z'
     if (( $(bc <<< "$(to_epochtime $(cat monitor/clock-zoned.json)) - $(epochtime_now)") < -55 )) || (( $(bc <<< "$(to_epochtime $(cat monitor/clock-zoned.json)) - $(epochtime_now)") > 55 )); then
         echo Pump clock is more than 55s off: attempting to reset it and reload pumphistory
+	        # Check for bolus in progress
+                smb_verify_status && \
```
scottleibrand: Probably should use a couple of `if` clauses here: I don't know what the `command1 && # comment \n command2` behavior is.  Also, looks like your indentation doesn't match the surrounding code.

jmcrawford: Consolidated comments, and fixed indentation...

https://xkcd.com/1513/

bin/oref0-pump-loop.sh
```diff
@@ -212,6 +212,9 @@ function smb_reservoir_before {
     echo -n " is within 90s of current time: " && date +'%Y-%m-%dT%H:%M:%S%z'
     if (( $(bc <<< "$(to_epochtime $(cat monitor/clock-zoned.json)) - $(epochtime_now)") < -55 )) || (( $(bc <<< "$(to_epochtime $(cat monitor/clock-zoned.json)) - $(epochtime_now)") > 55 )); then
         echo Pump clock is more than 55s off: attempting to reset it and reload pumphistory
+	# Check for bolus in progress and issue 3xESC to back out of pump bolus menu
+        smb_verify_status && \
+        try_return timerun openaps use pump press_keys esc esc esc | jq .completed | grep true && \
```
alimhassam: since this is targetting the dev branch, shouldn't this use the `mdt button` go command rather than `openaps use pump`

scottleibrand:
Yes, it should.

jmcrawford:
I've tested as far as ensuring that time is still reset when needed, and that it successfully backs out of any pump menus to do so.

PR #1141, opened 2018-10-16
Fix oref0 setup existing prefs

jpcunningh:
This address the failure identified at https://gitter.im/openaps/oref0?at=5bc4c006e65a634336e248c4


jpcunningh:
I have not had a chance to test this yet, but I believe it will solve the referenced issue with a clean install.

mhaeberli:
Scott - I'd vote for merging (and reverting only if bad problems appear) as it appears maybe at least two users in the last few hours could have benefited.  But I defer to your leadership. Thx.

PieterGit:
this breaks on a freshly installed rig, because jessie has jq 1.4 that has no slurpfile. see
https://gitter.im/openaps/oref0?at=5c24b0f492cf4d2242001be8

PR #1146, opened 2018-11-18
Make it possible to run on OpenBSD

viq:
Mostly just simple changes fixing finding `bash`, and system detection for `date` which I believe should work as well for FreeBSD.

The `seq` changes are quick and dirty and spit out warnings about not finding `seq` which I currently don't mind, but can try to fix the way that `date` was dealt with if it's preferred.

### DIFF
--- a/bin/all-autosens-history.sh
+++ b/bin/all-autosens-history.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 oref0_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
 interval=${1:-96}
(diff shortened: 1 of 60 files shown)


scottleibrand:
Thanks.  Have you already confirmed this still works and doesn't introduce any weirdness on Raspbian/Debian, or do we need someone to do that?

viq:
I did not yet test on Debian, so testing there would be a good idea.

Having said that, `#!/usr/bin/env bash` is a pretty standard way to do it, and already at least a couple of scripts in there use `#!/usr/bin/env node` from what I've seen.

PieterGit:
@viq I think a syntax of `seq || gseq` is quite ugly and should be fixed in the same way as `date` vs `gdate` (with an alias). Gitter link: https://gitter.im/openaps/oref0?at=5c2c98def6166a3027a8e236

scottleibrand:
I’ve found the alias approach to be somewhat brittle. I don’t mind “ugly” if it prevents failure modes.

PR #1152, opened 2018-11-25
ESLint configured for JavaScript 5 on Node

edwingustafson:
ESLint installed in the project and configured for JavaScript version 5 running on node. (Excludes tests and www.) (With NPM script and make target, both independent of existing scripts and targets.)


edwingustafson:
Main intention is to show hints in the code editor.

Added to `scripts` and `Makefile` to document how to run it manually.

Definitely not intended to run as part of any existing build, deploy, or integration process.

PR #1154, opened 2018-11-26
Revert "ESLint configured for JavaScript 5 on Node"

scottleibrand:
Reverts openaps/oref0#1152 due to breakage of `npm run global-install` - https://gitter.im/nightscout/intend-to-bolus?at=5bfb7577a115c91ef7932640

@edwingustafson can you see if you can figure out what about this commit is causing the issue?


scottleibrand:
Yeah, I don't know why that's a problem.  I can't really test it effectively myself, as when I don't seem to have eslint installed:

```
~/src/oref0 $ npm run lint

> oref0@0.7.0-dev lint /home/sleibrand/src/oref0
> eslint .

sh: 1: eslint: not found
```

But I agree, just leaving the npm stuff reverted and re-PR'ing the other stuff should be safe.

PR #1157, opened 2018-11-29
Use absolute path for ~/.profile

viq:
Use absolute path to `.profile` for consistency and since we may be in a different directory. Should fix error reported on gitter https://gitter.im/nightscout/intend-to-bolus?at=5c000746f59704348e22b3ad


viq:
Should I re-do this against `dev` ?

scottleibrand:
Yeah, that would be easier, thx.

viq:
Ah, looks like this is not necessary for `dev` - https://github.com/openaps/oref0/blob/dev/bin/oref0-setup.sh#L1193

PR #1165, opened 2018-12-08
Fix ns-loop for default NS auth denied

jpcunningh:


jpcunningh:
@scottleibrand, I tested with NS set to readable.

It appears to have no effect if NS is set to readable.

sulkaharo:
It's possible to set the global key to have less read access than non-authenticated users but not sure this is a case happening live?

jpcunningh:
@sulkaharo, no key (global or token) was being provided for the treatments collection query in this case.

If NS was setup to deny read access to any non-authenticated user, ns-loop wasn't able to query the treatments collection. I'm running my NS now with deny as the default access. I'll fix anything else I find over the next few days. :smile:

PR #1169, opened 2018-12-11
Fix oref0-upload-profile when NS default auth role denies read

jpcunningh:


jpcunningh:
The unmodified code could not query the current profile from NS if NS default access is set to denied and the user uses tokens because the token has to be provided as part of the query parameters instead of a header value.

PR #1170, opened 2018-12-13
Declare loop variables once and only once

edwingustafson:
Declare loop variables at first use. Do not redeclare on subsequent uses.


scottleibrand:
Does this have any effect on whether the variables are reset for nested loops?  Or is the presence or absence of `var` irrelevant for that?

edwingustafson:
No effect since I find each `for` loop diligently assigning its own initial value.

(To avoid loop variable reuse, some or all of the loops could be rewritten using `forEach` and the other array methods that take a callback function. That way JavaScript function scope makes each pass a sealed unit.)

PR #1150, opened 2018-11-24
Fix npm install/link, xdrip cgm sets bt_offline, file tests to preven…

rkresha:
…t stderr output, add bnep0 interface if doesn't exist, enhance bnep0 cycling logic, cleanup/standardization

Fix npm install/link, xdrip cgm sets bt_offline, file tests to prevent avoidable stderr output, add bnep0 interface if doesn't exist, enhance bnep0 cycling logic, cleanup/standardization of parsing files

Fixes #1145 for me

PR #1175, opened 2018-12-25
Fix for loss of wifi issue on Raspbian

cluckj:
Removes lines that edit `/etc/network/interfaces` in `oref0-setup.sh`, which cause dhcpcd to crash on the latest Raspbian. See: https://github.com/openaps/oref0/pull/1150#issuecomment-449660472


scottleibrand:
Merging this for now to unbreak things. 

Longer term, do we want to do this on Edison but not Pi?

PR #1176, opened 2018-12-25
restore ww pumps for 0.7.0-dev

PieterGit:
With a fresh install with a WW pump, the rig could not communicate to the pump.

- removing python `oref0_init_pump_comms.py` and `oref0_subg_ww_radio_parameters.py` (not needed with Go libaries)
- use .travis with latest Node 8 LTS and Node 10 LTS versions
- docfix: update version numbers of Go libraries to version with tokenbased authentication
- remove `--ww_ti_usb_reset`
- remove oref0 dependencies on `pump.ini`
- fix `jq` install problem, and don't install master version of oref0 by default on `bin/openaps-packages.sh`. It will be installed by `oref0-setup.sh`
- read pump.ini stuff from `preferences.json` instead of pump.ini
- mmtune fix if does not exist on install
- mmtune fix: `die ` in `bin/oref0-mmtune.sh`  if `monitor/mmtune.json` is empty or does not exist
- fix for `mmeowlink` if you're not a `mdt` cgm user
- removed `radio_locale pump.ini ` hacks 
- update npm libraries

Status: Ready for merge to dev, when @scottleibrand thinks it's ready.


bin/oref0-setup.sh
```diff
@@ -1025,8 +1030,6 @@ if prompt_yn "" N; then
         else
             # radio_locale requires openaps 0.2.0-dev or later
             openaps device add pump mmeowlink subg_rfspy $ttyport $serial $radio_locale || die "Can't add pump"
```
scottleibrand: These lines normally add the pump stanza for MDT CGM users. We only do so manually for non-Enlite users because they’re using the Go based pump comms and not the old openaps toolkit.

bin/oref0-setup.sh
```diff
@@ -831,6 +805,11 @@ if prompt_yn "" N; then
     echo Running apt-get autoclean
     sudo apt-get autoclean
 
+    # create pump.ini
+    echo '[device "pump"]' > pump.ini
+    echo "serial = $serial" >> pump.ini
+    echo "radio_locale = $radio_locale" >> pump.ini
```
scottleibrand: Would it be better to move just this one line after the else block?

PieterGit: it's required for `nightscout autoconfigure-device-crud ` to have a good pump.ini

bin/oref0-setup.sh
PieterGit: i'm just removing support for ww_ti_usb_reset and removing all the `oref0_init_pump_comms.py` references. I don't understand what you mean with the comment above

bin/oref0-setup.sh
scottleibrand: That doesn’t mean we should create it manually and then also create it with openaps device add pump.

PieterGit: i just moved your lines from https://github.com/openaps/oref0/blame/dev/bin/oref0-setup.sh#L1071 a bit more up (they were in an `else ` part of Medtronic CGM). radio_locale is generated several times and causes invalid pump.ini with a fresh 0.7.0-dev install. You created these lines with https://github.com/openaps/oref0/commit/13068f32ef3829967d2c4a0cf82e68131cf0ec71

scottleibrand: The first two of those lines are normally created by the openaps device add pump commands. If you’re not replacing the openaps commands, you only need to move the third line outside the else. I’m not sure if adding all three adds the first two twice, or if openaps device add ignores them if they already exist.

bin/oref0-setup.sh
```diff
@@ -284,7 +281,9 @@ function copy_go_binaries () {
 
 function move_mmtune () {
     request_stop_local_binary Go-mmtune
-    mv /usr/local/bin/mmtune /usr/local/bin/Go-mmtune || echo "Couldn't move mmtune to Go-mmtune"
+    if [ -f /usr/local/bin/mmtune ]; then
+      mv /usr/local/bin/mmtune /usr/local/bin/Go-mmtune || echo "Couldn't move mmtune to Go-mmtune"
+    fi
```
scottleibrand: Do we need an else here that dies if mmtune doesn't exist?

bin/oref0-setup.sh
```diff
@@ -986,110 +956,6 @@ if prompt_yn "" N; then
         cd $directory || die "Can't cd $directory"
     fi
 
-    # we only need spi_serial and mraa for MDT CGM, which Go doesn't support yet
```
scottleibrand: Why are you removing this?  Isn't it still needed for MDT CGM?

bin/oref0-setup.sh
```diff
@@ -1137,10 +1003,11 @@ if prompt_yn "" N; then
         # Add module needed for EdisonVoltage to work on jubilinux 0.2.0
         grep iio_basincove_gpadc /etc/modules-load.d/modules.conf || echo iio_basincove_gpadc >> /etc/modules-load.d/modules.conf
     fi
-    if [[ ${CGM,,} =~ "mdt" ]]; then # still need this for the old ns-loop for now
-        cd $directory || die "Can't cd $directory"
-        do_openaps_import $HOME/src/oref0/lib/oref0-setup/edisonbattery.json
-    fi
+#DEPRECATED?
```
scottleibrand: Is there another way to do MDT CGM now?

bin/oref0-setup.sh
PieterGit: implemented die in the else statement

bin/oref0-setup.sh
PieterGit: nope, deprecated too much. restored

PieterGit:
@scottleibrand I think i fixed all your open review remarks. Can you please re-rereview. I will test on a fresh edison and rpi later.

scottleibrand:
Looks like there's a conflict between @cclauss' and @PieterGit's recent changes.  
@PieterGit did:
```
node_js:
  - 8
  - 10
```
whereas @cclauss did:
```
matrix:
  include:
    - node_js: "8.11.1"
      <<: *node_js-steps
```
Which do y'all prefer?

cclauss:
Perhaps:
```
matrix:
  include:
    - node_js: “8”
      <<: *node_js-steps
    - node_js: “10”
      <<: *node_js-steps
    - node_js: “node”
      <<: *node_js-steps
```

PieterGit:
Done testing for today. Installs fine, but I got some problems with duplicate `radio_locale` in `pump.ini` and some others stuff I want to check. Will investigate further. Do not merge this PR yet.

PieterGit:
Tested on a explorer board and with WW-pump. Works. Ready for further test, review and merge.

bin/oref0-setup.sh
```diff
@@ -1040,39 +1011,37 @@ if prompt_yn "" N; then
...
+              echo -n "Cloning subg_rfspy: "
+              (cd $HOME/src && git clone https://github.com/ps2/subg_rfspy) || die "Couldn't clone oref0"
             fi
-
-            # Hack to check if radio_locale has been set in pump.ini.
-            # It will remove empty line at the end of pump.ini and then append radio_locale if it's not there yet
-            grep -q radio_locale pump.ini ||  echo "$(< pump.ini)" > pump.ini ; echo "radio_locale=$radio_locale" >> pump.ini
         fi
-    else
+    else # in all other case create a pump.ini with serial and radio_locale. TODO: can we skip creating a pump.ini. I can't find references of it's being used
```
cluckj: I think that xdrip-js uses `pump.ini`, we should let them know before removing it: @efidoman @jpcunningh @thebookins

PieterGit: i don't want to suggest to remove it now. As far as I can see I removed all the references to the pump.ini from the oref0 code base. I hope @efidoman @jpcunningh @thebookins can say if their software somehow uses pump.ini. I just checked Lookout, but can't find any reference: https://github.com/xdrip-js/Lookout/search?q=pump.ini&unscoped_q=pump.ini

cluckj: The only use might be for `fakemeter`, which is here: https://github.com/xdrip-js/Logger/blob/f7228485c7a66300f280ac789b5dce0252ffa279/xdrip-get-entries.sh#L245

PieterGit: Created https://github.com/xdrip-js/Logger/issues/137 for that

PieterGit:
rig is running fine for more than 2 days. I think this PR is ready to merge to dev. 
@scottleibrand can you review and merge to dev when you think it's ready?

PieterGit:
I resolved the conflict with the dev branch. Rig is used as primary rig, and didn't find issues for a week.
@scottleibrand Can you please review and merge to dev if you think it's ok?

PieterGit:
@cclauss thanks. fixed it with the Github UI. re-applied. 
@scottleibrand please merge to dev

scottleibrand:
This LGTM.  Once we iron out the issues with dev and decide if webpack is staying or going, I'll fix the merge conflicts and confirm tests pass before merging.

PieterGit:
Thanks for merging. Up to the next PR.

efidoman:
I remembered that having this variable set is no longer necessary for fakemeter, so I tested Logger with that line deleted and fakemeter works fine. In addition, the fakemeter processing has moved in the latest Logger and OpenAPS dev branches to process as part of the OpenAPS loop. I'll update the Logger code and remove this line.

efidoman:
All references to pump.ini have been removed from the Logger dev branch.

PR #1179, opened 2018-12-28
Re-adding bnep0 check into oref0-setup.sh

rkresha:
@scottleibrand  
Here is the re-add that shouldn't cause an issue with rpi. @cluckj would you be willing to validate this?


cluckj:
:+1: Tested & working as expected on the pi; thanks!

scottleibrand:
Do we need a test on Edison as well?

rkresha:
> Do we need a test on Edison as well?

Based on my previous work on this I feel confident with the changes without actually testing them. Really your comfort level though. Please advise @scottleibrand

rkresha:
@scottleibrand works as expected. Edison is good to go.

Issue #1149, opened 2018-11-24
cgm-loop.log: Not an openaps environment, run: openaps init in papertrail logs; similar in normal logs


mhaeberli:
**Describe the bug**

After installing dev branch on a freshly installed Edison, at least in the nightscout/G5 configuration, repeated error messages are seen in the logs, of the form:

(in papertrail, for example):
`cgm-loop.log: Not an openaps environment, run: openaps init`

Looping seems still to be proceeding normally.

**To Reproduce**
1. Flash a fresh Edison
2. Follow normal installation procedures, but hit ctrl-C at the point where it offers to finish installing 'master' oref0
3. Reboot Edison, set new password as requested
4. Install dev branch per instructions
5. Configure and run dev branch via standard oref0-setup interactive setup

6. Problem can be 'fixed' by adding:
`  "cgm_loop_path": "/root/myopenaps/",` as the first entry in preferences.json

**Expected behavior**
I would expect the above needed cgm_loop_path parameter in preferences.json to be set correctly, or the system otherwise to avoid reporting a (spurious) error.

**Smartphone (please complete the following information):**
Not Applicable, but iPhone 7 with latest production iOS

**Setup Information (please complete the following information):**
* Medtronic 723, v.2.4A firmware
* CGM type (e.g. Dexcom G6 via Dexcom Share and Nightscout)
* Edison/Explorer Board rig)
* oref0 dev latest commit hash - 7f3526e190c9455016c71f664de6e6bd1dce934a

**Additional context**
N/A

mhaeberli:
PS props to @cluckj for guidance !

cluckj:
Found the bug: oref0-setup.sh does not set cgm_loop_path for "G5" cgm type.

jaylagorio:
Fix included in #1217

danamlewis:
Resolved in https://github.com/openaps/oref0/pull/1211 as noted - thanks all.

PR #1182, opened 2019-01-12
Adding CGM support for Dexcom touchscreen receivers

jaylagorio:
* Adds ``g6`` and ``g6-upload`` support for Dexcom touchscreen CGM receivers. Note that these receivers work with G5 transmitters as well.

Must merge these two in order first for this to work without breaking things:
* https://github.com/openaps/dexcom_reader/pull/19
* https://github.com/openaps/openaps/pull/139


bin/oref0-setup.sh
```diff
@@ -682,6 +684,9 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
     if [[ ${CGM,,} =~ "g5" || ${CGM,,} =~ "g5-upload" ]]; then
         openaps use cgm config --G5
         openaps report add raw-cgm/raw-entries.json JSON cgm oref0_glucose --hours "24.0" --threshold "100" --no-raw
+    elif [[ ${CGM,,} =~ "g6" || ${CGM,,} =~ "g6-upload" ]]; then
+        openaps use cgm config --G6
+        openaps report add raw-cgm/raw-entries.json JSON cgm oref0_glucose --hours "24.0" --threshold "100" --no-raw
```
cluckj: Will need to add `set_pref_string .cgm_loop_path "$directory"` in this function to make sure `oref0-monitor-cgm.sh` runs in the right place. See: https://github.com/openaps/oref0/issues/1149 and https://github.com/openaps/oref0/pull/1211

jaylagorio: Thanks for fixing this bug! I ran into it a while ago and gave up on it because I couldn't track it down in the time I had. I'm closing this PR and opening a new one that both adds setup support for touchscreen receivers, fixes this bug, and makes some changes to `oref0-cron-every-minute` to make the CGM loop work with the new receivers too.

jaylagorio:
Replacing with #1217

PR #1188, opened 2019-01-27
Apache chainsaw log

tzachi-dar:
This PR will also create a common log file for all openaps logs.

This file has the format:

date logger log message.

This file can be read by apache chainsaw, which is an open source tool for viewing logs.
I'll soon send a PR to the docs files that explains how to use this file.

If needed we can do the following changes:
1) Keep the logs for a longer time (for example weekly).
2) Make the logging to this files optional. 

All comments are welcomed.


tzachi-dar:
See https://github.com/openaps/docs/pull/1410 on how to use apache chainsaw on openaps.

tzachi-dar:
Thanks for review.
Obviously, messing with existing logs, that other depend upon is not a good idea.

PR #1189, opened 2019-01-31
Travis CI: Add current Node.js and Python 2.7, 3.7

cclauss:
Blocked by #1190

PR #1193, opened 2019-02-09
Find Python 3 syntax errors and undefined names

cclauss:
Trying something new...

### DIFF
--- a/.github/main.workflow
+++ b/.github/main.workflow
@@ -1,0 +1,8 @@
+workflow "New workflow" {
+  on = "push"
+  resolves = ["Find Python 3 syntax errors and undefined names"]
+}
+
+action "Find Python 3 syntax errors and undefined names" {
+  uses = "cclauss/Find-Python-syntax-errors-action@master"
+}


.github/main.workflow
```diff
@@ -0,0 +1,8 @@
+workflow "New workflow" {
+  on = "push"
+  resolves = ["Find Python 3 syntax errors and undefined names"]
+}
+
+action "Find Python 3 syntax errors and undefined names" {
+  uses = "cclauss/Find-Python-syntax-errors-action@master"
```
scottleibrand: What does this reference?  Is it a new file you forgot to `git add`?

cclauss: It references https://github.com/cclauss/Find-Python-syntax-errors-action but I think that you need to be in the GitHub Actions beta program for this to work.  It works great on my repos but unfortunately I don't think it will work on yours until GitHub Actions graduates from beta state.

scottleibrand: Ok, in that case is there another approach we should try? Or should we close this PR and reopen when it’s GA?

cclauss:
Closed in favor of #1189

PR #1194, opened 2019-02-10
don't reduce CR for low temptargets

scottleibrand:
With the low_temptarget_lowers_sensitivity option enabled, oref0 doses more aggressively when an "eating soon" target is set.  However, in the current implementation, this becomes much *more* aggressive with COB, as the adjusted sensitivityRatio is used to adjust CR as well as basal/IOB and ISF.  This results in too-aggressive dosing when a low temptarget is started or left running after carb entry.

This PR adjusts the low_temptarget_lowers_sensitivity behavior so that the modified sensitivityRatio does not affected CR, yielding much more intuitive and predictable behavior before vs. after carb entry, and avoiding most of the risk of a low target causing a postprandial low.


scottleibrand:
We've been running this branch for 2 months now and have seen far fewer post-meal hypos after setting low targets.

scottleibrand:
Test is passing for me: re-running.

scottleibrand:

PR #1195, opened 2019-02-10
categorize UAM as basal if at least 1h of carb absorption observed

scottleibrand:
We've seen problems with autotune failing to converge, and instead constantly trying to reduce basals until it hits safety limits, because too much data that would otherwise indicate basals are too low is being excluded as UAM.  We were able to avoid this issue by using --categorize-uam-as-basal=true, so it appears that categorize-uam-as-basal should be the default behavior at least when meals are being properly announced and CSF exclusion is working.  To preserve the UAM exclusion for folks who don't routinely announce meals, this PR sets the --categorize-uam-as-basal=true behavior as the default whenever at least 1h per day of CSF exclusion is observed, which will be nearly all days where one or more meals is announced.


jpcunningh:
👍  We have had very similar experiences, and always use --categorize-uam-as-basal=true.

Unless we consistently miss an announcement at the same time every day (like coffee 😄), we don't see a negative effect of basals creeping up too high due to autotune's dampened response to basal changes.

zgohr:
We've got the same issue - without the setting, autotune consistently suggested decreased basals - so I think this would help a lot. 

_Semi-related_
Since turning autotune off, we're still taking manually-ran autotune suggestions with a bit of a grain - it's still suggesting basals that _seem_ to be too low. Combined with autosens, ISF dropping because "extra" basal required means ISF continues to be pushed down during the day which causes a roller coaster.

scottleibrand:
@zgohr have you been running with or without `--categorize-uam-as-basal=true`?

zgohr:
When I referred to running autotune manually, we're running it with the setting `true`.

scottleibrand:

PR #1172, opened 2018-12-16
Missing and redundant variable declarations

edwingustafson:
Add or remove variable declarations for variables ESLint found to be undeclared or redeclared

PR #1198, opened 2019-02-11
Declaration of UAMpredBG breaks determine-basal.js

cluckj:
https://github.com/openaps/oref0/pull/1172/files#r255598505


scottleibrand:
Would love to understand why this change broke things, but merging for now to fix them.

PR #1202, opened 2019-02-15
Convert json to jq

jpcunningh:
This PR is still a work in progress.

This PR converts most of the calls to the json command to the jq command.  The json command is very resource intensive on the Raspberry Pi Zero. I believe the conversion to the much more effecient jq command will make a noticeable difference in CPU cycles used on the Pi Zero.

### DIFF
--- a/bin/mm-format-ns-glucose.sh
+++ b/bin/mm-format-ns-glucose.sh
@@ -30,5 +30,5 @@ cat $HISTORY | \
     | .date = ( ( [ $time, "Z" ] | join("") ) | fromdateiso8601 ) * 1000 + $msec
     | .type = if .name and (.name | test("GlucoseSensorData")) then "sgv" else "pumpdata" end
     | .device = "openaps://medtronic/pump/cgm"
-    '$NSONLY' ]' > $OUTPUT
+    '"$NSONLY"' ]' > $OUTPUT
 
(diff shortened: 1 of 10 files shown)


bin/mm-format-ns-glucose.sh
```diff
@@ -13,19 +13,22 @@ Format Medtronic glucose data into something acceptable to Nightscout.
 EOT
 
 NSONLY=""
-test "$1" = "--oref0" && NSONLY="this.glucose = this.sgv" && shift
+test "$1" = "--oref0" && NSONLY="| .glucose = .sgv" && shift
```
scottleibrand: what does this `| .` syntax do?

jpcunningh: The `|` character is similar to a Unix `|`.  The output of the filter to its left is the input to the filter to its right.

The `.` character refers to the current filter input object.

The snippet above doesn't provide the context because NSONLY is simply tagged onto the end of a long set of filters if its presence is required. The pipe allows it to take the output of the rest of the filters.

scottleibrand:
LMK when you think this is adequately tested and ready for review and merge.

jpcunningh:
I went through each of the changes again and tested them on my running rig. I found a couple of latent issues, and corrected them.

I think it is ready to merge.

scottleibrand:
Looking good so far in on-rig testing.  Ready to merge to dev as soon as tomorrow if no one else wants to test it out first.

PR #1203, opened 2019-02-17
Travis CI: Run node_js tests on "8", "10", and "node"

cclauss:
As discussed at https://github.com/openaps/oref0/pull/1176#issuecomment-464483637

“__node__” is currently v11.10.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,9 @@ python-steps: &python-steps
 
 matrix:
   include:
-    - node_js: "8.11.1"
+    - node_js: "8"
+      <<: *node_js-steps
+    - node_js: "10"
       <<: *node_js-steps
     - node_js: "node"  # current version of Node.js on Travis CI
       <<: *node_js-steps


cclauss:
Closed in favor of #1176

PieterGit:
@scottleibrand @cclauss Note that with Nighscout we had problems at the beginning of a new Node version with broken Travis builds that were caused by Node and not by Nightscout. So therefore we chose to only support node LTS releases and only to use LTS releases in Travis. Current node 11 works fine with oref0, so there is no need to disable it, but I would recommend to only enable those builds if they are stable enough not to break the oref0 Travis builds.

PieterGit:
@cclauss I don't mind this PR already being merged to dev. That's up to @scottleibrand

cclauss:
On the comments above on LTS, it is possible to add under __matrix:__
```
matrix:
  allow-failures:
    node_js: "node"
  include:
```
This will make that run non-critical to build success or failure.  We can still see if oref0 is incompatible with the current node but if it is, that will not break the build.

PieterGit:
@cclauss that sounds like a good solution, to allow failures of the latest Node.

PR #1204, opened 2019-02-18
Uncolor fix

tzachi-dar:
One more coloring pattern removed.


tzachi-dar:
Thanks for the fast review.

scottleibrand:
I like easy ones.  ;-)

PR #1205, opened 2019-02-18
Travis CI: Run shellcheck on shell scripts in allow_failures mode

cclauss:
Output: https://travis-ci.org/openaps/oref0/jobs/494772101#L382


scottleibrand:
Did you find any of this output useful?  I'm not sure it found anything worth fixing, but it definitely found all kinds of stuff that doesn't really matter and therefore I don't care to fix (but wouldn't mind if someone else wants to)...

PR #1212, opened 2019-02-26
ability to add script plugins to oref0-pump-loop

efidoman:
This PR adds the ability to inject custom script(s) to run synchronously at the end of every oref0-pump-loop. Any script files with executable flag set will be run as follows:

If the script file is placed in the plugins/every folder then the script runs at the end of every loop.
If the script file is placed in the plugins/once folder then the script once at the end of the current loop.

Script files placed in the plugins/once folder will be removed once run. This allows for a use case such as the following:

Logger and Lookout both call a pump utility called fakemeter that helps out users in offline mode by sending the current CGM (Dexcom) glucose value as a meter glucose value. This shows the glucose on the pump every 5 minutes and allows you to quickly see the current glucose value on the pump. Since this script requires communication with the pump, it will collide with oref0 pump communications on occasion. This oref0-pump-loop plugin capability will allow Logger and Lookout to run fakemeter synchronously at the end of the pump loop and avoid any radio collisions.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -87,12 +87,17 @@ main() {
             touch /tmp/pump_loop_success
             echo Completed oref0-pump-loop at $(date)
             update_display
+            run_plugins
             echo
         else
             # pump-loop errored out for some reason
             fail "$@"
         fi
     fi
+}
+
+function run_plugins {
+    :
 }
 
 function update_display {
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -101,14 +101,16 @@ function run_script() {
 
   if [[ -x "$file" ]]
   then
-    echo "Running script file ($file)"
+    echo "Running plugin script ($file)... "
     timeout 60 $file
+
+    thedate=$(date)
+    echo "Plugins Completed - $thedate"
+
     if [[ "$2" == "-d" ]]
     then
-      echo "Removing script file ($file)"
+      #echo "Removing script file ($file)"
       rm $file
-    else
-      echo "Not Removing script file ($file)"
     fi
   fi
 }
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -96,8 +96,34 @@ main() {
     fi
 }
 
+function run_script() {
+  file=$1
+
+  if [[ -x "$file" ]]
+  then
+    echo "Running script file ($file)"
+    timeout 60 $file
+    if [[ "$2" == "-d" ]]
+    then
+      echo "Removing script file ($file)"
+      rm $file
+    else
+      echo "Not Removing script file ($file)"
+    fi
+  fi
+}
+
+
 function run_plugins {
-    :
+        once=plugins/once
+        every=plugins/every
+        mkdir -p $once
+        mkdir -p $every
+        echo "scripts placed in this directory will run once after curent loop and be removed" > $once/readme.txt
+        echo "scripts placed in this directory will run after every loop" > $every/readme.txt
+        find $once/* | while read file; do run_script "$file" -d ; done
+        find $every/* | while read file; do run_script "$file" ; done
+
 }
 
 function update_display {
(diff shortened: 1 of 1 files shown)


bin/oref0-pump-loop.sh
```diff
@@ -95,6 +96,38 @@ main() {
...
+
+function run_plugins {
+        once=plugins/once
+        every=plugins/every
+        mkdir -p $once
+        mkdir -p $every
+        echo "scripts placed in this directory will run once after curent loop and be removed" > $once/readme.txt
+        echo "scripts placed in this directory will run after every loop" > $every/readme.txt
+        find $once/* | while read file; do run_script "$file" -d ; done
+        find $every/* | while read file; do run_script "$file" ; done
```
scottleibrand: These will attempt to run the readme.txt as a script.  We should probably limit the find to just `*.sh` files.

efidoman: The run_script function checks for -x (executable flag) so it should skip any others. It has skipped all others in my testing so far.

scottleibrand: Ah, that works.  Added a comment noting that's how it works.

scottleibrand:
Code looks good.  If someone can post test results, I think we can merge this.

efidoman:
needs a quick review again. I found an issue while testing where the plugins didn't run upon any loop failure. The intention is for plugins to run at the end of every pump loop regardless of pass or fail. Also, I updated the comments and logging a bit.

efidoman:
Here's my latest test results that include the Logger PR to leverage this new plugin capability for running fakemeter:

```
o bolus needed. Settings less than 15 minutes old. Your instance of oref0 [0.7.0-dev, dev] is out-of-date by 28 commits: you may want to consider updating.
You are currently running a development branch of oref0.  Such branches change frequently.
Please read the latest PR notes and update with the latest commits to dev before reporting any issues.
Completed oref0-pump-loop at Wed Feb 27 08:28:30 EST 2019

Running plugin script (plugins/once/run_fakemeter.sh)... 
2019/02/27 08:28:31 connected to CC111x radio on /dev/spidev5.1
2019/02/27 08:28:31 setting frequency to 916.650
2019/02/27 08:28:31 model 522 pump
2019/02/27 08:28:34 disconnecting CC111x radio on /dev/spidev5.1
Completed plugin script (plugins/once/run_fakemeter.sh). 
Running plugin script (plugins/every/plugins.sh)... 
Plugins Starting - Wed Feb 27 08:28:34 EST 2019
Completed plugin script (plugins/every/plugins.sh). 
Running plugin script (plugins/every/yo.sh)... 
/root/myopenaps
Completed plugin script (plugins/every/yo.sh). 

Starting oref0-pump-loop at Wed Feb 27 08:29:03 EST 2019 with 1 second wait_for_silence:
```

bin/oref0-pump-loop.sh
```diff
@@ -92,9 +92,42 @@ main() {
...
```
jpcunningh: I would recommend adding `-executable` to the find commands to eliminate the need for the `if` block in run_plugins testing for execution permissions.  It makes run_plugins a little cleaner.
```
find $once/* -executable | while read file; do run_script "$file" -d ; done
find $every/* -executable | while read file; do run_script "$file" ; done
```

bin/oref0-pump-loop.sh
```diff
@@ -87,32 +87,27 @@ main() {
             touch /tmp/pump_loop_success
             echo Completed oref0-pump-loop at $(date)
             update_display
-            run_plugins
             echo
         else
             # pump-loop errored out for some reason
             fail "$@"
         fi
+        run_plugins
```
scottleibrand: Moving it here won't help make it run on failures, as the fail function calls exit 1.  You'll instead need to add run_plugins after update_display there as well as above.

efidoman: Thank you. I was not aware. I have updated run_plugins to run in both places so it runs either way.

scottleibrand: I just tweaked the placement to be more consistent between the two cases.  LMK if you don't think that will do the right thing.

efidoman: LGTM

efidoman:
It's ready. I've been watching the logs and testing - works great.

PR #1211, opened 2019-02-25
Set cgm_loop_path for plugged-in G5 receiver

cluckj:
Fix for https://github.com/openaps/oref0/issues/1149

PR #1217, opened 2019-03-02
Add support for touchscreen receivers, fix CGM loop bug for connected receivers

jaylagorio:
Changed files:
`oref0-setup.sh`
- Includes cgm_loop_path fix found by @chuckj for both G5 and G6 (https://github.com/openaps/oref0/pull/1211/commits/8799a13acfc11b7792b719f608258207df9691ba to fix https://github.com/openaps/oref0/issues/1149) and includes support for G5/G6 touch receivers (previously under https://github.com/openaps/oref0/pull/1182)

`oref0-cron-every-minute.sh`
- Reincluding support for offline CGM, with optional upload, for G5 and G6 receivers


jaylagorio:
It looks like a test went bad, but it's not clear to me why considering this PR doesn't have anything to do with basal/bolus/suspension or anything dosing related.

```
ok 191 IOB should calculate IOB without counting time pump suspended at end of basal
not ok 192 IOB should calculate IOB without counting time pump suspended in middle of basal
  AssertionError: expected -0.01 to be 0.235
      at Assertion.fail (node_modules/should/lib/assertion.js:180:17)
      at Assertion.prop.value (node_modules/should/lib/assertion.js:65:17)
      at Context.<anonymous> (tests/iob.test.js:973:38)
ok 193 IOB should calculate IOB without counting time pump suspended surrounding a basal
```

tzachi-dar:
Why did you remove lines like:
 adddate openaps.pump-loop | uncolor |tee -a /var/log/openaps/openaps-date.log ?

This are needed in order to allow people to read the logs using apache-chainsaw.
See https://openaps.readthedocs.io/en/latest/docs/While%20You%20Wait%20For%20Gear/monitoring-OpenAPS.html?highlight=chainsaw#apache-chainsaw

jaylagorio:
I didn't remove those lines. My suspician is there was a change between the time I pulled the file, those lines were added, and when and made my changes and committed because I wouldn't have removed those lines in those places. I'll add them back and commit.

tzachi-dar:
Thanks

בתאריך שבת, 9 במרץ 2019, 06:14, מאת jaylagorio ‏<notifications@github.com>:

jaylagorio:
Fixed the above logging lines for apache-chainsaw. As for the conflict listed above, I think I can't hit the resolve button without commit rights but as of right now the conflict diff between `dev` and this PR is that this PR adds three lines. The correct resolution is to add those three lines (the `dev` branch side of the diff is empty).

bin/oref0-cron-every-minute.sh
```diff
@@ -62,14 +62,10 @@ find /var/log/openaps/pump-loop.log -mmin +5 | grep pump && (
...
 
-if [[ ${CGM,,} =~ "g5-upload" ]]; then
-    oref0-upload-entries &
-fi
-
 if [[ ${CGM,,} =~ "g4-go" ]]; then
         cd $CGM_LOOPDIR
         if ! is_bash_process_running_named oref0-g4-loop; then
-            oref0-g4-loop | tee -a /var/log/openaps/cgm-loop.log | adddate openaps.cgm-loop | uncolor |tee -a /var/log/openaps/openaps-date.log &
+            oref0-g4-loop | tee -a /var/log/openaps/cgm-loop.log &
```
scottleibrand: Does this one need to be re-added too?

jaylagorio: That's right, sorry. Fixed.

PR #1218, opened 2019-03-02
Update oref0-bluetooth for pi

jpcunningh:
This adds some log messages and provides support for various locations of the bluetoothd executable.

My Pi location from Pi Bakery image was /usr/sbin/bluetooth. I've upgraded by building from source, and the location is now /usr/libexec/bluetooth.

### DIFF
--- a/bin/oref0-bluetoothup.sh
+++ b/bin/oref0-bluetoothup.sh
@@ -7,11 +7,12 @@ Usage: $self
 Attempt to establish a Bluetooth tethering connection.
 EOT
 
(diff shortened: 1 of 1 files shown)


scottleibrand:
LGTM.  Do we need any regression testing on this before merging?

jpcunningh:
I've successfully tested it on:

- Pi with the distro version of bluez
- Pi with the latest version of bluez compiled from source
- Edison with the distro version of bluez

Any others that need to be tested?

jpcunningh:
@scottleibrand, I believe this one is ready.

PR #1225, opened 2019-03-10
Remove manual x12 setup

cluckj:
Thanks to @ecc1 we can now pull looping data (sensitivities, basals, targets, and settings) from x12 model pumps, and remove the workarounds & manual configurations.


cluckj:
Tested it locally a couple times & walked someone through it remotely. Maybe one more tester?

tzachi-dar:
As a 712 user, thanks...
What will happen to existing setups? will it silently ignore the locale settings and take the information from the pump?

cluckj:
Yeah, it'll overwrite the manually-configured files in favor of what is configured on the pump. If you decide to change a setting, you only have to change it on the pump :-)

cluckj:
(the commit was updating my x12-fix repo to latest `dev`)
Should be good to go. Tested with another x12 user this evening!

PR #1229, opened 2019-03-12
Fix check for WW pumps

cluckj:
Bug introduced by #1176: oref0-mmtune was always starting Go-mmtune with the -ww flag; this should make the check do what it's supposed to.

### DIFF
--- a/bin/oref0-mmtune.sh
+++ b/bin/oref0-mmtune.sh
@@ -28,7 +28,7 @@ fi
 
 function mmtune_Go() {
   set -o pipefail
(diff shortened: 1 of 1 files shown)


cluckj:
oref0-mmtune was always starting `Go-mmtune` with the `-ww` flag; this should make the check do what it's supposed to.

PieterGit:
Checked rest of my PR, and I think this one the only one.
It was introduced with commit https://github.com/PieterGit/oref0/commit/1d73dbb93d6786b620488532991c3ff0d5688e50 I pulled from @cluckj

PR #1231, opened 2019-03-18
Sort treatments before autotune

jpcunningh:
I noticed autotune was not working properly and tracked it down to a mixed set of Zulu vs. local created_at timestamps in the treatments collection.

The Mongo sort in NS does not account for the different zones, so sort them locally before sending them to autotune-prep.

### DIFF
--- a/bin/oref0-autotune-prep.js
+++ b/bin/oref0-autotune-prep.js
@@ -21,6 +21,8 @@
 */
 
 var generate = require('../lib/autotune-prep');
+var _ = require('lodash');
+var moment = require('moment');
 
 if (!module.parent) {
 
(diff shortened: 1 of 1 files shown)


scottleibrand:
looks like this may be a fairly major bug in autotune, where a mixed set of Zulu vs. local created_at timestamps in the treatments collection would cause insulin dosing to be misattributed, thereby causing autotune to adjust basals at the wrong time of day

I just re-ran autotune using the fixed branch (https://github.com/openaps/oref0/tree/jpcunningh-sort-treatments-before-autotune), and found that at certain times of day it adjusted basals all the way from autosens_min to autosens_max, i.e. from 0.8 u/h to 1.5 u/h.

Dana Lewis @danamlewis 20:10
(results w/ the updated code look more realistic to me)

Scott Leibrand @scottleibrand 20:11
^^ yeah, that. I'm gonna run this on our rigs to confirm it gives better BG control, but it'd probably be good to get a few more testers to see if you have similarly large changes in results, and ideally confirm whether using the fixed branch's results gives better glycemic control

(from https://gitter.im/openaps/autotune?at=5cb547380aad6350198eff10)

scottleibrand:
This is clearly affecting others as well. Based on feedback in Gitter (starting at https://gitter.im/openaps/autotune?at=5cb622be990feb4518b1f999) I think this is ready to merge.

PR #1232, opened 2019-03-19
Add a script to delete device status entries that are older than 2 months.

tzachi-dar:
openaps uploads a lot of date to nightscout.
This script makes it easy to delete old entries.
To use it simply run:

 bin/ns-delete-old-devicestatus.sh delete <NIGHTSCOUT_HOST>  <asn1 of API_SECREAT>

### DIFF
--- a/bin/ns-delete-old-devicestatus.sh
+++ b/bin/ns-delete-old-devicestatus.sh
@@ -18,7 +18,7 @@ function get_tid ( ) {
 }
 
 function write_backup() {
-tee >(json -a >> $BACKUP_DIR/devicestatus.txt) 
+tee >(json -a -o jsony-0 >> $BACKUP_DIR/devicestatus.txt) 
 }
 
 function debug_cmd ( ) {
--- a/bin/ns-delete-old-devicestatus.sh
+++ b/bin/ns-delete-old-devicestatus.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+
+source $(dirname $0)/oref0-bash-common-functions.sh || (echo "ERROR: Failed to run oref0-bash-common-functions.sh. Is oref0 correctly installed?"; exit 1)
+
+usage "$@" <<EOF
+Usage: $self --find <NIGHTSCOUT_HOST> <asn1 of API_SECREAT>- No-op version, find out what delete would do.
+$self delete <NIGHTSCOUT_HOST>  <asn1 of API_SECREAT>- Delete duplicate entries from ${NIGHTSCOUT_HOST-<NIGHTSCOUT_HOST>}
+EOF
+
+function fetch ( ) {
+  two_month_ago=$(date -d "-2 month" +%Y-%m-%d)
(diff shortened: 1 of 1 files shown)


bin/ns-delete-old-devicestatus.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+
+source $(dirname $0)/oref0-bash-common-functions.sh || (echo "ERROR: Failed to run oref0-bash-common-functions.sh. Is oref0 correctly installed?"; exit 1)
+
+usage "$@" <<EOF
+Usage: $self --find <NIGHTSCOUT_HOST> <asn1 of API_SECREAT>- No-op version, find out what delete would do.
+$self delete <NIGHTSCOUT_HOST>  <asn1 of API_SECREAT>- Delete duplicate entries from ${NIGHTSCOUT_HOST-<NIGHTSCOUT_HOST>}
```
scottleibrand: API_SECREAT should be API_SECRET.  What exactly is `asn1`?

tzachi-dar: Obviously, I meant to say sha1.

The reason that I wanted to give it as the sha1 and not the api_secreat itself, is because as far as I have noticed, openaps only stores the sha1 of the api secret.

My long term result is to call this automatically, and thus stop the rigs from failing after a few months of good work.

tzachi-dar:
I'll add en extra parameter to pass the number of days.

Currently don't have time to work on this, but hopefully, in a few days I'll be on vacation so I can do the changes that I was requested.

tim2000s:
If you’re going to make it run automatically in the long run, can that wait until there is an automated OpenHumans donate option, so that valuablensata isn’t lost?

tzachi-dar:
When is such an automatic option expected?

The main issue that I see is that there are complains on openaps stops working that are related to device status collection data size.
What people end up doing is deleting this collection at all. So this PR will not change anything as data is being deleted.
That said, I can definitely add an option to save the data locally, and then once there is a tool to upload the data to open humans, it can also read this location.

tzachi-dar:
Did the following changes:

1) Create a local backup of the data.
2) Add number of days to delete as a parameter.
3) Add a nightly mode (parameters are given as environment variables).

Please review again.

dramageek:
If the goal is to do this automatically I get it, but this function can be done directly from Nightscout too. 
Nightscout Page>Menu>Admin Tools>Delete Old Documents. Default is older than 30 days, but you can enter a different length.

tzachi-dar:
> If the goal is to do this automatically I get it, but this function can be done directly from Nightscout too.
> Nightscout Page>Menu>Admin Tools>Delete Old Documents. Default is older than 30 days, but you can enter a different length.

Obviously the goal is to make this process automatic. It is not a good practice to ask everyone to do a manual step, otherwise the ring will stop functioning at the middle of the night.

And, the option you mentioned deletes the information. This pr, is keeping it locally on the ring.

scottleibrand (approved):

PR #1238, opened 2019-03-30
Categorize improvements for libre 1 minute data

rob-andy:
* Determine date & dateString properties in prepGlucose map function
  * Filter invalid records in filterRecords filter function
  * Clean up for loop to remove previously performed checks
  * Average any values that fall within the 2 minute deadband

### DIFF
--- a/bin/oref0-autotune.py
+++ b/bin/oref0-autotune.py
@@ -156,7 +156,7 @@ def get_nightscout_bg_entries(nightscout_host, start_date, end_date, directory):
     date_list = [start_date + datetime.timedelta(days=x) for x in range(0, (end_date - start_date).days)]
 
     for date in date_list:
-        url="{0}/api/v1/entries/sgv.json?find\[date\]\[\$gte\]={1}&find\[date\]\[\$lte\]={1}`&count=1000"
+        url="{0}/api/v1/entries/sgv.json?find\[date\]\[\$gte\]={1}&find\[date\]\[\$lte\]={1}`&count=1500"
         url = url.format(nightscout_host, date)
         #TODO: Add ability to use API secret for Nightscout.
         res = requests.get(url)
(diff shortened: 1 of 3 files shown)


scottleibrand:
Is this still a WIP, or ready for review?  If ready, can you explain what you're doing with the various changes?

tzachi-dar:
A few small notes:
1) xdrip already has an option to send the average of 5 minutes libre data. See https://github.com/NightscoutFoundation/xDrip/pull/828

2) when checking for legal bg, one should also check that it is smaller than 500 (maybe also from 400).

3) The function  calculate_expected_delta() assumes that data is received every 5 minutes.  (see https://github.com/openaps/oref0/blob/c68c55d78de6fcdd87f39132c2ba091375cbab02/lib/determine-basal/determine-basal.js#L30)
The function     getLastGlucose also assumes a 5 minute between reading (see)            avgdelta = change/minutesago * 5; (https://github.com/openaps/oref0/blob/c68c55d78de6fcdd87f39132c2ba091375cbab02/lib/glucose-get-last.js#L33)
So I wander how supplying per minute data should work.
(Maybe the two errors well be cancelled).

rob-andy:
@tzachi-dar we aren't working on per minute data.  Because everything is translated to be in mg/dl/5 minutes.

Readings between -2 and 2.5 minutes are averaged, to create an average mg/dl/5minute delta for the latest readings.

rob-andy:
My changes in this pull request were solely related to making autotune produce a meaningful result, not a full audit of all places that use svg values in oref0.
The pull request comment explains the changes made;
1. stop calculating date and datestring in different places.  just do it once
2. perform all data sanitisation at the beginning & in one spot.  ie. removing below 39 values.  and any future cleanups people wish to add
3. put all data into the desired bucket size correctly.  the previous algorithm assumed at most 2 records would fall in the same bucket.  with 1 minute data, this isn't the case.  so average all values that fall into the bucket.

scottleibrand:
Just pushed https://github.com/openaps/oref0/tree/autotune-1m for easier testing.

scottleibrand:
I just ran autotune using both the `dev` and `autotune-1m` branches, and got identical results against our (5m) NS data, so looks like there are unlikely to be any major regressions there.

Are there any other tests we should run to validate this for 1m data or anything else before merging it to dev?

PR #1239, opened 2019-04-10
use existing var

tynbendad:
noticed on inspection, untested.
i'm also wondering why my temp target of >1 day only lasted about 18 hours - there's likely another issue in here (given the ns fetch limits to 24 hours, and i'm in mountain timezone, i'd guess its a timezone issue, but i'm not familiar with these date functions to see it...)

### DIFF
--- a/lib/profile/targets.js
+++ b/lib/profile/targets.js
@@ -42,12 +42,12 @@ function lookup (inputs, profile) {
         //console.error(start);
         var expires = new Date(start.getTime() + temptargets_data[i].duration * 60 * 1000);
         //console.error(expires);
-        if (now >= new Date(temptargets_data[i].created_at) && temptargets_data[i].duration === 0) {
+        if (now >= start && temptargets_data[i].duration === 0) {
             // cancel temp targets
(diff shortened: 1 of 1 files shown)


scottleibrand:
The 18h thing is expected, given how far back we fetch temptarget records. I don’t recommend trying to set temp targets that long. But you can schedule future ones in NS if you really need them that long.

tynbendad:
shouldn't it have been 24 hours, per this code:
https://github.com/openaps/oref0/blob/fa395b4b5fa02539ccd892a3d8fea62d4e3564e0/bin/nightscout.sh
    temp_targets)
    expr=${1--24hours}

or am i looking in the wrong place for the fetch?

scottleibrand:
I believe there's a UTC conversion issue that causes it to pull 24h +/- your UTC offset, so on the west coast we get temp target records from about 16-17h back.

tynbendad:
ok, so the fetch code is off by the timezone, not this code, that makes sense. the change here is unrelated btw, just a tiny optimization.

PR #1241, opened 2019-04-13
0.6.3 patch release

cluckj:
- Fixes for #1236 to continue support on Jubilinux 0.2.0 (Debian Jessie):
 -Modifies /etc/sources.list to use archive.debian.org
 -Disables "Valid Until" check (certificates are expired on archive.debian.org)
- Install node/npm v8 (using Nodesource) on the Edison
- Install v1.7.0 of MRAA
- Only install bluez from source if on Jubilinux 0.2.0 (Debian Jessie), Jubilinux 0.3.0 uses a newer version (5.43).

PR #1242, opened 2019-04-15
Apply #1241 fixes to dev too

scottleibrand:
Just tried to update a dev rig running jessie, and had to merge cluckj:master into dev to get it to update.  @cluckj any reason not to merge this to dev first before we merge it to master and release?


cluckj:
Yeah, as-is will break stuff; I think that `dev` needs the repo updates (it already installs node v8, and doesn't use MRAA).

PR #1246, opened 2019-04-21
Turn off screen only after 5 minutes.

tzachi-dar:
Some times, (many times starting on first install) one can not connect the pi to WiFi, but can connect a screen.
This pr will give them 5 minutes before the screen is turned off, to allow them to either enable connection to WiFi, or disable screen turning off.

One more thing that is changes is that hat might be on spidev0.[01] so screen is shutdown in both cases. (without this change, on adafruit bonnet screen is not turned off at all).


scottleibrand:
LGTM. Any testing needed before merge?

tzachi-dar:
I have tested this files on my system.
If we want to be 100% sure, I can test the pr itself tomorrow to make sure it works.

tzachi-dar:
Thanks for the fast merge.
Just did a fresh install, and it worked.

PR #1247, opened 2019-04-22
Fixes for openaps-packages.sh in dev

cluckj:
- Remove calls to oref0-bash-common-functions.sh (it is not present on fresh installs)
- Add `die()` to the script so critical errors will stop the install
- Use nodesource (installs nodejs+npm) on all platforms except armv6 (pi0)
- Use `dev` openaps-packages.sh instead of $BRANCH (remove this before merging `dev` to `master`)
- Update npm to latest version on armv6 (pi0) using npm's self-updater (integrates https://github.com/openaps/docs/pull/1444)

### DIFF
--- a/bin/openaps-install.sh
+++ b/bin/openaps-install.sh
@@ -58,7 +58,9 @@ fi
 sed -i "s/daily/hourly/g" /etc/logrotate.conf
 sed -i "s/#compress/compress/g" /etc/logrotate.conf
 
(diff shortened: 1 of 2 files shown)


cluckj:
I think this should also fix the spi error issues we're seeing on `dev`, since `master`'s openaps-packages was doing `sudo npm install -g oref0`.

PR #1252, opened 2019-05-03
Remove binary download options

cluckj:
- Removes pre-compiled binary download options and builds pump communication binaries from source
- Minor changes to setup options to provide more clarity around RFM69HCW radio modules

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -28,10 +28,9 @@ DIR=""
 directory=""
 EXTRAS=""
 radio_locale="US"
-buildgofromsource=false
+radiotags="cc111x"
 ecc1medtronicversion="latest"
 ecc1dexcomversion="latest"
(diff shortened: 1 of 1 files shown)


scottleibrand:
do we need any pre-merge testing?  or is this confirmed to be better than existing dev given the lack of binaries?

cluckj:
Extra testing is always good (I ran through an install on the HAT), but until then we can direct folks to select "build from source" in the existing setup.

scottleibrand:
Does this cleanly handle runagains?

cluckj:
Yep, the defaults are the same and buildgofromsource wasn't passed to runagain. I'm going to fix an issue with the radiofruit, and add radiotags to runagain.

scottleibrand:
Ok, ready to merge to dev for further testing there?

cluckj:
Yeah.

PR #1253, opened 2019-05-06
Use Edison's FAT32 Partition

cluckj:
Mount the Edison's FAT32 partition to give us enough room to install golang.


bin/oref0-setup.sh
```diff
@@ -647,7 +647,16 @@ if prompt_yn "" N; then
         # Replace apt sources.list with archive.debian.org locations
         echo -e "deb http://security.debian.org/ jessie/updates main\n#deb-src http://security.debian.org/ jessie/updates main\n\ndeb http://archive.debian.org/debian/ jessie-backports main\n#deb-src http://archive.debian.org/debian/ jessie-backports main\n\ndeb http://archive.debian.org/debian/ jessie main contrib non-free\n#deb-src http://archive.debian.org/debian/ jessie main contrib non-free" > /etc/apt/sources.list
     fi
-
+    
+    #Mount the Edison's fat32 partition at /usr/local/go to give us lots of room to install golang
+    if is_edison && [ -e /dev/mmcblk0p9 ] && ! mount | grep -qa mmcblk0p9 ; then
+        echo 'Removing golang from /usr partition...' && rm -rf /usr/local/go && mkdir -p /usr/local/go
```
scottleibrand: Does the rm -rf return a success code and allow the mkdir -p to complete even if the directory doesn’t get exist? Or should the mkdir -p be a new command on its own line?

cluckj: Yes, the former.

cluckj:
Been tested a few times by a couple folks, probably good to go.

PR #1256, opened 2019-05-16
Remove old golang install before reinstalling.

alimhassam:
Remove old golang install before re-installing. 

Installing a new version of golang on top of an old version gives difficult to diagnose compile errors. 

Also changed the golang version to 1.12.5. I compiled the pump communication library with 1.12.5 while troubleshooting my issue, and everything worked well so might as well update it here i think.

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -1273,14 +1273,17 @@ if prompt_yn "" N; then
     # Install Golang
     mkdir -p $HOME/go
     source $HOME/.bash_profile
-    if go version | grep go1.11.; then
+    golangversion=1.11
+    if go version | grep go${golangversion}.; then
         echo Go already installed
     else
+        echo "Removing possible old go install..."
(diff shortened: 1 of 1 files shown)


bin/oref0-setup.sh
```diff
@@ -1273,14 +1273,17 @@ if prompt_yn "" N; then
     # Install Golang
     mkdir -p $HOME/go
     source $HOME/.bash_profile
-    if go version | grep go1.11.; then
+    golangversion=1.12.5
+    if go version | grep go${golangversion}.; then
         echo Go already installed
     else
+        echo "Removing possible old go install..."
+        rm -rf /usr/local/go
```
cluckj: `rm -rf /usr/local/go/*` might work better here, as the directory itself can't be removed on Edisons with the FAT partition mounted at `/usr/local/go` (it produces a non-fatal error).

alimhassam: fixed

Issue #1257, opened 2019-05-18
Adverse Event: incorrect insulin delivery due to falsely high and flat Libre/LimiTTer/xDrip+ readings


scottleibrand:
We received an Adverse Event report today (via private message to @AdrianLxM) of a user running OpenAPS oref0 version 0.6.2, with CGM data from Libre/LimiTTer/xDrip+, who received excess insulin due to failure to detect a failing sensor.  The event in question did not result in hypoglycemia or any other negative health outcome, but could have done so if they had not noticed the problem and removed the failing sensor.

As shown at https://slack-files.com/TAWP5KA0L-FJV9J8U5U-478bd8cf56, the failing Libre sensor gradually drifted low over the course of the morning.  At approximately 10am local time, they recalibrated it to approximately 16 mmol/L.  However, because the Libre raw readings were so low and barely changing, the resulted calibrated readings remained high and nearly flat, independent of actual BG.  The user, a teenager, uses oref0's enabledSMB_always feature, to keep BG in range despite frequently forgetting to bolus for meals.  As a result, oref0 delivered an incorrect quantity of insulin, as the CGM data indicated BG was still high and not coming down much.  Once the user noticed the incorrect CGM readings, they removed the sensor, thereby stopping automated insulin delivery around 7pm, and resumed manual diabetes care until a new sensor session was started early the next morning. The reporter did not recall any particular interventions that were required due to the insulin delivered by OpenAPS.

After Adrian put me in contact with the reporter, I requested access to the user's Nightscout instance, which was provided, and downloaded the CGM entries for the time period in question.  I also requested access to the pump-loop.log output for that time, which was provided via papertrail.  I performed a full characterization of the issue, contacted the developer of xDrip+, and began evaluating a workaround for oref0 that would have mitigated the issue, on which I'll open a PR shortly.

A relevant subset of the information provided by the reporter is available at https://gist.github.com/scottleibrand/4c8c84d9989afc98bc92214a08e849f0 and the full logs can be made available upon request.

efbest:
It might be even more complicated: these data I pulled from a patient‘s Libre Monitor into a commercial software (‚Diabass 5 pro‘). The graphs show oscillating cgm values 10 minutes apart over a couple of dass. https://share.icloud.com/photos/0Z0JAXZMjgampKmjuDMqJflAQ

scottleibrand:
[...]

PR #1258, opened 2019-05-18
don't loop on too-flat CGM data

scottleibrand:
As detailed at https://github.com/openaps/oref0/issues/1257, a failing Libre sensor, uploaded to Nightscout via LimiTTer/xDrip+, can result in falsely high and too-flat CGM readings and incorrect insulin delivery.  The existing mitigation for this, to avoid looping on completely unchanging CGM data, was insufficient in this case.  To further mitigate this and similar issues, this PR changes that check to look for CGM data that is changing less than 1 mg/dL/5m for 45m.  As shown at https://gist.github.com/scottleibrand/4c8c84d9989afc98bc92214a08e849f0, that would have been sufficient to prevent further insulin dosing in this case after CGM data falsely flattened out at 290 mg/dL.

This will likely result in occasional false-positives when BG is truly flat for an extended period of time, or transiently when both the short_avgdelta and long_avgdelta are between -1 and 1 mg/dL/5m.  However, such false-positives should be brief, and will simply result in OpenAPS refraining from SMBs or high-temps until more-plausible CGM data arrives.

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -86,7 +86,7 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
             rT.reason = "Error: CGM data is unchanged for the past ~45m";
         }
     }
-    if (bg <= 10 || bg === 38 || noise >= 3 || minAgo > 12 || minAgo < -5 || ( glucose_status.short_avgdelta > -1 && glucose_status.short_avgdelta < 1 && glucose_status.long_avgdelta > -1 && glucose_status.long_avgdelta < 1 ) {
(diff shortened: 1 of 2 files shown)


scottleibrand:
Fixed tests.  Have been running without issue on our rigs for over a week now.  Ready to merge if no objections.

scottleibrand:
Also reviewed all the logs from the rigs this was running on, and confirmed that the largest microbolus delivered after a period of too-flat CGM data was 0.1U.  Also updated the too-flat logic to only kick in when BG > 60, to avoid the possibility of a zero-temp expiring prematurely when BG is low and flat.

scottleibrand:
Since 5962202 this is now looking good on our rigs.  Ready to merge if no objections.

PR #1259, opened 2019-05-18
don't loop on too-flat CGM data (to master)

scottleibrand:
As detailed at https://github.com/openaps/oref0/issues/1257, a failing Libre sensor, uploaded to Nightscout via LimiTTer/xDrip+, can result in falsely high and too-flat CGM readings and incorrect insulin delivery.  The existing mitigation for this, to avoid looping on completely unchanging CGM data, was insufficient in this case.  To further mitigate this and similar issues, this PR changes that check to look for CGM data that is changing less than 1 mg/dL/5m for 45m.  As shown at https://gist.github.com/scottleibrand/4c8c84d9989afc98bc92214a08e849f0, that would have been sufficient to prevent further insulin dosing in this case after CGM data falsely flattened out at 290 mg/dL.

This will likely result in occasional false-positives when BG is truly flat for an extended period of time, or transiently when both the short_avgdelta and long_avgdelta are between -1 and 1 mg/dL/5m.  However, such false-positives should be brief, and will simply result in OpenAPS refraining from SMBs or high-temps until more-plausible CGM data arrives.

See also https://github.com/openaps/oref0/pull/1258 for the same change to `dev`

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -74,11 +74,11 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     if (minAgo > 12 || minAgo < -5) { // Dexcom data is too old, or way in the future
         rT.reason = "If current system time "+systemTime+" is correct, then BG data is too old. The last BG data was read "+minAgo+"m ago at "+bgTime;
     }
-    // if BG is too old/noisy, or is completely unchanging, cancel any high temps and shorten any long zero temps
-    if ( glucose_status.short_avgdelta == 0 && glucose_status.long_avgdelta == 0 ) {
+    // if BG is too old/noisy, or is changing less than 1 mg/dL/5m for 45m, cancel any high temps and shorten any long zero temps
+    if ( glucose_status.short_avgdelta > -1 && glucose_status.short_avgdelta < 1 && glucose_status.long_avgdelta > -1 && glucose_status.long_avgdelta < 1 ) {
         rT.reason = "Error: CGM data is unchanged for the past ~45m";
     }
-    if (bg <= 10 || bg == 38 || noise >= 3 || minAgo > 12 || minAgo < -5 || ( glucose_status.short_avgdelta == 0 && glucose_status.long_avgdelta == 0 ) ) {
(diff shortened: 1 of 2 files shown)


Bender1061:
would this effect the MDT giving a 400 when the pump is high? I rely on this to keep treating my loop if I'm having a bad diabetes day, and if this gets broken that is really going to be a problem.

scottleibrand:
This shouldn't change the behavior when the CGM is reporting the exact same value repeatedly for >45m, although it might make that behavior kick in after 40m or 35m if the 40m-ago value was > 393 or so.  If that behavior is already suboptimal, we should open a new PR/issue to discuss changing it.  I don't know for sure whether there are any other special-case exceptions in the code for LOW and HIGH values that would kick in if the CGM remains at that value for 45m, but I suspect it hasn't been a problem on the HIGH side, mostly because if you get all the way to 400, you'll have quite a lot of IOB, which (absent a site failure) will be enough to bring BG back down into the 300s where the loop can resume normal dosing.  On the LOW side, we should verify that the system won't resume normal basal after 45m+30m if BG is still LOW.

Bender1061:
On a bad diabetes day I will be at 400 for hours. It will bring my blood
sugar down eventually. Ideal case
Yes I can be more proactive and give myself a large bolus to being me down
faster but that generally is like a reverse rebound. And then there are the
days where I have no internet where I might not even notice I'm high
because I'm working in a high stress environment.

I see why you want to do this, maybe we can add the special case of MDT and
400 not to be effected by this.

scottleibrand:
We're not adding this functionality.  We're slightly relaxing the conditions for activating it.  If the situation you describe (>400 for hours) would interact poorly with this check, then it is already doing so.

If you're concerned about this, you can check your logs and see if you've ever seen the "Error: CGM data is unchanged for the past ~45m" message when short_avgdelta=0 and long_avgdelta=0.

sulkaharo:
Wonder how this affects looping for low carb loopers who can genuinely get a super flat curve. Would it make sense to treat flat curves differently if the flat happens on high or low BGs?

scottleibrand:
The same behavior (canceling high temps and long zero temps) would have different (and approximately correct) results in each case.

Bender1061:
I guess you are right, the rig I've been running and carrying on me since last summer, has that code, strange that I've not seen it cut out like that. but either way, I guess this is not really a problem for me. 

Yes it's likely time to update my rig, but I can't believe I've gotten this rig running pretty smooth edison/explorer board for about 10 months with no issues! 

thanks

scottleibrand:
Closing this in favor of releasing dev.

PR #1260, opened 2019-05-18
Bgproxy integration.

alimhassam:
A new preference option is added enableEnliteBgproxy. When set to true
the bgproxy command will be call at the end of every loop sending the
glucose data to the medtronic pump.

This gives an option to setup low and high alarms directly from the
pumps and also see the glucose trend when offline.

To enable this feature:
 - set enableEnliteBgproxy: true in preferences.json
 - turn on the sensor feature in the medtronic pump.
 - set the sensor id to 0000000 in Sensor setup
 - select sensor->sensor start->Find lost sensor. Glucose should appear after a couple of
cycles.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -88,6 +88,7 @@ main() {
             echo Completed oref0-pump-loop at $(date)
             update_display
             run_plugins
+            update_bgproxy
             echo
         else
             # pump-loop errored out for some reason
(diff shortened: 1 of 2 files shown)


scottleibrand:
LGTM. Any more testing needed before we merge to dev?

scottleibrand:
It won’t do anything without enableEnliteBgproxy set, so merging.

PR #1266, opened 2019-05-31
install flask-cors - enable in www/app.py

efidoman:
installing flask-cors module and enabling in app.py allows a web-based script (such as IOS shortcut to running javascript) to access the rig's local web server for retrieving JSON information such as BG records, IOB, COB, Temp Basal/Duration, Battery Percentage, etc.

### DIFF
--- a/www/app.py
+++ b/www/app.py
@@ -46,35 +46,35 @@ def index():
 
 @app.route("/suggested")
 def suggested():
-    json_url = os.path.join("/root/myopenaps/enact/suggested.json")
+    json_url = os.path.join(myopenaps_dir + "enact/suggested.json")
     data = json.load(open(json_url))
     return jsonify(data)
 
 @app.route("/enacted")
 def enacted():
-    json_url = os.path.join("/root/myopenaps/enact/enacted.json")
+    json_url = os.path.join(myopenaps_dir + "enact/enacted.json")
     data = json.load(open(json_url))
     return jsonify(data)
 
 @app.route("/glucose")
 def glucose():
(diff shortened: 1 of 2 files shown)


efidoman:
I added two more json endpoints for the web service. 

cgm - the xdripjs cgm pill json record

hostname - the hostname of the rig. This is important for ensuring the calling application / script is accessing the correct rig and can display the rig hostname to the user. Since many use multiple rigs per person and multiple family members with rigs, it is important to know which rig the data is coming from.

efidoman:
Anyone know why Travis integration tests are failing? I can't tell by looking at the details what I did to break them.

www/app.py
```diff
@@ -74,6 +88,20 @@ def temptargets():
     data = json.load(open(json_url))
     return jsonify(data)
 
+@app.route("/cgm")
+def cgm():
+    json_url = os.path.join("/root/myopenaps/monitor/xdripjs/cgm-pill.json")
```
scottleibrand: This should use myopenaps_dir, not be hardcoded to the default path.

scottleibrand: Looks like this is incorrect throughout this file.  It appears that @zjedr removed the only uses of myopenaps_dir in 648369222d8939c9a3183045fd5439ff17075869 and replaced them with hard-coded paths.

So if you'd like to defer fixing that to a later PR, I'd also be OK with that.

efidoman: Good suggestions. I updated all hard-coded paths to use myopenaps_dir. Please take another look while I'm testing tonight.

efidoman:
Tested out good for me.

efidoman:
Yes, it's all working for me in my tests from my browsers on my laptop and from my phone.

PR #1268, opened 2019-06-11
Allow G4 Share serial to be blank

jimrandomh:
A small bug in `oref0-setup.sh` made it so that you couldn't proceed without providing a G4 Share serial number, breaking the use case where you don't have G4 Share. The problem is that `prompt_and_validate` takes an optional fourth argument for a default value, and if no default value is provided, then `prompt_and_validate` will re-prompt until you enter something that isn't blank.


scottleibrand:
Have you tested that this still allows you to enter a serial number, and uses it if you do?

scottleibrand:
@jimrandomh have you (or has anyone else) tested that this still allows you to enter a serial number, and uses it if you do?

PR #1274, opened 2019-06-24
Prevent basal rate increase on low delta

twain47:
Revolve a situation in which a non-smb zero temp can be reverted to full basal when delta is less than expected.

Situation can occur where only enableSMB_with_COB is enabled and carbs hit zero after a smb. It can result in extra basal insulin when a zero tempt is required.

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -972,19 +972,26 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
 
     // if eventual BG is above min but BG is falling faster than expected Delta
     if (minDelta < expectedDelta) {
-        // don't increase the basal rate in response to a low delta
-        if (basal < currenttemp.rate) {
+        // if in SMB mode, don't cancel SMB zero temp
+        if (! (microBolusAllowed && enableSMB)) {
             if (glucose_status.delta < minDelta) {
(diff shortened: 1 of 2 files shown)


lib/determine-basal/determine-basal.js
```diff
@@ -979,12 +979,19 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
...
                 rT.reason += "Eventual BG " + convert_bg(eventualBG, profile) + " > " + convert_bg(min_bg, profile) + " but Min. Delta " + minDelta.toFixed(2) + " < Exp. Delta " + convert_bg(expectedDelta, profile);
             }
-            if (currenttemp.duration > 15 && (round_basal(basal, profile) === round_basal(currenttemp.rate, profile))) {
-                rT.reason += ", temp " + currenttemp.rate + " ~ req " + basal + "U/hr. ";
-                return rT;
+
+            // don't increase the basal rate in response to a low delta
+            if (currenttemp.duration > 0 && basal > currenttemp.rate) {
+                rT.reason += "; extending current temp basal of " + currenttemp.rate + " U/hr. for 10 minutes. ";
+                return tempBasalFunctions.setTempBasal(currenttemp.rate, currenttemp.duration + 10, profile, rT, currenttemp);
```
scottleibrand: This won’t work as you’d expect: the pump requires all temp basals to be set at multiples of 30m.

scottleibrand:
Further discussion in Gitter starting at https://gitter.im/openaps/oref0?at=5d17e0198dff05627b93bd2c

PR #1282, opened 2019-07-31
raspbian password checking adjustment (also expire correct user)

rkresha:
old line 12: the existing logic didn't make sense to me check for edison  but expire root so adjusting to check for root, expire root
old line 16-21: If you changed the password for 'pi' before running the script you'd never get to the part about changing root's password as it's disabled and locked by default and therefore `/run/sshwarn` will not exist.
new line 16-32: Check if we are running Raspbian, check the build date against the current password date for both pi and root. If they are the same, we are using default passwords and prompt for change. Checking them independently and only display information message once.

Let me know if you have questions.

### DIFF
--- a/bin/openaps-install.sh
+++ b/bin/openaps-install.sh
@@ -9,16 +9,28 @@ sed -r -i"" "s/localhost( jubilinux)?$/localhost $myrighostname/" /etc/hosts
 sed -r -i"" "s/127.0.1.1.*$/127.0.1.1       $myrighostname/" /etc/hosts
 
 # if passwords are old, force them to be changed at next login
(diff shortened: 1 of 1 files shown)


rkresha:
The code block was mainly surrounding pi installs and I've added extra validation to verify it only gets run for the raspbian OS. Only thing touched for Edison was correcting the logic and worst-case scenario it prompts the user to reset the root password on next login. Unless checking the Edison user for password date and then forcing a password change for the root user makes sense to someone else.

I've tested on a new install on rpi and it works as expected
No real need to test it on Edison builds because of the reasons mentioned above.

Issue #79, opened 2016-03-06
Calculate stacked COB correctly in advanced-meal-assist


scottleibrand:
Don't blindly calculate from first or biggest carb treatment: properly figure out when COB drops to zero and start over after that.

scottleibrand:
Fixed with https://github.com/openaps/oref0/commit/78b79d173166944f4fb36fcb0de8bcd7ca3444b1

Issue #68, opened 2016-02-11
Advanced Meal Assist using COB estimated from post-meal deviations


scottleibrand:
I'm not yet happy with Meal Assist: I think we need to switch to using an advanced meal assist algorithm that is aware of COB / carb absorption and can tell when to high-temp aggressively for carbs being absorbed, and when insulin activity is already high enough to risk going low, and we need to preemptively low-temp instead.

To accomplish this, we need to determine how many meal carbs have been absorbed, at any given point after a meal, so we know how much more of a rise we can eventually expect from the remaining meal carbs, and whether we can expect any dips as insulin activity picks up.  This needs to be determined based on the actual observed effect on BG of carb absorption, not based on any assumptions about how fast carbs "should" absorb for any given meal.  By doing so, we should be able to safely high-temp for ~half the remaining unabsorbed carbs, as long as carbs are currently absorbing fast enough to outpace insulin activity.  If carb absorption is too slow for insulin activity and a low is projected to occur before all carbs absorb, we can instead low-temp until BG and/or deviation picks up sufficiently to make it safe to high-temp again.

To calculate carbs absorbed to date, we can sum up the 5m deviations, for each 5m period since the meal carbs were entered, to determine how much effect carb absorption has actually had on BG so far, and by using the IC ratio and ISF, determine how many carbs have been absorbed, and conversely how many g of COB are remaining.

We can then extrapolate that the current deviation (the current impact of carb absorption) will continue into the future until most of those carbs have absorbed.  To simplify the calculations, I'll probably extrapolate the current rate until those projected deviations have accounted for half the remaining COB.  (In reality, deviations tend to gradually return to zero as carb absorption slows down, but the simpler math should be sufficient for estimating the total area under that curve.)  We can then use the projected BGI and deviations for the next few hours to determine the projected minimum and eventual BG.  If projected minimum BG is below targeted min_bg (and below current BG), we can low-temp as needed to help bring that up.  If projected minimum BG is above target (or above current BG), we can high-temp as required to bring eventual BG down to target.

dm61:
[...]

PR #1284, opened 2019-08-07
Unused local variables

edwingustafson:


scottleibrand:
carbDelay was a legacy of the old Meal Assist algorithm, and was removed in #68. The maxCarbs one was removed in #79.

PR #1286, opened 2019-08-13
fix wrong interface choose

Kurbezz:
By default, choose the first interface found with a control socket in the socket path.


Kurbezz:
> Do you have any test results you can share?

No. I had problems with automatically connecting to another network on 2 boards. After use this fix problems. The problems disappeared after this fix.

PR #1287, opened 2019-08-25
Get clock from pump when offline

juehv:
Another snippet from my "towards long term offline looping"-project. I had the issue when the rig was down for more than a few minutes, the missaligned clockes prevented the loop from running. Therefore I take the clock from the pump when the rig boots and is offline. That does not effect online operation as this script does nothing when wifi is connected. Also when the rig comes online again, the rig clock gets ntp syncronized and updates the pump clock at some point.

Tested on my PI+HAT rig but should work for every pi based and edison based rig

### DIFF
--- a/bin/oref0-cron-post-reboot.sh
+++ b/bin/oref0-cron-post-reboot.sh
@@ -18,6 +18,27 @@ if [[ "${CGM,,}" =~ "xdrip" ]]; then
     python "$XDRIP_PATH/xDripAPS.py" &
 fi
 
+# Get time from pump for offline looping in case a rig was down for a while
+# At some point the rig will come online again and calibrates the local clock 
+# and the clock of the pump
+export MEDTRONIC_PUMP_ID=`get_pref_string .pump_serial | tr -cd 0-9`
+export MEDTRONIC_FREQUENCY=`cat monitor/medtronic_frequency.ini`
+
+sudo wpa_cli -i wlan0 scan
+sleep 60 # wait for wifi to connect
+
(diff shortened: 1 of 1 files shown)


scottleibrand:
Does this pose a race condition with oref0-pump-loop using the radio?  Your sleep 60 would allow the pump-loop to start, at which point the radio would be locked, right?

I know we originally put similar logic to this into oref0-pump-loop itself: not sure if that isn't working, or ?

https://github.com/openaps/oref0/blob/dev/bin/oref0-pump-loop.sh#L387

straykatz:
Hi - while backpacking for a week offline the last few days, I ran into the chip clock issue big time when rig lost power and when I had to swap batteries. I am looking forward to see what you are doing! And I can confirm that this is an issue at least with the chip I was using, which was up-to-date in dev last week (there have been two new commits since then.)

scottleibrand:
Actually, looking into this further, this shouldn't be able to do any harm, since setting the system clock just reads a file, and doesn't actually talk to the pump.  I'm also attempting to fix this with https://github.com/openaps/oref0/tree/offline-clock-set and will let you know if that works.

scottleibrand:
@juehv would you care to test out https://github.com/openaps/oref0/pull/1290 and see if that fixes offline looping for you without the new cron job?

juehv:
@scottleibrand Yes of course. I'll install it tomorrow evening.

scottleibrand:
Thanks.  LMK (preferably by commenting on #1290) when you've had a chance to confirm that it helps and doesn't break anything.

juehv:
#1290 works better

PR #1288, opened 2019-08-25
minZTUAMPredBG should only be used if UAM is enabled

twain47:
I think this is a bug.

I've observed minPredBG ending up as high as 460 (where the 999 value of minUAMPredBG ends up being used as real data).

The problem is mitigated by the `minPredBG = Math.min( minPredBG, avgPredBG );` line later on but may still cause problems.

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -754,18 +754,22 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     // if any carbs have been entered recently
     if (meal_data.carbs) {
 
-        // if UAM is disabled, use max of minIOBPredBG, minCOBPredBG
-        if ( ! enableUAM && minCOBPredBG < 999 ) {
-            minPredBG = round(Math.max(minIOBPredBG, minCOBPredBG));
-        // if we have COB, use minCOBPredBG, or blendedMinPredBG if it's higher
-        } else if ( minCOBPredBG < 999 ) {
-            // calculate blendedMinPredBG based on how many carbs remain as COB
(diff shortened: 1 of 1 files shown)


scottleibrand:
I'm having trouble figuring out what the situation is in which the current code would have problems, and this would fix it.  Perhaps it would be useful to write some unit tests replicating the behavior you observed (with the 460 for example) so we can see how they change with this PR?

twain47:
Struggling to find time to write a unit test.  In case it helps:

This occurs where enableUAM = false and there are no carbs on board (hence minCOBPredBG == 999).  In this case it ends up running the code in the final 'else' block
```
} else {
    minPredBG = minZTUAMPredBG;
}
```
but because enableUAM = false minZTUAMPredBG is still set to its default value of 999 and hence minPredBG ends up being set to 999

sleibrand:
Ok. If you have time to write a unit test, great. If not, I would be comfortable changing that last else to an else if minZTUAMPredBG < 999 and adding another else block. As it is I’m worried I don’t fully understand how you’re changing all the possible paths through the modified conditionals.

twain47:
Adding the extra if you suggest would resolve the issue with less code changes.

scottleibrand:
Can you review https://github.com/openaps/oref0/pull/1292 and see if that accomplishes what you wanted more simply?

twain47:
Yup, #1292 works.

Sorry - was trying to clean up the code but I can see without test coverage it is hard to validate.

scottleibrand:
Thanks.  Merged #1292, so I'll go ahead and close this out. Thanks for finding this and pointing the way to a fix.

PR #1290, opened 2019-09-18
run oref0-set-system-clock after oref0-set-device-clocks in case rig is offline

scottleibrand:
When a rig is offline, it is using this code path and running oref0-set-device-clocks (which fails), but never getting to the part of oref0-pump-loop that runs oref0-set-system-clock. In order to make sure that the rig sets its time from the pump when NTP is unavailable, this simply adds a oref0-set-system-clock run after the attempt to oref0-set-device-clocks.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -259,7 +259,9 @@ function smb_reservoir_before {
         # Check for bolus in progress and issue 3xESC to back out of pump bolus menu
         smb_verify_status \
         && try_return mdt -f internal button esc esc esc 2>&3 \
-        && oref0-set-device-clocks || oref0-set-system-clock
+        && oref0-set-device-clocks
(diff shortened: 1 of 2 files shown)


scottleibrand:
It looks like one of our unit tests may have a midnight-UTC boundary condition:
```
not ok 189 IOB should calculate IOB without counting time pump suspended in middle of basal
  AssertionError: expected -0.01 to be 0.235
      at Assertion.fail (node_modules/should/cjs/should.js:275:17)
      at Assertion.value (node_modules/should/cjs/should.js:356:19)
      at Context.<anonymous> (tests/iob.test.js:973:38)
```

scottleibrand:
Just did some whitespace cleanup to prompt another travis-ci run.

juehv:
I tested it several times during the week on my PI+HAT rig. Works very well! 👍 for merge :)

PR #1292, opened 2019-09-19
per #1288, only use minZTUAMPredBG if enableUAM

scottleibrand:
and use minGuardBG otherwise

This is intended to address the problem posed in #1288, and I'm hoping simplifies the solution proposed there and makes it easier to validate.


scottleibrand:
@twain47 gave this a 👍 at https://github.com/openaps/oref0/pull/1288#issuecomment-533107242

PR #1295, opened 2019-09-25
Improved rounding of mmol/L units in oref0-upload-profile

p5nbTgip0r:
When running `oref0-upload-profile` with an autotune report, the BG targets and ISF values in the report are rounded to the nearest whole number. This is not an issue if your autotune reports are generated in mg/dL units, however when using mmol/L a decimal is necessary to have precise ISF and BG target values. 

This PR solves these issues by changing the math for both ISF and BG targets to round to one decimal place if the profile units are in mmol/L. The rounding for mg/dL units is kept the same.

### DIFF
--- a/bin/oref0-upload-profile.js
+++ b/bin/oref0-upload-profile.js
@@ -178,11 +178,17 @@ if (!module.parent) {
 
         _.forEach(profiledata.isfProfile.sensitivities, function(isf_entry) {
 
-            var value = Math.round(isf_entry.sensitivity);
+            var value = isf_entry.sensitivity;
+            var conversionFactor = 1;
+            var decimals = new_profile.units === 'mmol' ? 10 : 1;
 
-            if (new_profile.units === 'mmol' && profiledata.isfProfile.units === 'mg/dL') {
-                value = +(Math.round(isf_entry.sensitivity / 18 + 'e+1') + 'e-1');
+            if (profiledata.isfProfile.units && new_profile.units !== profiledata.isfProfile.units) {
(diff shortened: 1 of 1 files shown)


bin/oref0-upload-profile.js
```diff
@@ -144,14 +144,22 @@ if (!module.parent) {
...
-                low_value = +(Math.round(target_entry.low / 18 + 'e+1') + 'e-1');
-                high_value = +(Math.round(target_entry.high / 18 + 'e+1') + 'e-1');
+            var low_value = target_entry.low;
+            var high_value = target_entry.high;
+            var conversionFactor = 1;
+            var decimals = new_profile.units === 'mmol' ? 10 : 1;
+
+            if (new_profile.units !== profiledata.bg_targets.units) {
+                // if the new profile is using a different unit than our BG targets, then set the conversion factor accordingly
+                conversionFactor = (new_profile.units == 'mmol' ? 0.055 : 18);
```
sleibrand: It’s not immediately obvious what you’re doing here. Is this to support both cases where the units are mismatched, and multiply or divide appropriately?

Either way, it’d probably best to add an explanatory comment here.

p5nbTgip0r: Yeah, this is to handle cases where the autotune profile is in different units from the Nightscout profile. The old code essentially did the same thing but only handled converting autotune values from mg/dL to mmol/L if the Nightscout profile was using mmol/L: `if (new_profile.units === 'mmol' && profiledata.bg_targets.units === 'mg/dL')` 

I'll add comments to explain this.

PR #1296, opened 2019-10-01
Fix values in mg/dl erroneously multiplied by 18 in profile upload to NS

WhatWouldHoneyBadgerDo:
Currently, a case-sensitive comparison of strings 'mg/dl' and 'mg/dL' causes the profile uploader to multiply the values measured in mg/dl by 18.  So an ISF of 50 mg/dl in OpenAPS becomes an ISF of 900 mg/dl in Nightscout.  And similarly, a target of 90 mg/dl in OpenAPS becomes a target of 1620 mg/dl in Nightscout.   This PR fixes that issue.

### DIFF
--- a/bin/oref0-upload-profile.js
+++ b/bin/oref0-upload-profile.js
@@ -150,7 +150,7 @@ if (!module.parent) {
             var decimals = new_profile.units === 'mmol' ? 10 : 1;
 
             // Check if the input profile units don't match the Nightscout profile units
-            if (new_profile.units !== profiledata.bg_targets.units) {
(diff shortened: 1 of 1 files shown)


WhatWouldHoneyBadgerDo:
@scottleibrand  Not sure if your question is directed to me or not, but I did test it and it works for me.

WhatWouldHoneyBadgerDo:
@scottleibrand I did think of one possible issue that I didn't account for.  If any of these values could be undefined, then they would throw an error when calling toUpperCase() on them.  I can add a check.

PR #1298, opened 2019-10-06
Add ability to calculate glucose noise

jpcunningh:
This adds the option for oref0 to calculate the glucose noise level prior based on the glucose history.

It is disabled by default and controlled with the calc_glucose_noise preference.

### DIFF
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -921,7 +921,7 @@ function compare_with_fullhistory() {
 }
 
 function update_glucose_noise() {
-    if check_pref_bool .calc_glucose_noise true; then
+    if check_pref_bool .calc_glucose_noise; then
       echo "Recalculating glucose noise measurement"
       oref0-calculate-glucose-noise monitor/glucose.json > monitor/glucose.json.new
       mv monitor/glucose.json.new monitor/glucose.json
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -921,7 +921,7 @@ function compare_with_fullhistory() {
 }
 
 function update_glucose_noise() {
-    if check_pref_bool .calc_glucose_noise; then
+    if check_pref_bool .calc_glucose_noise false; then
       echo "Recalculating glucose noise measurement"
       oref0-calculate-glucose-noise monitor/glucose.json > monitor/glucose.json.new
       mv monitor/glucose.json.new monitor/glucose.json
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -922,6 +922,7 @@ function compare_with_fullhistory() {
 
 function update_glucose_noise() {
     if check_pref_bool .calc_glucose_noise true; then
+      echo "Recalculating glucose noise measurement"
       oref0-calculate-glucose-noise monitor/glucose.json > monitor/glucose.json.new
       mv monitor/glucose.json.new monitor/glucose.json
     fi
--- a/bin/oref0-pump-loop.sh
+++ b/bin/oref0-pump-loop.sh
@@ -316,6 +316,9 @@ function smb_suggest {
 
 function determine_basal {
     cat monitor/meal.json
+
+    update_glucose_noise
+
     if ( grep -q 12 settings/model.json ); then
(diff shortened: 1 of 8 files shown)


bin/oref0-pump-loop.sh
```diff
@@ -917,6 +920,14 @@ function compare_with_fullhistory() {
   fi
 }
 
+function update_glucose_noise() {
+    if check_pref_bool .calc_glucose_noise false; then
```
scottleibrand: This seems backwards?

lib/glucose-stats.js
```diff
@@ -0,0 +1,244 @@
...
+const log = console.error;
+
+/* eslint-disable-next-line no-unused-vars */
+const error = console.error;
+const debug = console.error;
+
+module.exports = {};
+const calcStatsExports = module.exports;
+
+// Calculate the sum of the distance of all points (overallDistance)
```
scottleibrand: Should this be sod instead of overallDistance?

jpcunningh:
@efidoman originally implemented it in Logger and I ported it to Javascript for Lookout.  I believe Logger has a different algorithm in use now.

It's currently in use in Lookout. We have had good results for well over a year.  It's response for us has mapped very well to the ??? reported by the official Dexcom algorithm when a sensor gets noisy.

It overwrites any CGM provided noise value if calc_glucose_noise is true.

bin/oref0-pump-loop.sh
jpcunningh: It seems to work right.  It defaults to false (the second argument to check_pref_bool) if .calc_glucose_noise is not set, so the if expression only evaluates to true if .calc_glucose_noise is in the preferences.json file and is set to true.

lib/glucose-stats.js
jpcunningh: yes. Thanks!

bin/oref0-pump-loop.sh
scottleibrand: Ok, that's a really confusing implementation of check_pref_bool (not your fault). I read it to be equivalent to "if calc_glucose_noise === false".

scottleibrand:
Do you have any examples of a scenario where the CGM provides noise and we'd want to overwrite it?

jpcunningh:
For discussion on the original algorithm implemented in Logger: https://gitter.im/thebookins/xdrip-js?at=5a88098a35dd17022eb64c4e

jpcunningh:
I think ultimately we would want to override the CGM provided noise value if we calculate a medium or heavy noise, but the CGM is sending a clean noise value.

Or, in the case of the current state of the code, if one is using a CGM source they know does not correctly report noise, they can enable the option in oref0 to calculate the noise value and overwrite the CGM reported noise.

bin/oref0-pump-loop.sh
jpcunningh: Yes! that's the way I wrote it first, but it didn't work! 😄

scottleibrand:
Is it possible for the CGM to report medium or heavy noise, but for us to calculate light or zero noise and override the CGM? Seems like we'd want to take the max() of the two.

jpcunningh:
Added logic to do as you suggested - use the highest noise value between the calculated noise and the CGM reported noise.

scottleibrand:
Anyone want to volunteer to test this?

Tornado-Tim:
@scottleibrand I can test this.....

scottleibrand:
@Tornado-Tim any progress on testing?

scottleibrand:
Should we go ahead and merge this into dev for testing there?

jpcunningh:
We've been running it for a month without any noticeable issues... I believe it can be safely merge it to dev.

lib/glucose-stats.js
```diff
@@ -0,0 +1,244 @@
...
+  const lastDelta = 0;
+
+  for (let i = 1; i < n; i += 1) {
+    // y2y1Delta adds a multiplier that gives
+    // higher priority to the latest BG's
+    let y2y1Delta = (sgvArr[i].glucose - sgvArr[i - 1].glucose) * 1000.0 * (1 + i / (n * 3));
+
+    const x2x1Delta = xarr[i] - xarr[i - 1];
+
+    if ((lastDelta > 0) && (y2y1Delta < 0)) {
```
jotomo: @jpcunningh `lastDelta` is a never-updated const set to 0 in line 49. Given the name I'd expect this to be set at the end of the loop. Like it is now, this `if` and following `else if` never match, do they?

jpcunningh: Thanks, @jotomo!  You are right, the assignment line was missing.

PR #1299, opened 2019-10-06
Refactor enable smb in determine-basal.js

stevebell117:
So this is actually a step towards what I'm hoping to implement, but I think the `enable_smb` logic should be refactored to make it easier to read and add additional SMB toggles for more customize-ability in the future. 

Unfortunately my oAPS rigs are broken (radios are dead), so I'm not able to test this in oAPS but I have this running in AAPS and it's running with no issue. Would welcome some testing to make sure I didn't accidentally change some logic.


scottleibrand:
Merging to dev to test there.

PR #1300, opened 2019-10-08
Upload preferences.json and oref0 version to devicestatus

scottleibrand:
One repeated request from researchers is to be able to tell what preferences a user has enabled, and what version of oref0 they're running.  This uploads such information to the Nightscout devicestatus collection, so it can be included if the user chooses to donate their data to the OpenAPS Data Commons or similar. Sensitive PII from preferences.json is redacted prior to upload.


scottleibrand:
This has been working well and is ready to merge to dev IMO.

danamlewis:
LGTM

PR #1304, opened 2019-10-21
Travis CI: Upgrade to Python 3.8 and add more flake8 tests

cclauss:


scottleibrand:
What exactly does this do?

cclauss:
https://flake8.pycqa.org/en/latest/user/error-codes.html

On the flake8 test selection, this PR does not focus on "style violations" (the majority of flake8 error codes that python/black can autocorrect). Instead these tests are focus on runtime safety and correctness:

E9 tests are about Python syntax errors usually raised because flake8 can not build an Abstract Syntax Tree (AST). Often these issues are a sign of unused code or code that has not been ported to Python 3. These would be compile-time errors in a compiled language but in a dynamic language like Python they result in the script halting/crashing on the user.
F63 tests are usually about the confusion between identity and equality in Python. Use ==/!= to compare str, bytes, and int literals is the classic case. These are areas where a == b is True but a is b is False (or vice versa).
F7 tests logic errors and syntax errors in type hints
F82 tests are almost always undefined names which are usually a sign of a typo, missing imports, or code that has not been ported to Python 3. These also would be compile-time errors in a compiled language but in Python a NameError is raised which will halt/crash the script on the user.

scottleibrand:
Ok.  Anything we need to do differently as a result of these changes? Looks like the automated tests are passing on your commit...

cclauss:
No changes needed on the current code but this puts guardrails in place for future code changes.

PR #1318, opened 2019-11-14
Allow to upload profile even when the url ends with / 

tzachi-dar:
(for example: https://site.herokuapp.com/)


scottleibrand:
LGTM. If someone can paste in your testing results, I can go ahead and merge.

tzachi-dar:
Here is the picture with url ending with '/'
image

PR #1319, opened 2019-11-15
Remove aCOBpredBGs (accelerated carbohydrate absorption) as it is not used anymore.

tzachi-dar:


tzachi-dar:
Tested for a couple of hours, and works well.

PR #1321, opened 2019-12-03
More comments

scottleibrand:
Add more code comments giving an overview of what determine-basal is doing. Also removes some old comments and TODOs that we're never likely to do, and removes an old unused var missed in a previous cleanup.

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -475,7 +475,7 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
         console.error("Limiting carb impact from",ci,"to",maxCI,"mg/dL/5m (",maxCarbAbsorptionRate,"g/h )");
         ci = maxCI;
     }
-    var remainingCATimeMin = 3; // h; duration of expected not-yet-observed carb absorption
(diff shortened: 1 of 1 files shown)


scottleibrand:
This should be a no-op, so I'll merge tomorrow. If someone can give it a quick look over before that, even better.

lib/determine-basal/determine-basal.js
```diff
@@ -465,7 +477,7 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
         console.error("Limiting carb impact from",ci,"to",maxCI,"mg/dL/5m (",maxCarbAbsorptionRate,"g/h )");
         ci = maxCI;
     }
-    var remainingCATimeMin = 3; // h; duration of expected not-yet-observed carb absorption
+    var remainingCATimeMin = 3; // h; minimum duration of expected not-yet-observed carb absorption
```
danamlewis: should this comment be separated above the line to match the other comments?

danamlewis:
LGTM other than one question about comment placement

lib/determine-basal/determine-basal.js
scottleibrand: No, this is an in-line more-detailed comment, of a style I use occasionally after short variable assignment lines etc.

PR #1322, opened 2019-12-03
Mraa ccprog install

scottleibrand:
@cluckj I just found this from 6 months ago. Did we end up re-implementing it elsewhere?


cluckj:
Yes, the MRAA install part of this was implemented in #1302, but pinned to MRAA v1.7.0 (as the latest version, v2.0.0, wasn't working).

The ccprog stuff is in the documentation as of https://github.com/openaps/docs/pull/1519, but I don't see why that part couldn't be automated as well?

scottleibrand:
Ok, updated to just add the ccprog stuff to oref0-mraa-install.

PR #1323, opened 2019-12-03
Only install local hostspot if selected

scottleibrand:
When upgrading rigs, we often see them go offline after updating their hostapd or dnsmasq packages, necessitating a hard reboot before we can log back in and re-run oref0-setup. This PR only installs such local-only hotspot packages if the user wishes to use them, which should improve reliability for everyone else.

This PR also fixes the xdrip-js setup to run `cgm-transmitter` automatically.  If desired, we could commit or PR that to dev separately: the commit is https://github.com/openaps/oref0/commit/72e6a6e6121d2e84c310e7780f316e0bad54897f

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ python-steps: &python-steps
   install: pip install flake8
   script:
     # stop the build if there are Python syntax errors or undefined names
-    - flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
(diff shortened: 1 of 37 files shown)


danamlewis:
Fine keeping 72e6a6e in as long as they both get merged in soon-ish to dev. This LGTM.

scottleibrand:
We probably need (someone) to re-run oref0-setup with and without this option to confirm everything works as expected.

scottleibrand:
If no objections, it'd probably be easiest to merge this to dev for testing.

PR #1326, opened 2019-12-03
fix glucose-noise lastDelta logic error

jpcunningh:
This makes the noise calculation a little more sensitive to trend direction changes by inserting a missing lastDelta assignment used to increase the weight of the glucose deltas.


jpcunningh:
I have run it through a few loops successfully.

Worse case is noise will be a little more aggressive than some users may prefer. I think it's good to merge.

PR #1329, opened 2019-12-04
Fix bug when parsing most recent date

jakobsandberg:
There's a bug when we are looking at the undefined `then` entry when parsing the date for `now`. This PR fixes that bug, does a slight refactor to reduce bugs like this in the future, and adds a test to make sure we're falling back to the correct `dateString` property.

### DIFF
--- a/lib/glucose-get-last.js
+++ b/lib/glucose-get-last.js
@@ -6,7 +6,7 @@ var getLastGlucose = function (data) {
     });
 
     var now = data[0];
-    var now_date = now.date || Date.parse(now.display_time) || Date.parse(then.dateString);
+    var now_date = now.date || Date.parse(now.display_time) || Date.parse(now.dateString);
     var change;
     var last_deltas = [];
     var short_deltas = [];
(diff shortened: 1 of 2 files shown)


scottleibrand:
Can you describe any real-world scenario that has the particular combination of input data that causes problems here? I’d like to understand what this impacts, and what will change from fixing it.

jakobsandberg:
I didn't run into this through a real world scenario, just reading through the source code.

I noticed that we're using the properties `display_time` and `dateString` as fallbacks when the `date` property is unavailable. However, when we fallback to `dateString` we are accessing `dateString` from the wrong object (we are trying to access `dateString` from `then`, which is not yet defined, instead of `now`).

Based on the existing code, I assume there is a scenario where `dateString` is available but both date and display_time are unavailable, thus we fallback to dateString. Otherwise the existing fallback logic is unnecessary and can be removed. I can spend some time looking into data sources to see if this condition is ever met.

scottleibrand:
Ok. I suspect that scenario only occurs when running autotune against non-OpenAPS data sources, or something like that.

PR #1330, opened 2019-12-05
Warn users that the status screen is hard coded to use myopenaps

tepidjuice:


jakobsandberg:
Rather than printing a warning, what do you think about configuring the `openapsDir` variable automatically after we clone `openaps-menu`?

scottleibrand:
I think that would also be a good idea, but for now I'll go ahead and merge this, as it's better than nothing.

PR #1331, opened 2019-12-06
Add public ip information to status file

juehv:
I want to show the public ip in my system menu on the disply and don't want to do an extra ping :)
This code is tested on my pi rig with explorer-HAT (for around two weeks).


bin/oref0-online.sh
```diff
@@ -141,7 +141,7 @@ function check_ip {
         return 1
     else
         echo $PUBLIC_IP
-        touch /tmp/hasPublicIp
+        echo $PUBLIC_IP > /tmp/hasPublicIp
```
jakobsandberg: `hasPublicIp` was a great file name when we were treating the file as a flag, but now that we're storing the value of the IP in the file, should we rename the file to something like `publicIp`?

juehv: I don't care about the exact name. I just need the feature :) So feel free to change it ;)

scottleibrand: Like that?

PR #1332, opened 2019-12-06
Add a shutdown function for pi based rigs

juehv:
Added code to shutdown a pi based rig when battery level is low. I tested this code on my pi rig with explorer HAT several times using my lab bench power supply and one time with a standard 18650 cell.

### DIFF
--- a/bin/oref0-cron-every-15min.sh
+++ b/bin/oref0-cron-every-15min.sh
@@ -15,6 +15,11 @@ if is_edison; then
     sudo ~/src/EdisonVoltage/voltage json batteryVoltage battery | jq .batteryVoltage | awk '{if ($1<=3050)system("sudo shutdown -h now")}' &
 fi
 
+# proper shutdown of pi rigs once the battery level is below 2 % (should be more than enough to shut down on a standard 18600 ~2Ah cell)
+if is_pi; then
+    sudo ~/src/openaps-menu/scripts/getvoltage.sh | tee ~/myopenaps/monitor/edison-battery.json | jq .battery | awk '{if ($1<2)system("sudo shutdown -h +5")}' &
+fi
+
 # temporarily disable hotspot for 1m every 15m to allow it to try to connect via wifi again
 (
     touch /tmp/disable_hotspot
--- a/bin/oref0-cron-every-15min.sh
+++ b/bin/oref0-cron-every-15min.sh
@@ -17,7 +17,7 @@ fi
 
 # proper shutdown of pi rigs once the battery level is below 2 % (should be more than enough to shut down on a standard 18600 ~2Ah cell)
 if is_pi; then
-    sudo ~/src/openaps-menu/scripts/getvoltage.sh | tee ~/myopenaps/monitor/edison-battery.json | jq .battery | awk '{if ($1<2)system("sudo shutdown -h +5")}' &
+    sudo ~/src/openaps-menu/scripts/getvoltage.sh | tee ~/myopenaps/monitor/edison-battery.json | jq .battery | awk '{if ($1<2)system("sudo shutdown -h now")}' &
 fi
 
 # temporarily disable hotspot for 1m every 15m to allow it to try to connect via wifi again


scottleibrand:
Does this interact in any way with the “radio reboot” cancel code? Since you’re using +5 instead of now, I worry the shutdown might get canceled.

juehv:
You're right! I didn't think about the radio reboot. I did work in my testing but that does not mean it works every time... I'll change that.

Dune-jr:
If it is not out of place, I would like to express a concern about this. I have not tested this patch, but my pi0 rig shows random values (ranging from very low to very high % of battery) when it is home, because it is plugged directly to some large battery which voltage changes a lot. Example, as I write those lines:

image

If I understand this right, this false value would cause a shutdown. Is it on my end to adjust some voltage setting, or could something be done about that?

scottleibrand:
Is that the rig battery voltage? There are also uploader battery readings in NS. Can you check the logs on the rig and see if they ever report anomalously low values?

Dune-jr:
@scottleibrand  Yes, that is absolutely the rig battery voltage. It jumped back up to 99% as it usually does:
image

> Can you check the logs on the rig and see if they ever report anomalously low values?

Yes:

```console
$ tail -n 5000 /var/log/openaps/openaps-date.log | grep battery --color=none
2019-12-31 00:04:22 openaps.ns-loop {"batteryVoltage":3920,"battery":76}
2019-12-31 00:10:25 openaps.ns-loop {"batteryVoltage":4140,"battery":98}
2019-12-31 00:15:57 openaps.ns-loop {"batteryVoltage":4210,"battery":100}
2019-12-31 00:23:23 openaps.ns-loop {"batteryVoltage":4019.9999999999995,"battery":89}
2019-12-31 00:30:59 openaps.ns-loop {"batteryVoltage":3230,"battery":2}
2019-12-31 00:36:19 openaps.ns-loop {"batteryVoltage":3160,"battery":1}
2019-12-31 00:42:26 openaps.ns-loop {"batteryVoltage":4130,"battery":98}
2019-12-31 00:51:33 openaps.ns-loop {"batteryVoltage":3650,"battery":17}
2019-12-31 00:56:16 openaps.ns-loop {"batteryVoltage":3180,"battery":1}
2019-12-31 01:01:06 openaps.ns-loop {"batteryVoltage":3260,"battery":2}
2019-12-31 01:06:26 openaps.ns-loop {"batteryVoltage":3260,"battery":2}
2019-12-31 01:13:20 openaps.ns-loop {"batteryVoltage":4150,"battery":99}
2019-12-31 01:17:41 openaps.ns-loop {"batteryVoltage":4200,"battery":100}
2019-12-31 01:22:35 openaps.ns-loop {"batteryVoltage":4070.0000000000005,"battery":94}
2019-12-31 01:27:33 openaps.ns-loop {"batteryVoltage":4190,"battery":100}
2019-12-31 01:33:33 openaps.ns-loop {"batteryVoltage":4130,"battery":98}
2019-12-31 01:38:43 openaps.ns-loop {"batteryVoltage":4190,"battery":100}
2019-12-31 01:45:46 openaps.ns-loop {"batteryVoltage":3270,"battery":2}
2019-12-31 01:50:37 openaps.ns-loop {"batteryVoltage":3930,"battery":77}
2019-12-31 01:55:11 openaps.ns-loop {"batteryVoltage":3300,"battery":2}
2019-12-31 01:59:56 openaps.ns-loop {"batteryVoltage":4180,"battery":99}
2019-12-31 02:05:45 openaps.ns-loop {"batteryVoltage":4160,"battery":99}

```

scottleibrand:
Thanks. That is very odd. What kind of charging circuit are you using? The one on the Explorer HAT? And what kind of battery is it?

Dune-jr:
@scottleibrand I am in the countryside for holidays so I do not have physical access to the device, and electronics is not my strong topic.
It is a pi 0 W with an Explorer HAT. It is connected to a 10 Ah external battery, that is plugged into the wall. Maybe this photo of the device will help you:
image

I will try to look for the reference of this charging device tomorrow.

scottleibrand:
The battery % indicator is measuring the on-board Li-po battery on the JST connector, not the external USB voltage. That’s the battery I was wondering about.

Dune-jr:
@scottleibrand Ah, I understand now. The batteries are outside (e.g. for charging, they are 18650s)

scottleibrand:
Multiple? How are they wired?

Dune-jr:
I am not sure, but this is the case being used: https://www.thingiverse.com/thing:3502320/ It has some details in the description, especially:

> 4 Keystone 209 Battery Contacts

>  Push Keystone 209 battery contacts onto the printed Terminal pieces, making sure that the contacts are facing the correct direction

Not sure what this means though, sorry

scottleibrand:
So you’re using an Explorer HAT rig without an internal battery? That’s a nonstandard setup, and I have no idea what battery voltage it reports when there’s no battery. Is there any reason not to connect a small internal battery?

Dune-jr:
If I'm not mistaken, 18650s need to be taken off to charge, so the rig is plugged into this first.
Since we are home anyway, they are not put back before the morning (that could be changed, but it would still be nice if the device doesn't shutdown)

scottleibrand:
Have you tried leaving the 18650s connected and letting the HAT charge them?

Is there any way to reliably detect when the batteries are removed?

Dune-jr:
- Yes, leaving the 18650s connected works.
- I do not know of any way to detect that the batteries are removed, other than the huge swings in voltage
We did not see the use to feed this rig with batteries when it has been a stay-at-home antenna for a week. It is not a difficult fix (to leave the batteries on) but I imagine that someone who tries to do the same will be left wondering why the system keeps shutting down as soon as it starts

cluckj:
If you have this set up like this all the time (with the rig on external power), you should set up openaps-menu's voltage checker to upload fake values. 

You can do this by changing line 7 in [~/src/openaps-menu/scripts/getvoltage.sh](https://github.com/openaps/openaps-menu/blob/883ef71ac03f030c57f003fb1ae2deed851e6906/scripts/getvoltage.sh#L7) to `echo '{"batteryVoltage":3340,"battery":99}'`

telabu:
Hi, I'm the maker of that case. If you'd like support for the case specifically, feel free to leave comments on the thingiverse page or contact me through the listed methods on that page. 18650s are not part of the standard build in the openaps docs, the case I designed uses them for higher capacity and assumes some minor knowledge of soldering and electronics. It seems you used a printing service to make the case, but without the battery contacts, threaded inserts, and wire leads it is essentially just a plastic shell without proper closures.

scottleibrand:
If you don’t want to completely disable battery voltage uploading, we could also make the shutdown code something that can be disabled by adding something to preferences.json. Or you could just comment out that shutdown line for such a nonstandard setup.

Dune-jr:
Thanks for the tip, @cluckj . Explains why I get so many `99% / 3340mV`
I will do that, although I am afraid this setting might get overridden / conflicts with later updates.
I'm a bit concerned other people that run into the issue will scratch their head for a while, but I don't see any pretty solution here.
I think @scottleibrand's solution would be best.

Dune-jr:
@telabu Thank you for offering support, that's generous. I only handle the software part, and someone more knowledgeable is doing the electronics, soldering and printing. The components aren't all there yet but I don't think that will be an issue.

PR #1325, opened 2019-12-03
simple oref0-simulator

scottleibrand:
Basic usage: just run ~/src/oref0/bin/oref0-simulator.sh. the first time you run it,  it will initialize a /tmp/oref0-simulator directory to work in.  On subsequent runs it’ll randomly introduce BG noise (defaulting to +/- 10 mg/dL per 5m) and show how the loop responds to that.
If you specify any arguments, it’ll use the first argument as the 5m deviation (a positive or negative integer representing how much you want BG to rise or fall), the second argument as the magnitude of the random noise (a positive integer indicating the maximum absolute value of noise to introduce, which defaults to 10), and the third argument, if provided, as a carb entry. Note that you have to have over 30 simulated CGM readings before it stops reporting Not enough glucose data to calculate carb absorption and starts paying attention to carbs you enter.

oref0-backtest allows you to run the simulator on historical data from Nightscout, or from the deviations from an autotune log file:
Usage: NS mode: /usr/bin/oref0-backtest [--dir=/tmp/oref0-simulator] --ns-host=https://mynightscout.herokuapp.com [--start-days-ago=number_of_days] [--end-days-ago=number_of_days] [--start-date=YYYY-MM-DD] [--end-date=YYYY-MM-DD] [--preferences=/path/to/preferences.json] [--autosens-override=/path/to/autosens-override.json]
Usage: file mode: /usr/bin/oref0-backtest [--dir=/tmp/oref0-simulator] /path/to/autotune*.log [--profile=/path/to/profile.json] [--preferences=/path/to/preferences.json] [--autosens-override=/path/to/autosens-override.json]

PR #1335, opened 2019-12-08
Tool to pull profile from Nightscout

scottleibrand:


scottleibrand:
@viq looks like you've already done most of the work needed to help with a TODO I have for the oref0-simulator: `# TODO: download profile.json from Nightscout profile.json endpoint, and copy over to pumpprofile.json for autotuning`
https://github.com/openaps/oref0/pull/1325#issuecomment-562814662
Thanks!

Is there anything we need to know about this branch before pulling it in for simulator development?

viq:
I haven't touched it in a while, it's a bit work in progress and IIRC
currently a bit chatty, but there's a flag inside to turn it off.

It was mostly done with autotune in mind. There are some corner cases it
doesn't deal with yet, for example AndroidAPS local profile in the profile
change event doesn't seem to include time zone.

There are some things that OpenAPS expects that aren't in nightscout that I
just hardcode.

There's one thing missing that I didn't do yet: DIA and insulin profile.
That data does not exist in nightscout profile and I intended to hardcode
some defaults and add flags to select other values, but didn't get around
to that.

Other than that, IIRC I was changing only that file, and it Works For Me.

scottleibrand:
Cool, thanks. That's basically my use-case: running autotune, and the new oref0-backtest / oref0-simulator.

Is there an easy way to resolve `ImportError: No module named texttable` automatically? Ideally I'd like `pip` or something to install that without requiring it to be done manually.

viq:
I could remove that dependency, it's only for printing profile as a table.
Otherwise, most people are on debian derivatives, and there are packages
for that, just `apt install python-texttable` or `python3-texttable`,
depending which python version you want to use (I tried to make it work
with both). I strongly dislike the idea of it performing such modifications
to the system without my knowledge, especially as you get different result
depending whether you do it as root, user, or run it in virtualenv.

scottleibrand:
Ah, ok.  I'll just comment the singleton stuff out for now then. Thanks.

viq:
Apparently it's possible to do conditional imports in python, I'll try
having a look at that.

PR #1339, opened 2019-12-23
Ping to default gateways (per network) to see if local connection is …

tzachi-dar:
…alive.

This helps to distinguish between the case that pi does not have connection to the phone,
and the case that connection to the phone exists but phone has no external IP connection.

Also in the case that both WiFi and BT tethering is enabled one can see which of them is used (the one with the lower metric).

Here is an example to the output (when both WiFi and BT are enabled.):

At Mon 23 Dec 2019 04:50:08 PM IST my local Bluetooth IP is: 192.168.44.226
At Mon 23 Dec 2019 04:50:10 PM IST ping to default gateway 192.168.44.1 (bnep0 metric = 0) passed
At Mon 23 Dec 2019 04:50:10 PM IST ping to default gateway 192.168.44.1 (bnep0 metric = 214) passed
At Mon 23 Dec 2019 04:50:11 PM IST ping to default gateway 192.168.3.1 (wlan0 metric = 302) passed
At Mon 23 Dec 2019 04:50:11 PM IST my public IP is: 37.26.149.235

### DIFF
--- a/bin/oref0-online.sh
+++ b/bin/oref0-online.sh
@@ -43,6 +43,7 @@ main() {
     else
         echo "At $(date) my Bluetooth PAN is not connected"
     fi
+    ping_default_gateways
     echo -n "At $(date) my public IP is: "
     if check_ip; then
         stop_hotspot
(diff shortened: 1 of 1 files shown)


tzachi-dar:
Yws, only diagnostics.

PR #1344, opened 2019-12-31
Deprecate and remove support for adv_target_adjustments

scottleibrand:
adv_target_adjustments was an early experiment, prior to the ZTpredBG changes, attempting to reduce highs. as it exists today, it does not succeed in reducing highs more than it increases lows. For example, in a 7-day oref0-backtest simulator run, it increased % time <70 by 2.6% while only reducing % time >180 by 0.8%.

This PR deprecates and removes adv_target_adjustments to clean up code in preparation for other efforts to accomplish the same thing (for example by extending the ZTpredBG work).

This PR also reorders some code (per earlier TODO comments) to make it more readable.

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -138,6 +138,20 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     var bg = glucose_status.glucose;
     var noise = glucose_status.noise;
 
+// Prep various delta variables.
+    var tick;
+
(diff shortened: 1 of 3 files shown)


scottleibrand:
Any objections to this approach? Does anyone use these features? Does anyone want to do any more testing on whether adv_target_adjustments are useful? Does anyone want to post to FB to see if anyone there uses it? If no to all of those, I'll go ahead and merge it to dev.

PR #1345, opened 2019-12-31
remove deprecated support for max_bg and wide_bg_target_range

scottleibrand:
Wide BG target ranges result in worse outcomes than a single BG target (at a preferred point within the wider range), as they prevent OpenAPS from taking earlier action to correct high and/or low BG.

In order to simplify code paths, allow for easier testing, and make room for future improvements, this PR removes support for wide_bg_target_range and max_bg, and instead sets target to min_bg (which has been the default behavior for some time).

This PR does not (yet) deprecate separate high and low temporary targets, but that would be an obvious next step.

### DIFF
--- a/lib/profile/index.js
+++ b/lib/profile/index.js
@@ -25,9 +25,6 @@ function defaults ( ) {
     // (If someone enters more carbs or stacks more; OpenAPS will just truncate dosing based on 120.
     // Essentially, this just limits AMA/SMB as a safety cap against excessive COB entry)
     , maxCOB: 120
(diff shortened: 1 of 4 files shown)


scottleibrand:
Any objections to this approach? Does anyone use these features? Does anyone want to post to FB to see if anyone there uses it? If no to all of those, I'll go ahead and merge it to dev.

sulkaharo:
LGTM. I guess the subsequent pass after this is merged to refactor the code to only have one BG target (and update unit tests accordingly, need to continue on that).

PR #1346, opened 2020-01-02
Upload NS status when possible.

tzachi-dar:
Before this checkin, ns-loop created the status data, and uploaded it to nightscout (if possible).
But in case of failure, the file will get overwritten.
This means that in the case of no internet, OAPS decisions get lost.
This checkin fixes it by adding timestamps to the files.
When internet is available, the files will be uploaded according to the correct order.
Files older than a day will be deleted.

Testing:
I have prevented uploading of files for a few hours, and when it was enabled again, data was shown correctly in nightscout site.
Decreased timeout, and saw the files being deleted.

### DIFF
--- a/bin/ns-status.js
+++ b/bin/ns-status.js
@@ -142,7 +142,8 @@ if (!module.parent) {
                 battery: safeRequire(cwd + battery_input),
                 reservoir: safeRequire(cwd + reservoir_input),
                 status: requireWithTimestamp(cwd + status_input)
-            }
+            },
+            created_at: new Date()
         };
 
         if (mmtune_input) {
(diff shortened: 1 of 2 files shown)


sulkaharo:
😍 How typical is it currently for the upload skipping to happen? Wondering about the impact on the amount of data sent to Nightscout, as this could impact the NS performance on many levels

tzachi-dar:
I guess that it depends on your network. Our rigs that are at home never miss an upload, and the one that are at school miss a few hours every week.
Probably there are people who don't have internet at day, and connect only at night, and for them there will be some load at night once they connect.
That said, In my experiment, data of a whole day was loaded in 5-10 minutes, and I did not notice any issue with NS.

tzachi-dar:
Please note that I have added another commit to this pr.

scottleibrand:
Tested and working well.

PR #1349, opened 2020-01-11
Add timeout to oref0-get-ns-entries request call

jpcunningh:
Twice in the last week we have had oref0-get-ns-entries hang in the call to request for an hour or more. 

This puts in a 90 second timeout to the call to request to prevent it from hanging.


jpcunningh:
We've been running this on our production rig with no negative impacts and no further hangs.

PR #1342, opened 2019-12-31
Replace expensive node json calls by jq

Dune-jr:
Since I'm having CPU load issues on my pi0, I went to check for some expensive calls that could be optimized away. I found this call:
https://github.com/openaps/oref0/blob/acb0150021e7bf284b6741a9510076a13f690747/bin/oref0-ns-loop.sh#L94-L97

`node /usr/local/bin/json --help` takes 14 seconds on my rasp, so I gladly replaced it with a `jq` call. It's quite inexpensive and I did some testing to make sure it outputs the exact same string.

The only caveat is that it assumes that the time in `cgm/ns-glucose.json` is formatted like
```json
"%Y-%M-%DT%H:%M:%S.000Z"
```
which I believe is safe.

Here is some output (with -100 instead of -10):

- with node json (14s user, 30s system):
<details><summary>CLICK ME</summary>
 <p>

```console
pi@pancreas:~/myopenaps $ cat cgm/ns-glucose.json | node /usr/local/bin/json -c "minAgo=(new Date()-new Date(this.dateString))/60/1000; return minAgo < 100 && minAgo > -5 && this.glucose > 38"
time [
  {
    "_id": "1577798746303128152c1fbb",
    "noise": 1,
    "device": "MIAOMIAO",
    "sysTime": "2019-12-31T13:25:46.000Z",
    "dateString": "2019-12-31T13:25:46.000Z",
    "rssi": 100,
    "direction": "Flat",
    "sgv": 96,
    "date": 1577798746196,
    "unfiltered": 105412,
    "filtered": 105412,
    "type": "sgv",
    "utcOffset": 0,
    "glucose": 96
  },
  {
    "_id": "157779844630559233724d87",
    "noise": 1,
    "device": "MIAOMIAO",
    "sysTime": "2019-12-31T13:20:46.000Z",
    "dateString": "2019-12-31T13:20:46.000Z",
    "rssi": 100,
    "direction": "Flat",
    "sgv": 100,
    "date": 1577798446195,
    "unfiltered": 109647,
    "filtered": 109647,
    "type": "sgv",
    "utcOffset": 0,
    "glucose": 100
  },
  {
    "_id": "1577798146394e2f22f3eb02",
    "noise": 1,
    "device": "MIAOMIAO",
    "sysTime": "2019-12-31T13:15:46.000Z",
    "dateString": "2019-12-31T13:15:46.000Z",
    "rssi": 100,
    "direction": "FortyFiveDown",
    "sgv": 105,
[...]

PR #1350, opened 2020-01-12
Add code to run js code on a shared server.

tzachi-dar:
This saves initilaztion time.

Worked for a day, which means testing only started.


tzachi-dar:
This is the initial version of the code that runs on a shared node.

ns-status and oref0-normalize-temps have changed.

This contains the changes to the js files, but does not contain changes to shell script to start the node, and make sure it stays alive.

tzachi-dar:
As for running "json" commands:
json itself can not be used directly from another module.
So, our options are:
1) make a short change to allow it to support remote calls.
2) Since we are only making use of a small part of it, implement this outself. (should also be a short code - this is code to go over json files and filter it).
3) continue with https://github.com/openaps/oref0/pull/1342 (assuming that all issues there have been solved)

Dune-jr:
Thank you for that work, @tzachi-dar! Looking forward to test that

tzachi-dar:
I have added some more code to harden the solution.
@Dune-jr you can start testing if you wish.
Please note that you need to make sure that oref0-shared-node-loop is on the path.
Right now it is on bin/oref0-shared-node-loop.sh so you will probably have to create a symbolic link for it.

@scottleibrand still waiting for your input regarding https://github.com/openaps/oref0/pull/1350/#issuecomment-573369555
All solutions will work well, but still looking for your prefered method to solve the issue.

scottleibrand:
normalize-temps error:
/home/sleibrand/src/oref0/bin/oref0-normalize-temps.js:64
   var result = oref0-normalize-temps(command)
                               ^

ReferenceError: oref0 is not defined
    at Object.<anonymous> (/home/sleibrand/src/oref0/bin/oref0-normalize-temps.js:64:32)

tzachi-dar:
This is replaced by https://github.com/openaps/oref0/pull/1352

PR #1352, opened 2020-01-26
Tzachi share node2

tzachi-dar:
This pr adds code to run on shared node.

Solution overview:
1) A shared node is created to accept different requests.
2) js modules have been changed to return their values based on a function call.
3) "Clients" use a unix socket to talk with shared node.
More details:
Keeping shared node alive:
1) Shared node is started every time that it is killed.
2) Every minute, ping is sent to the shared node. If it does not answer it, it is killed.
3) On my testing, I did not see it never failing.

Return value of shared node:
The result of the node consist of 3 things:
1) The exit code of the function.
2) The output to stdout of the js code.
3) The output to stderr of the js code (was not still used).

Testing:
Test rig has been looping for a long time without any problem. It was able to complete pump loops within 5 minutes except where communication with pump was problematic.
All input to the shared node is captured, and copied to a shared directory.
Later, the old code and the new code have been running on this input. No difference was found between the two runs.
With this change the rigs become much less loaded (and very likely this is also a needed step in reducing battery consumption).

### DIFF
--- a/bin/mm-format-ns-treatments.sh
+++ b/bin/mm-format-ns-treatments.sh
@@ -18,7 +18,7 @@ EOT
 # | json -e "this.type = 'mm://openaps/$self'" \
 model=$(jq -r . $MODEL)
 
-oref0-normalize-temps $HISTORY \
+run_remote_command "oref0-normalize-temps $HISTORY" \
   | jq '[ .[]
     | .medtronic = ( [ "mm://openaps/'$self'/", ( . | if ._type then ._type else .eventType end ) ] | join("") )
     | .created_at = if .created_at then .created_at else .timestamp end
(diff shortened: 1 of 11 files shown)


Dune-jr:
@tzachi-dar Shouldn't this PR be for a merge on `dev` instead of `master`?

tzachi-dar:
> @tzachi-dar Shouldn't this PR be for a merge on `dev` instead of `master`?

Sure, I have fixed this now.

scottleibrand:
Is there any special installation required to use this? I'm getting failures using run_remote_command ns-status. It could also be due to merge conflicts with https://github.com/openaps/oref0/pull/1346 that I resolved incorrectly. Could you merge those two branches properly and push the results somewhere I can test them?

tzachi-dar:
I have merged both branches into https://github.com/openaps/oref0/pull/1359

scottleibrand:
Is this still needed, or is it superceded by #1370 ?

scottleibrand:
I'm going to close this on the assumption it's no longer needed after #1370. If that's incorrect, we can reopen.

PR #1357, opened 2020-02-03
floating carbs

scottleibrand:
If `floating_carbs: true`, then dose slightly more aggressively by using all entered carbs for calculating COBpredBGs. This avoids backing off too quickly as COB decays. Even with this option, oref0 still switches gradually from using COBpredBGs to UAMpredBGs proportionally to how many carbs are left as COB.

In our testing with `floating_carbs: true` and a higher BG target during the day (120 instead of 100), we've so far found that oref0 with SMB and UAM is better able to keep BG in range and avoid high BGs after COB decays.

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -545,10 +545,15 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
         // avoid divide by zero
         cid = 0;
     } else {
-        cid = Math.min(remainingCATime*60/5/2,Math.max(0, meal_data.mealCOB * csf / ci ));
+        if (profile.floating_carbs === true) {
+            // with floating_carbs preference set, use all carbs, not just COB
+            cid = Math.min(remainingCATime*60/5/2,Math.max(0, meal_data.carbs * csf / ci ));
+        } else {
+            cid = Math.min(remainingCATime*60/5/2,Math.max(0, meal_data.mealCOB * csf / ci ));
+        }
     }
     // duration (hours) = duration (5m) * 5 / 60 * 2 (to account for linear decay)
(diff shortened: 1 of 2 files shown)


sulkaharo:
Does this indicate the COB model is not accurate? Or that the COB model as such is fine, but the model that's currently being used not accounting for proteins / fats and not decaying carbs as fast as now accounts for the total meal absorption better?

scottleibrand:
No, I don't think it means the COB model is inaccurate, just that we are using it more conservatively than may be ideal. With the default min_5m_carbimpact value of 8 mg/dL/5m, COB pretty reliably decays to zero by the time carb absoroption stops, but sometimes there is residual carb absorption after it decays to zero. With lower values like 2 mg/dL/5m, COB doesn't always decay completely when there are "disappearing carbs" (situations where carbs are digested but don't have their full predicted effect on BG). Floating carbs aims to preserve proper (conservative) COB decay (so users can see how much COB they almost-certainly have left), but allow slightly more insulin dosing as COB decays, if and only if high rates of carb absorption (high deviations) are observed.

tim2000s:
I'm finding that floating carbs isn't running Autotune correctly overnight. 
Every time it looks like it has run, but there's no change to the basal profile.

scottleibrand:
I'm not seeing that.  Can you take a look at the latest autotune.2020-02-*.log file and see if you can tell why? Feel free to post a suitably redacted autotune log file / snippet if you'd like us to take a look at it.

tim2000s:
My mistake. User error on what I was diffing.

scottleibrand:
We ended up not using this, and didn't merge it to 0.7.1. Is anyone else still using it who wants to try to get it working and tested for 0.7.2? Or should we close it out?

PR #1359, opened 2020-02-03
Tzachi share node ns status

tzachi-dar:
This version includes oref0-shared code + the code of upload-status-when-possible.

It is based on the current head version.
I'll only be able to test it tonight.

### DIFF
--- a/bin/mm-format-ns-treatments.sh
+++ b/bin/mm-format-ns-treatments.sh
@@ -18,7 +18,7 @@ EOT
 # | json -e "this.type = 'mm://openaps/$self'" \
 model=$(jq -r . $MODEL)
 
-oref0-normalize-temps $HISTORY \
+run_remote_command "oref0-normalize-temps $HISTORY" \
   | jq '[ .[]
     | .medtronic = ( [ "mm://openaps/'$self'/", ( . | if ._type then ._type else .eventType end ) ] | join("") )
     | .created_at = if .created_at then .created_at else .timestamp end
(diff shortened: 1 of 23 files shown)


tzachi-dar:
Please note that this assumes that " oref0-shared-node-loop" is in the path.
Try running "which  oref0-shared-node-loop".
It should be in /usr/local/bin/oref0-shared-node-loop

this is a symbolic link. See output of the following command. 
ls -l /usr/local/bin/oref0-shared-node-loop
lrwxrwxrwx 1 root root 55 Jan 13 00:04 /usr/local/bin/oref0-shared-node-loop -> ../lib/node_modules/oref0/bin/oref0-shared-node-loop.sh

tzachi-dar:
I have run with it tonight and it seems to work fine.
Please try this version.

scottleibrand:
Looks like the problem is `bash: socat: command not found` - we'll need to add an install step for that.

scottleibrand:
I've been running this successfully on a couple of rigs for a few months now, so I think it's good to merge. Anything else we should do/test before doing so?

tzachi-dar:
I think this one was replaced by 1361.
I think that the next one to be merges is https://github.com/openaps/oref0/pull/1370

scottleibrand:
Is this a subset of #1370? If so, I’d prefer to merge this first, as it’s been tested, and then see what’s left in 1370 and test that.

tzachi-dar:
No #1370 is a PR by itself that should be viewed against dev branch.

PR #1361, opened 2020-02-14
shared-node

scottleibrand:
Replacement for #1359 and #1352. Installs socat on all platforms so shared-node will work.

### DIFF
--- a/bin/oref0-ns-loop.sh
+++ b/bin/oref0-ns-loop.sh
@@ -93,7 +93,6 @@ function glucose_fresh {
 
 function find_valid_ns_glucose {
     dir_name=~/test_data/find-glucose$(date +"%Y-%m-%d-%T")
-    echo dir_name = $dir_name
     mkdir -p $dir_name
     cp cgm/ns-glucose.json  $dir_name
     date --iso-8601=seconds >  $dir_name/date_string
--- a/bin/oref0-ns-loop.sh
+++ b/bin/oref0-ns-loop.sh
@@ -92,8 +92,13 @@ function glucose_fresh {
 }
 
 function find_valid_ns_glucose {
-    # TODO: use jq for this if possible
-    cat cgm/ns-glucose.json | json -c "minAgo=(new Date()-new Date(this.dateString))/60/1000; return minAgo < 10 && minAgo > -5 && this.glucose > 38"
+    dir_name=~/test_data/find-glucose$(date +"%Y-%m-%d-%T")
+    echo dir_name = $dir_name
+    mkdir -p $dir_name
+    cp cgm/ns-glucose.json  $dir_name
+    date --iso-8601=seconds >  $dir_name/date_string
+
+    run_remote_command 'json -f cgm/ns-glucose.json -c "minAgo=(new Date()-new Date(this.dateString))/60/1000; return minAgo < 10 && minAgo > -5 && this.glucose > 38"'
 }
 
 function ns_temptargets {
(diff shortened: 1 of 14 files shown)


bin/oref0-ns-loop.sh
```diff
@@ -92,8 +92,13 @@ function glucose_fresh {
...
 function find_valid_ns_glucose {
-    # TODO: use jq for this if possible
-    cat cgm/ns-glucose.json | json -c "minAgo=(new Date()-new Date(this.dateString))/60/1000; return minAgo < 10 && minAgo > -5 && this.glucose > 38"
+    dir_name=~/test_data/find-glucose$(date +"%Y-%m-%d-%T")
+    echo dir_name = $dir_name
+    mkdir -p $dir_name
+    cp cgm/ns-glucose.json  $dir_name
+    date --iso-8601=seconds >  $dir_name/date_string
+
+    run_remote_command 'json -f cgm/ns-glucose.json -c "minAgo=(new Date()-new Date(this.dateString))/60/1000; return minAgo < 10 && minAgo > -5 && this.glucose > 38"'
```
scottleibrand: Should this be using $dir_name/ns-glucose.json instead of cgm/ns-glucose.json?

As it is, I get:
```
Found recent valid BG:
dir_name = /root/test_data/find-glucose2020-02-14-19:46:10 [ { "_id": "5e476842403a6cc03ad98637", "device": "removed", "date": 1581738040315, "dateString": "2020-02-15T03:40:40.315Z", "sgv": 136, "direction": "FortyFiveDown", "type": "sgv", "filtered": 189, "unfiltered": 176, "rssi": -97, "noise": 1, "trend": 5, "state": "OK", "status": "OK", "glucose": 136 } ] (NOT VALID JSON: parse error: Invalid numeric literal at line 1, column 9)
```

scottleibrand:
With those fixes, this is now running successfully on one of our Edison rigs, and uploading successfully to NS. Updating wasn't very smooth, so will need to try another rig and see whether all those issues are fixed.

scottleibrand:
Looks like the test_data directory is growing unbounded and using up all the disk space on my rig:

619M    ./root/test_data

scottleibrand:
@tzachi-dar should we just stop storing stuff in ~/test_data, or do we still need that? If we need it, we'll need to add some code to delete old data so it doesn't grow indefinitely.

scottleibrand:
Looks like there are also a bunch of shared_node.(random string) files filling up /tmp/

tzachi-dar:
As for the test data, it is only needed to make sure that we are still getting the same result.
That is one has to go to branch dev, and create the output that would have been generated by the old code. Then move to the share node branch and the script will run it with the new code, and compare results.

If needed, I can send you the script, or you can send me the data and I'll test it.

Then we can remove the code that creates the test data.
As for the temp files in /tmp. If your disk is big enough it is not a problem, but I can also add code that deletes these temp files (probably just delete it after we have read them, or write to a file based on the pid).

So, to summarize:
1) Should I send code to stop collecting test data?
2) Should I send code to delete the temp files?
3) Do you want to verify the test data? How?

Tzachi

tzachi-dar:
By the way, as for installing the two extra files, we can probably add an update script and add there what is needed when updating.
It will run every 15 minutes, and make sure that upgrades go smoothly.
(Needed for this pr, but also for other prs).

scottleibrand:
Yeah, an upgrade script would be a good idea.

scottleibrand:
I haven’t seen any unusual dosing predBG or upload behavior with the new code. If you’ve verified it works identically on your system, I’m fine removing the test stuff.

tzachi-dar:
I have updated my pr:
1) remove temp files.
2) Stop writing to test_data
3) Added an upgrade script that runs every 15 minutes. (please note that I call it on it's path directly otherwise it will require an update script by itself).

scottleibrand:
Is this tested and ready to merge?

tzachi-dar:
Give me 2-3 more days for final testing, and I'll get back to you.

tzachi-dar:
OK, did some more testing with this code, and more running on it with a live system.
Everything is OK.
Let's merge it.

Issue #1362, opened 2020-02-14
Is autotune using the correct values for Loop users with Omnipod


bartsluimer:
Loop with Omnipod is using a different timing with basal as Medtronic. This is causing a big difference in for example the day to day report inside Nightscout. 
The question is what values is autotune using. 

Example on difference between bolus and basal wit Tidepool and Nightscout:
![64612394-f85e4a80-d3d3-11e9-996d-2fe064bd0280](https://user-images.githubusercontent.com/25899216/74513020-e1370200-4f09-11ea-9cc0-f4aa67ccaa51.png)
![64612213-8c7be200-d3d3-11e9-9c30-ebacb61f3e15](https://user-images.githubusercontent.com/25899216/74513031-e5631f80-4f09-11ea-90ed-1d31ae174ecf.png)

Tidepool values are correct.

Loop is now sending different values, to prevent any mistakes. Only is Autotune using these new timing and amount value?:

**Old TempBasal**

{
"_id": {
"$oid": "5d53811e7cc3d56f0c3fe930"
},
"timestamp": "2019-08-14T03:28:47Z",
"eventType": "Temp Basal",
"enteredBy": "loop://Marion iPhone 7+",
"duration": 4,
"temp": "absolute",
"absolute": 2.55,
"rate": 2.55,
"created_at": "2019-08-14T03:28:47.000Z",
"utcOffset": 0
}

**New TempBasal**

{
"_id": {
"$oid": "5d792b1f7cc3d56f0c739952"
},
"absolute": 1.75,
"rate": 1.75,
"temp": "absolute",
"duration": 4.767875816424688,
"timestamp": "2019-09-11T17:12:57Z",
"enteredBy": "loop://Marion iPhone 7+",
"eventType": "Temp Basal",
"amount": 0.1,
"created_at": "2019-09-11T17:12:57.000Z",
"utcOffset": 0
}

scottleibrand:
The easiest way to test this would be using the NS URL of someone using Omnipod Loop. Do we have anyone willing to disclose their NS URL for that purpose?

bartsluimer:
Will delete the url once finished. https://amklaassen.herokuapp.com/

sethgagnon:
Thanks @bartsluimer for opening. So, this is related to this Issue then as well:
https://github.com/nightscout/cgm-remote-monitor/issues/4971

bartsluimer:
[...]

PR #1364, opened 2020-02-20
calculate effective temp basal rate for Omnipod Loop

scottleibrand:
Fix for #1362 to accurately calculate IOB for Omnipod Loop users running Autotune from the `amount` field added to Temp Basal records to record how much insulin was actually delivered while the temp basal was running.

Needs further testing. Side-by-side autotune runs on a single day's worth of Omnipod Loop data show consistent small differences in calculated IOB, resulting in some minor differences in recommendations.

### DIFF
--- a/lib/iob/history.js
+++ b/lib/iob/history.js
@@ -326,6 +326,11 @@ function calcTempTreatments (inputs, zeroTempDuration) {
             var temp = {};
             temp.rate = current.rate;
             temp.duration = current.duration;
+            // Loop reports the amount of insulin actually delivered while the temp basal was running
+            // use that to calculate the effective temp basal rate
+            if (typeof current.amount !== 'undefined') {
+                temp.rate = current.amount / current.duration * 60;
+            }
             temp.timestamp = current.timestamp;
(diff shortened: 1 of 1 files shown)


naromero77:
I have this set-up on my desktop locally. What is the best way to test this?

sethgagnon:
just an update that i have been using the omnipod-iob branch for quite some time now and the results definitely seem better using autotune. i can share my latest log file a recommendations file with you @scottleibrand . just dont want to post here with my NS site details in the log files.

scottleibrand:
I've put most of my algorithm work on pause while we work on things pertaining to COVID-19 (like CoEpi.org). If someone else wants to pick this up, though, I'd be happy to help however I can.

scottleibrand:
At this point I'd be happy to just merge this, assuming no one has seen any problems using it to date. Is there any other testing anyone would like to do first?

sethgagnon:
I have been using this branch for quite some time and it seems fairly accurate for me! Thanks!

scottleibrand:
Ok, thanks. Merging.

PR #1366, opened 2020-02-24
Installer fixes

cluckj:
- Changes the way nodejs/npm are installed: Edison rigs will get it v8 from nodesource (Jessie/Stretch), Pi rigs will get it from the distro (Raspbian Buster)
- openaps toolchain will be installed from the latest version on git, rather than via pypi.

### DIFF
--- a/bin/openaps-packages.sh
+++ b/bin/openaps-packages.sh
@@ -12,30 +12,35 @@ apt-get install -y sudo
 sudo apt-get update && sudo apt-get -y upgrade
 sudo apt-get install -y git python python-dev software-properties-common python-numpy python-pip watchdog strace tcpdump screen acpid vim locate lm-sensors || die "Couldn't install packages"
 
-# We require jq >= 1.5 for --slurpfile for merging preferences. Debian Jessie ships with 1.4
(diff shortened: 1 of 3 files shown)


bin/openaps-packages.sh
```diff
@@ -12,30 +12,35 @@ apt-get install -y sudo
...
+   if getent passwd edison; then
+     # Only on the Edison, use nodesource setup script to add nodesource repository to sources.list.d, then install nodejs (npm is a part of the package)
+     curl -sL https://deb.nodesource.com/setup_8.x | bash -
+     sudo apt-get install -y nodejs=8.* || die "Couldn't install nodejs"
+   else
+     sudo apt-get install -y nodejs npm || die "Couldn't install nodejs and npm"
+   fi
+   
+   # Upgrade npm to the latest version using its self-updater
+   sudo npm install npm@latest -g || die "Couldn't update npm"
```
scottleibrand: What version will we end up with on each platform from this?

cluckj: This should install node v8 on the Edison (both Jessie & Stretch), node v8 on Pi's with Stretch, and node v10 on all Pi's with Buster. The version of npm should be the same across all platforms because it uses the self-updater.

PR #1373, opened 2020-05-16
Adding new functionality to flask server

Kurbezz:
Oref stoping by edit oref.json file (/oref_enable and /oref_disable)
Getting about pump serial number, pump status and etc
Getting autotune recommendations

Has authorization:
Reboot system (has authorization)
Append local temp target

Need pump loop kill and has authorization:
* Enter bolus
* Press keys (Using for disable alarms on pump)
* Set temp basal
* Suspend/resume pump

### DIFF
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -1238,8 +1238,9 @@ if prompt_yn "" N; then
     copy_go_binaries
     move_mmtune
 
-    # create oref0.json
+    # create oref0.json and flask.json
     echo "{ \"OREF0_CAN_RUN\": true }" > "$directory/oref0.json"
+    echo "{ \"AUTHORIZATION_ENABLED\": true }" > "$directory/flask_server.json"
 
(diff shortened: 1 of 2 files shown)


cluckj:
Remote control of a pump isn't consistent with the safety guidelines of this project.

TrachYuri:
I've merged it into my fork. And this what I think: 
1. It will be better to split the fix for glucose JSON and additional flask functionality. In that case, it will have more chances to be merged in dev. And it will be helpful for many people. @Kurbezz thanks for it.
2. Some of added requests would be better to make POST instead of GET(append_local_temptarget for example). Because Http shortcuts don't allow to send request body with GET request. On iOS started from iOS 13 it's impossible to do using system methods at all.
My point in this is: Using OpenAps is a risk. So why don't we allow people to take more risk if that does it consciously? It's would be ok if we have some hidden property to switch on that additional functionality.
But, it only my opinion ;))
Thanks for your job and have a nice weekend.

PR #1380, opened 2020-07-11
Record the input data of the node operations.

tzachi-dar:
This information is needed to compare the current output to the output of the shared node.

This PR does the following:
1) Change the bolus.js not to have a hard coded file name in it. (For all other programs, the files used is input are passed as parameters. For this program, one file was hard coded in the code. This pr fixes this as well).
2) For each of our node operation, creates a directory under ˜/test_data and stores the input of the program.
3) Add a script src/oref0/tests/run_node_programs.sh

### DIFF
--- a/bin/mm-format-ns-treatments.sh
+++ b/bin/mm-format-ns-treatments.sh
@@ -17,9 +17,11 @@ EOT
 
 # | json -e "this.type = 'mm://openaps/$self'" \
 model=$(jq -r . $MODEL)
+dir_name=~/test_data/normalize-temps$(date +"%Y-%m-%d-%H%M")
+mkdir -p $dir_name
+cp  $HISTORY  monitor/iob.json $dir_name
 
-
-run_remote_command "oref0-normalize-temps $HISTORY" \
+run_remote_command "oref0-normalize-temps $HISTORY monitor/iob.json" \
   | jq '[ .[]
     | .medtronic = ( [ "mm://openaps/'$self'/", ( . | if ._type then ._type else .eventType end ) ] | join("") )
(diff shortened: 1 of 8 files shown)


tzachi-dar:
One thing that still bothers me is that there are still some functions that don't pass the time as input param.
For example: 
https://github.com/openaps/oref0/blob/5313363b5d564c274ba928b1ec7d32b0b64343b9/lib/profile/index.js#L156
https://github.com/openaps/oref0/blob/5313363b5d564c274ba928b1ec7d32b0b64343b9/lib/profile/carbs.js#L5
https://github.com/openaps/oref0/blob/5313363b5d564c274ba928b1ec7d32b0b64343b9/lib/profile/targets.js#L11

This forces that clock of the pi has to be updated in order to have the tests run corrctly.
If there are no objections, I'll also fix this.

tzachi-dar:
Any progress with reviewing this?

scottleibrand:
If this is ready to test, I could put it on a rig. Not sure my reviewing the code in detail is likely to catch anything that tests and basic "does it run" testing wouldn't.

You could also run it with the oref0-simulator using oref0-backtest to make sure it works well for that use case.

tzachi-dar:
Thanks for your help.

This PR only captures test data from real life scenarios, so if you apply this, we will have a lot of data to run both versions and see if there is a change.

https://github.com/openaps/oref0/pull/1370 is the pr that does the real work. It is ready to test. We can also go and apply that version but is a little more risky.

scottleibrand:
I'm testing #1370 on a rig w/o issue so far, so probably won't use this branch for testing. If no objections, I'll go ahead and close it in favor of #1370.

Issue #1383, opened 2020-08-23
0.7.0 Autotune: UAM always categorized as basal even if categorize-uam-as-basal setting is false


Philoul:
I made hundred tests to compare results with or without categorized-uam-as-basal setting
=> Tests done with AutotuneWeb and on a virtual machine on my computer (I use AndroidAPS)

UAM data are always categorized as basal whatever ```categorize-uam-as-basal``` setting (even in my worst days with several hours of uam data)

This is confirmed by a code analysis of categorized.js file lines 289-396
```
    if (opts.categorize_uam_as_basal) {
        console.error("--categorize-uam-as-basal=true set: categorizing all UAM data as basal.");
        basalGlucoseData = basalGlucoseData.concat(UAMGlucoseData);
    } else if (CSFLength > 12) {
        console.error("Found at least 1h of carb absorption: assuming all meals were announced, and categorizing UAM data as basal.");
        basalGlucoseData = basalGlucoseData.concat(UAMGlucoseData);
    } else {
        ...
    }
```

It's just quite impossible to have test line 392  ```else if (CSFLength > 12)``` false: if we have only one meal announced with 20 or 25g of carbs, we will have at least one hour of CSF data so even if we have several hours of UAM data, all uam data will always be categorized as basal...
=> I think code between lines 396 and 424 is never executed

I also made some tests replacing line 392 by  ```else if (UAMLength < 12)``` 
=> With this line if I have less than 1 hour of UAM data (assuming all meals were announced) UAM data is categorised as basal)
=> But if I have more than one hour of UAM data then UAM are categorized between ISF and basals according to lines between 396 and 424
I think with this modification Autotune works as expected (there are security warning in documentation if we use ```--categorize-uam-as-basal=true```...)

scottleibrand:
Yes, this is on purpose. We originally had the behavior you expected, but too many people were seeing "problems with autotune failing to converge, and instead constantly trying to reduce basals until it hits safety limits, because too much data that would otherwise indicate basals are too low is being excluded as UAM", as described at in the PR where we added the code you found: 
https://github.com/openaps/oref0/pull/1195

Philoul:
Sorry wrong manip

PR #1384, opened 2020-08-23
Autotune: Modify test for correct meals announcement

Philoul:
see https://github.com/openaps/oref0/issues/1383

With current code uam data are always categorized as basal even if categorize-uam-as-basal is false (default setting)

I propose this modification to have autotune works as expected in documentation (there is a security warning in documentation for this setting...)

### DIFF
--- a/lib/autotune-prep/categorize.js
+++ b/lib/autotune-prep/categorize.js
@@ -394,8 +394,8 @@ function categorizeBGDatums(opts) {
     if (opts.categorize_uam_as_basal) {
         console.error("--categorize-uam-as-basal=true set: categorizing all UAM data as basal.");
         basalGlucoseData = basalGlucoseData.concat(UAMGlucoseData);
-    } else if (CSFLength > 12) {
-        console.error("Found at least 1h of carb absorption: assuming all meals were announced, and categorizing UAM data as basal.");
+    } else if (UAMLength < 12) {
+        console.error("Found less than 1h of uam absorption: assuming all meals were announced, and categorizing UAM data as basal.");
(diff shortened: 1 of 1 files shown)


scottleibrand:
The description isn't quite correct: UAM data are categorized as basal for individuals who sometimes announce meals. For those doing full UAM all the time (what the UAM detection code was written for), they're unlikely to always get UAMs categorized as basal.

Per https://github.com/openaps/oref0/issues/1383#issuecomment-685291622, the current behavior is what we found to work best during development, for multiple individuals. If you want to propose a modification, we'll need to demonstrate across multiple people's Nightscout data that autotune correctly converges on realistic basals.

Philoul:
Ok, thank you for the answer.
I understand better the reason for the current code (maybe the log message and doc could be improved a little bit to be consistent with your answer ...).
If this has already been tested I close the PR

scottleibrand:
If you can think of any ways to improve the documentation or good comments or similar, I would really appreciate a pull request for those. That’s one of the hardest things to do as someone who is so close to the algorithms: figure out how to explain them to people who aren’t.

Philoul:
Ok, I'll try to propose some little improvements in log message (easier to do) or doc...
Today I'm still working on raw iob calculation for autotune in my androidAPS plugin, because I still not have same results than oref0 calculation :( 
AutotuneCore is ok (exactly same results with 3 digits) but not autotunePrep due to `getIOB(IOBInputs)[0]` call (current functions in AAPS doesn't integrate a "currentBasal" average value as reference and I have difficulties to understand exact calculation algorithm in iob lib of oref0...)

PR #1385, opened 2020-09-03
Add mills to status upload to reduce NS client workload

jpcunningh:
There is a corresponding NS PR to use the mills fields if they are available to drastically reduce the number of times moment has to parse the date strings when scrolling in NS.

### DIFF
--- a/bin/ns-status.js
+++ b/bin/ns-status.js
@@ -3,6 +3,7 @@
 
 var os = require("os");
 var fs = require('fs');
+var moment = require("moment");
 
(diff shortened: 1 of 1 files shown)


scottleibrand:
Will this have any effect (positive or negative) on folks running older versions of Nightscout without the corresponding NS PR?

jpcunningh:
There should be no effect on folks running older versions of Nightscout because the text timestamp fields are retained for backward compatibility.

scottleibrand (approved):

PR #1389, opened 2020-12-06
Replace git with https for install and setup scripts

Foxy7:
openaps-install.sh and oref0-setup.sh installation bug due to git:// instead of https://

Line 74 of openaps-install.sh calls the git clone via `git://` instead of `https://` for some reason my PC doesn't like it an always fails.

Suggested change, replacing with `git clone https://github.com/openaps/oref0.git`

https://github.com/openaps/oref0/blob/ee154b044948c76f064d59db93cc3f7b7691a679/bin/openaps-install.sh#L74

After this line 74 call, i get the following:

`Cloning into 'oref0'...
fatal: unable to connect to github.com:`

Work around is simply to use `cd src && git clone git://github.com/openaps/oref0.git`

Checking the code i found the following git call too, so requrest both changed at the same time.

https://github.com/openaps/oref0/blob/ee154b044948c76f064d59db93cc3f7b7691a679/bin/oref0-setup.sh#L736

Thanks.

### DIFF
--- a/bin/openaps-install.sh
+++ b/bin/openaps-install.sh
@@ -73,7 +73,7 @@ sed -i "s/daily/hourly/g" /etc/logrotate.conf
 sed -i "s/#compress/compress/g" /etc/logrotate.conf
 
 curl -s https://raw.githubusercontent.com/openaps/oref0/$BRANCH/bin/openaps-packages.sh | bash -
-mkdir -p ~/src; cd ~/src && ls -d oref0 && (cd oref0 && git checkout $BRANCH && git pull) || git clone git://github.com/openaps/oref0.git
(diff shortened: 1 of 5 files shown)


scottleibrand:
Have you tested this change? What results did you get?

Foxy7:
I will test this. Haven't done as yet.

Foxy7:
just found an issue with my install of Dev it's missing EdisonVoltage and i think related my git issue vs https, so have installed it manually running the git clone using https://

https://github.com/openaps/oref0/blob/acb0150021e7bf284b6741a9510076a13f690747/bin/oref0-setup.sh#L1017

Foxy7:
@scottleibrand any issue for you if i make a PR to change all `git://github.com` to be `https://github.com` in the set up scripts?

scottleibrand:
Sure: I can’t think of any reason that would break anything.

Foxy7:
thanks, will get on it.

PR #1391, opened 2021-01-04
Update app.py to show sgv via workaround

Foxy7:
As per @renegadeandy pull request https://github.com/renegadeandy/oref0/pull/1

just tested it and it works nicely, so making PR for Dev.

### DIFF
--- a/www/app.py
+++ b/www/app.py
@@ -59,7 +59,7 @@ def enacted():
 
 @app.route("/glucose")
 def glucose():
-    if os.path.getmtime(myopenaps_dir + "xdrip/glucose.json") > os.path.getmtime(myopenaps_dir + "monitor/glucose.json"):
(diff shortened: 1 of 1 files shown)


scottleibrand:
I don’t use the enlite and haven’t been closely following this issue. Can someone fill me in on what testing has been done to confirm this fixes the issue and doesn’t cause any issues for non-enlite users?

Foxy7:
Hi, i wasn't aware this was enlite related, or at least it isn't for me. 

We're using Dexcom G6, Edison and a Medtronic 715 pump.

I've got two rigs running, one using xdrip+ (overnight) and the other xdrip-js (daytime). And neither local webpages have shown sgv data from nightscout since I've been using OpenAPS. The xdrip-js one does work via Logger if the rig is in range/carried around.

I added the line to both rigs last night and the xdrip+ one has worked, but i see the xdrip-js one hasn't.

I've not done any further testing, perhaps this should be an issue?


Screenshots from this morning, left picture is xdrip-js:

image

scottleibrand:
Sorry, I confused it with another issue. So this only affects the “offline” web page display on the rig?

scottleibrand (approved):

Foxy7:
to confirm, yes only affects the offline webpage.

PR #1394, opened 2021-04-05
Fix python script and allow autotune prep/core to write to file directly

hsghori:
While running the autotune python script i noticed a couple issues:

1. `bin/oref0-autotune.py` was not compatible with python3.
2. The output of `bin/oref0-autotune-core.js` and `bin/oref0-autotune-prep.js` is piped to an output file which was causing autotune to crash because unrelated logs were being piped to the output file.

This PR introduces py3 compatibility (and fixes a couple of style issues) and adds an optional output file argument to the autotune core and prep modules to prevent extraneous data from being written to the files.


bin/oref0-autotune.py
```diff
@@ -161,50 +152,50 @@ def get_nightscout_bg_entries(nightscout_host, start_date, end_date, directory):
...
+            # autotune.$RUN_NUMBER.$DATE.json
             autotune_run_filename = os.path.join(autotune_directory, 'autotune.{run_number}.{date}.json'
                                                  .format(run_number=run_number, date=date.strftime("%Y-%m-%d")))
-            with open(autotune_run_filename, "w+") as output:
-                logging.info('Running {script}'.format(script=autotune_prep))
-                call(autotune_prep, stdout=output, shell=True)
-                logging.info('Writing output to {filename}'.format(filename=autotune_run_filename))
-        
-            # Autotune  (required args, <autotune/glucose.json> <autotune/autotune.json> <settings/profile.json>), 
+            autotune_prep = 'oref0-autotune-prep {ns_treatments} {profile} {ns_entries} {pump_profile} --output-file {autotune_run_filename}'.format(ns_treatments=ns_treatments, profile=profile, ns_entries=ns_entries, pump_profile=pump_profile, autotune_run_filename=autotune_run_filename)
```
hsghori, line 179: `pumpprofile.json` is a required argument to `autotune-prep`. Though FWICT it's the same as `profile.json` so I'm not sure...why it's required here.

bin/oref0-autotune.py
```diff
@@ -25,29 +25,22 @@
 import datetime
 import os, errno
 import logging
+import pytz
 from subprocess import call
 import shutil
+import six
 
 
-DIR = ''
```
hsghori, line 32: IMO we shouldn't be using global variables here.

requirements.txt
```diff
@@ -0,0 +1,3 @@
+requests==2.25.1
```
hsghori, line 1: pin python dependencies.

bin/oref0-autotune.py
scottleibrand: The pumpprofile.json represents the pump's settings. Once autotune has been run, the profile.json should get updated to reflect the autotuned settings, but the pumpprofile.json should keep the pump's settings in order to anchor the range over which autotune can safely adjust settings.

scottleibrand:
Has anyone tested this branch, such as by running a fresh install from it on a rig?

scottleibrand:
@hsghori did you see my question above about testing?

scottleibrand (approved):
Installed this on a rig and manually ran oref0-cron-nightly successfully.

PR #1398, opened 2021-06-12
Replace git

Foxy7:
@scottleibrand
i've made the changes, less than i'd have guessed (note not tested out), but will use this next time i install.

only line i'm not sure about is this bin/oref0-setup.sh L1049, which has the `git+git://github.com/...` and now looks like:
```
sudo pip install --default-timeout=1000 --user git+https://github.com/mddub/pancreabble.git
```


scottleibrand:
I don't know whether pip supports git+https:// - we'll need someone to test that (and a full install).

scottleibrand:
I individually tested the `pip` and `git clone` commands successfully, and ran oref0-setup via oref0-runagain. Merging to dev where we can use it on the next full install from scratch after reflash.

Issue #1401, opened 2021-07-06
halfBasalExerciseTarget and lowTT settings can lead to negative ISF and Basal


checkatillo:
When we set halfBasalExerciseTarget to (for example) 120 and lowTempTarget of 75, the sensitivityRatio is negative
halfBasal of 100 leads to sensitivity of 0
the code with the calculations is in determine-basal.js:
https://github.com/nightscout/AndroidAPS/blob/65b0c5ccda53e2533c6b987254c0dcc6d27a7661/app/src/main/assets/OpenAPSSMB/determine-basal.js#L205

formula is (halfBasalTarget - 100) / (halfBasalTarget - 100 + target - 100)
in the situation mentioned above we get (120 - 100) / (120 - 100 + 75 - 100) = 20 / (-5) = -4
that leads to negative basal and ISF and calculations are all wrong

setting halfBasal to 100 leads to ration = 0.0 and thus when we calculate ISF, we divide the original by the ratio, so divide by zero

this can happen only on low TT because the numerator is almost guaranteed to be above 0 and the
halfBasalTarget + target - 200 < 0 means that we get negative results with low targets
and low TT usually require low sensitivity, so we may set it as a max available in preferences

I suggest we add the check after the `sensitivityRatio = Math.min(sensitivityRatio, profile.autosens_max);` as follows:
`if (sensitivityRatio <= 0.0) sensitivityRatio = profile.autosens_max;`

**To Reproduce**
1. set halfBasalExerciseTarget to 120
2. set low temptarget lowers sensitivity to true
3. activate TT of 75
4. see the negative ration, ISF and Basal needed

**Expected behavior**
With low TT the ration needs to be high

**Smartphone (please complete the following information):**

**Setup Information (please complete the following information):**
* All pump, CGMs and rigs are affected
* the issue exists in the current master branch

mountrcg:
There are several possibilities to arrive at denominator of 0 in marked formula.
eg:
set halfBasalExerciseTarget to 120
set low tt lowers sensitivity
activate low tt of 80
.. and so on, always symmetric around 100.

SeregaYakovlev:
I confirm these bugs. My screenshots.
Negative InsulinReq, negative senstitivity ratio, negative ISF:
![photo5413536385614328682](https://user-images.githubusercontent.com/52125266/124591740-4ffd5e00-de65-11eb-8d35-d94f66c449a9.jpg)

Deviding to zero: 
[...]

PR #1402, opened 2021-07-06
per #1401 ensure that halfBasalTarget is >=130

scottleibrand:
As discussed in https://github.com/openaps/oref0/issues/1401 a halfBasalTarget of 120 causes a divide by zero, and halfBasalTarget <120 causes negative sensitivityRatio. Anything 121-129 would IMO cause the system to behave way too strongly to very slight changes in temp target, so I propose we hard-code the minimum halfBasalTarget to 130 mg/dL.


SeregaYakovlev:
My autotune_max is 1.2 My ISF is 2.0. The minimum ISF is 2.0 / 1.2 = 1.7. The difference from the minimum ISF and default ISF is 0.3 for me.
ISF can have strongly difference only for children, but OpenAPS is not recommended for toddlers.
I think that the difference 0.5 will easy cover by low sensitivity on high sugar.
Consequently, I wish to limit the half_basal_exercise_target to 121.

scottleibrand:
Merged #1403 instead.

PR #1403, opened 2021-07-08
per #1401 set highest sensitivity on low TT and halfBasal

checkatillo:
as described in #1401 we have a possibility of zero or negative ratio on low TT and halfBasalTarget.
This ensures that we don't get over the autosens_max even in these situations

getting multiplication less or equal to 0 means that we have a really low target with a really low halfBasalTarget
with low TT and lowTTlowersSensitivity we need autosens_max as a value
we use multiplication instead of the division to avoid "division by zero error"


SeregaYakovlev:
@scottleibrand, I think that is the best solution of #1401 bug.

scottleibrand (approved):
Code LGTM. If someone can give it a test on a rig and paste the output, this should be good to merge to dev.

SeregaYakovlev:
@scottleibrand, I removed the file determine-basal.js from src/oref0/lib/determine-basal/ directory and rebooted the Intel Edison. But OpenAps continues to work. Why?

scottleibrand:
Generally, this would indicate that you were using a packaged install, rather than an install linked to the source code in src/.

scottleibrand:
Tested this on a rig, and it doesn't affect sensitivityRatio during normal operation.

PR #1408, opened 2021-08-26
Add autogenerated docstrings summarizing each function

scottleibrand:
From https://github.com/scottleibrand/codex-tools/tree/main/auto-docstring


Foxy7:
@scottleibrand  i like this idea of this. one of biggest barriers to entry is that once you find a new function you have to search everywhere to find an example of it to copy or the initial issue/PR it was developed from.

scottleibrand:
Are you saying that's a barrier to implementing auto-docstrings? Or that's the current situation that auto-docstrings will help alleviate?

Foxy7:
the latter. auto-docstrings will help.

PR #1411, opened 2021-11-14
Bugfixes

jimrandomh:
Significant changes:
 * Check for bad (super slow) RPi nodejs versions, install nvm version if found. This fixes excessive CPU usage and slow looping interval on Raspberry Pi Zero.
 * Add a pair of config settings, `pi_battery_shutdown_percent` and `edison_battery_shutdown_voltage`, for controlling when the rig shuts down due to low battery. This config setting is needed for USB-powered Pi rigs, where the battery level indicator just returns a random number, which might occasionally be zero.

Minor changes:
 * If unable to update with git pull, prompt before aborting `oref0-runagain.sh`. This is mainly for convenience during development; a checkout in some weird branch doesn't necessarily return success when you "git pull".
 * Fix bug that would clobber sshd_config in some circumstances
 * Extend timeout on JS syntax check unit tests
 * Clean up require() usage in IOB unit test, speeding it up enough to pass without timing out on Pi Zero
 * Add bash-unit-test-temp to gitignore

### DIFF
--- a/bin/oref0-cron-every-15min.sh
+++ b/bin/oref0-cron-every-15min.sh
@@ -12,12 +12,24 @@ assert_cwd_contains_ini
 
 # proper shutdown once the EdisonVoltage very low (< 3050mV; 2950 is dead)
 if is_edison; then
-    sudo ~/src/EdisonVoltage/voltage json batteryVoltage battery | jq .batteryVoltage | awk '{if ($1<=3050)system("sudo shutdown -h now")}' &
+    BATTERY_VOLTAGE="$(sudo ~/src/EdisonVoltage/voltage json batteryVoltage battery | jq .batteryVoltage)"
+    echo "Battery level is $BATTERY_VOLTAGE."
+    BATTERY_CUTOFF=$(get_pref_float .edison_battery_shutdown_voltage 3050)
+    if (( "$BATTERY_VOLTAGE" <= "$BATTERY_CUTOFF" )); then
+        echo "Critically low battery! Shutting down."
+        sudo shutdown -h now
+    fi
 fi
 
 # proper shutdown of pi rigs once the battery level is below 2 % (should be more than enough to shut down on a standard 18600 ~2Ah cell)
 if is_pi; then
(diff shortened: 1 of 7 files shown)


bin/oref0-cron-every-15min.sh
```diff
@@ -12,12 +12,24 @@ assert_cwd_contains_ini
 
 # proper shutdown once the EdisonVoltage very low (< 3050mV; 2950 is dead)
 if is_edison; then
-    sudo ~/src/EdisonVoltage/voltage json batteryVoltage battery | jq .batteryVoltage | awk '{if ($1<=3050)system("sudo shutdown -h now")}' &
+    BATTERY_VOLTAGE="$(sudo ~/src/EdisonVoltage/voltage json batteryVoltage battery | jq .batteryVoltage)"
+    echo "Battery level is $BATTERY_VOLTAGE."
```
scottleibrand, line 16: To differentiate this from the percentage-based alert below, maybe have it echo "Battery voltage is $BATTERY_VOLTAGE." here.

lib/profile/index.js
```diff
@@ -66,7 +66,9 @@ function defaults ( ) {
     //, maxRaw: 200 // highest raw/noisy CGM value considered safe to use for looping
     , calc_glucose_noise: false
     , target_bg: false // set to an integer value in mg/dL to override pump min_bg
-  };
```
scottleibrand, line 69: Is this ; needed here?

scottleibrand (commented):
Added a couple of minor comments

lib/profile/index.js
jimrandomh: I didn't mean to change this line, but it doesn't make a difference; semicolons are optional in JS with only rare exceptions.

bin/oref0-cron-every-15min.sh
jimrandomh: Done

tzachi-dar:
I have run the part that updates the node manually, and indeed start up time is much shorter.

PR #1416, opened 2021-12-21
comment out shared-node cp to avoid full disk

scottleibrand:
The various cp *.json $dir_name commands introduced for testing/debugging shared-node are writing hundreds of MB per day.  Commenting out the cp commands to avoid quickly filling up the rigs' disks.


scottleibrand:
@tzachi-dar any reason to do this differently?

tzachi-dar:
This looks good to me.
The files have been copied only to be able to create test data. This lines can probably be deleted at all.

PR #1419, opened 2022-01-05
Improved handling of node installation (& miscellaneous improvements)

ChanceHarrison:
Note: This PR has changed somewhat significantly since inception. The "very minor" changes listed below have been removed to keep the focus on changes to node/npm install improvements. The original text otherwise follows, but please see the comments for relevant context.
---

#### Very Minor Changes
- Escape quotes that are themselves in a quoted string. Before, these quotes would not print, but it did not cause any errors or other unexpected behavior.
- Moved a comment to (what I think is) the right place

#### NodeJS Changes
- Remove the nodesource JS installation (in the case of edison boards) from the install script, as it is already done in the openaps-packages script and is not needed until then
- The biggest change of this PR. Instead of Installing `nodejs` and `npm` from the package manager, we instead install [the node version manager n](https://github.com/tj/n) since in my first attempt to install OpenAPS found the scripts failing due to issues with getting JS setup (particularly with getting `npm` to a proper version and keeping it there). There may be an easier fix, but `n` seems like a more robust solution. I chose `n` over `nvm` because of issues with `nvm` in multi-user/root environments. `n` is an actual binary and does a proper global install of `node` that should avoid those issues. I changed the setup script to reflect the preference of `n` over `nvm`. This change also includes a subtle but important modification to the npm self-update command. It now, instead of updating itself to the latest version of npm, it should update itself to the latest version of npm _that is compatible with the installed version of node_. Trying to install the latest version of npm with our relatively ancient version(s?) of `node` does not work.

Copy and pasted code comment that I (re)moved (in a recent commit) but wanted to save for posterity:
```
# Raspbian Buster has (at the time of writing) npm 5.8 in the repo, which is not compatible with the repo nodejs version of 10.24
# An npm self-upgrade (as is attempted below) is successful (bringing npm to major verison 8), but between there and this script trying to `npm install -g json` npm has somehow reverted to the package manager version and become broken (any invocation of npm immediately crashes with a syntax error)
# At this point, the root cause is unclear and I don't want to deal with it, so I'm deferring to `n`
# `n` instead of `nvm` to avoid issues with nvm not being available to all users (see https://stackoverflow.com/questions/21215059/cant-use-nvm-from-root-or-sudo)
```

If you would like me to squash/rebase any of my commits or remove some (perhaps excessive) code comments, I'd be happy to do so.

Please let me know your thoughts/suggestions. I look forward to making future contributions.

### Testing

I used this modified code to conduct my own install and it worked well for me (a lot better than regular `dev` at the moment). Some of these changes are non-functional and could likely be merged separately, but I could understand the desire for some testing around the JS changes. It shouldn't change anything for existing users who re-run `oref0-setup.sh`; their existing version of node is likely fine. 

Perhaps contributors that are doing fresh installs (and just updating their rigs) can do so with these changes included?

### DIFF
--- a/bin/openaps-install.sh
+++ b/bin/openaps-install.sh
@@ -46,13 +46,6 @@ if cat /etc/os-release | grep 'PRETTY_NAME="Debian GNU/Linux 8 (jessie)"' &> /de
     echo "Jubilinux 0.2.0, based on Debian Jessie, is no longer receiving security or software updates!"
 fi
 
-#Workaround for Jubilinux to install nodejs/npm from nodesource
-if getent passwd edison &> /dev/null; then
-    #Use nodesource setup script to add nodesource repository to sources.list.d
-    curl -sL https://deb.nodesource.com/setup_8.x | bash -
-fi
-
-#dpkg -P nodejs nodejs-dev
 # TODO: remove the `-o Acquire::ForceIPv4=true` once Debian's mirrors work reliably over IPv6
 apt-get -o Acquire::ForceIPv4=true update && apt-get -o Acquire::ForceIPv4=true -y dist-upgrade && apt-get -o Acquire::ForceIPv4=true -y autoremove
 apt-get -o Acquire::ForceIPv4=true update && apt-get -o Acquire::ForceIPv4=true install -y sudo strace tcpdump screen acpid vim locate ntpdate ntp
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -295,7 +295,7 @@ function move_mmtune () {
 function check_nodejs_timing () {
     # Redundant check that node is installed
     # It is installed as part of openaps-packages.sh
-    if ! node --version | grep -e 'v[89]\.' -e 'v1\d\.' &> /dev/null ; then
+    if ! node --version | grep -q -e 'v[89]\.' -e 'v1[[:digit:]]\.'; then
         die "No version of node (>=8,<=19) was found, which is an unexpected error (node installation should have been handled by previous installation steps)"
     fi
 
--- a/bin/oref0-setup.sh
+++ b/bin/oref0-setup.sh
@@ -292,14 +292,11 @@ function move_mmtune () {
     fi
 }
 
-function install_or_upgrade_nodejs () {
-    # install/upgrade to latest node 8 if neither node 8 nor node 10+ LTS are installed
-    if ! nodejs --version | grep -e 'v8\.' -e 'v1[02468]\.' >/dev/null; then
-        echo Installing node 8
-        # Use nodesource setup script to add nodesource repository to sources.list.d
-        sudo bash -c "curl -sL https://deb.nodesource.com/setup_8.x | bash -" || die "Couldn't setup node 8"
-        # Install nodejs and npm from nodesource
-        sudo apt-get install -y nodejs=8.* || die "Couldn't install nodejs"
+function check_nodejs_timing () {
+    # Redundant check that node is installed
+    # It is installed as part of openaps-packages.sh
+    if ! node --version | grep -e 'v[89]\.' -e 'v1\d\.' &> /dev/null ; then
+        die "No version of node (>=8,<=19) was found, which is an unexpected error (node installation should have been handled by previous installation steps)"
     fi
 
     # Check that the nodejs you have installed is not broken. In particular, we're
@@ -314,17 +311,39 @@ function install_or_upgrade_nodejs () {
         echo "Your installed nodejs ($(node --version)) is very slow to start (took ${NODE_EXECUTION_TIME}s)"
         echo "This is a known problem with certain versions of Raspberry Pi OS."
 
-        if prompt_yn "Install a new nodejs version using nvm?" Y; then
-            echo "Installing nvm and using it to replace the system-provided nodejs"
-    
-            # Download nvm
-            curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh | bash
-            # Run nvm, adding its aliases to this shell
-            source ~/.nvm/nvm.sh
-            # Use nvm to install nodejs
-            nvm install 10.24.1
-            # Symlink node into /usr/local/bin, where it will shadow /usr/bin/node
-            ln -s ~/.nvm/versions/node/v10.24.1/bin/node /usr/local/bin/node
+        if prompt_yn "Confirm installation of replacement nodejs/npm versions?" Y; then
+            echo "Attempting to uninstall current nodejs/npm versions (apt-get remove)"
+            sudo apt-get -y remove nodejs npm
+            if [[ -n $NVM_DIR ]]; then
+                echo "Removing nvm ($NVM_DIR)..."
+                echo "(you may wish to optionally remove the nvm-related lines that still exist in ~/.bashrc; this script won't do it for you)"
+                rm -rf "$NVM_DIR"
+            fi
+
+            # Check that there node and npm are no longer available. If they are, warn the user.
+            nodePath=$(command -v node)
+            npmPath=$(command -v npm)
+            if [[ -e "$nodePath" ]]; then
+                echo "Note: A 'node' binary (located at '$nodePath') still exists and may interfere with the new installation of node"
+            fi
+            if [[ -e "$npmPath" ]]; then
+                echo "Note: A 'npm' binary (located at '$npmPath') still exists and may interfere with the new installation of npm"
+            fi
+
+            if [[ ! $(command -v n) ]]; then
+                echo "n already exists on the system, using it to install a new version of node..."
+                sudo n current
+            else
+                echo "Installing n and using it to replace the system-provided nodejs"
+                echo "Installing node via n..."
+                curl -L https://raw.githubusercontent.com/tj/n/master/bin/n -o n
+                # Install the latest version of node that is supported on this platform
+                sudo bash n current
+                # Delete the local n binary used to boostrap the install
+                rm n
+                # Install n globally
+                sudo npm install -g n
+            fi
 
             NEW_NODE_EXECUTION_TIME="$(\time --format %e node -e 'true' 2>&1)"
             echo "New nodejs took ${NEW_NODE_EXECUTION_TIME}s to start"
@@ -730,7 +749,7 @@ if prompt_yn "" N; then
     echo Running apt-get autoclean
     sudo apt-get autoclean
 
-    install_or_upgrade_nodejs
+    check_nodejs_timing
 
     # Attempting to remove git to make install --nogit by default for existing users
     echo Removing any existing git in $directory/.git
--- a/bin/openaps-packages.sh
+++ b/bin/openaps-packages.sh
@@ -30,15 +30,17 @@ else
    sudo apt-get -y install jq || die "Couldn't install jq"
 fi
 
-# Install/upgrade to latest version of node (v10) using apt if neither node 8 nor node 10+ LTS are installed
-if ! nodejs --version | grep -e 'v8\.' -e 'v1[02468]\.' &> /dev/null ; then
-   if getent passwd edison; then
-     # Only on the Edison, use nodesource setup script to add nodesource repository to sources.list.d, then install nodejs (npm is a part of the package)
-     curl -sL https://deb.nodesource.com/setup_8.x | bash -
-     sudo apt-get install -y nodejs=8.* || die "Couldn't install nodejs"
-   else
-     sudo apt-get install -y nodejs npm || die "Couldn't install nodejs and npm"
-   fi
+# Install node using n if there is not an installed version of node >=8,<=19
+# Edge case: This is not likely to work as expected if there *is* a version of node installed, but it is outside of the specified version constraints
+if ! node --version | grep -e 'v[89]\.' -e 'v1\d\.' &> /dev/null ; then
+   echo "Installing node via n..." # For context why we don't install using apt or nvm, see https://github.com/openaps/oref0/pull/1419
+   curl -L https://raw.githubusercontent.com/tj/n/master/bin/n -o n
+   # Install the latest compatible version of node
+   sudo bash n current
+   # Delete the local n binary used to boostrap the install
+   rm n
+   # Install n globally
+   sudo npm install -g n
    
    # Upgrade to the latest supported version of npm for the current node version
    sudo npm upgrade -g npm|| die "Couldn't update npm"
--- a/bin/openaps-packages.sh
+++ b/bin/openaps-packages.sh
@@ -32,7 +32,7 @@ fi
 
 # Install node using n if there is not an installed version of node >=8,<=19
 # Edge case: This is not likely to work as expected if there *is* a version of node installed, but it is outside of the specified version constraints
-if ! node --version | grep -e 'v[89]\.' -e 'v1\d\.' &> /dev/null ; then
+if ! node --version | grep -q -e 'v[89]\.' -e 'v1[[:digit:]]\.'; then
    echo "Installing node via n..." # For context why we don't install using apt or nvm, see https://github.com/openaps/oref0/pull/1419
    curl -L https://raw.githubusercontent.com/tj/n/master/bin/n -o n
    # Install the latest compatible version of node
--- a/bin/openaps-packages.sh
+++ b/bin/openaps-packages.sh
@@ -40,8 +40,8 @@ if ! nodejs --version | grep -e 'v8\.' -e 'v1[02468]\.' &> /dev/null ; then
      sudo apt-get install -y nodejs npm || die "Couldn't install nodejs and npm"
    fi
    
-   # Upgrade npm to the latest version using its self-updater
-   sudo npm install npm@latest -g || die "Couldn't update npm"
+   # Upgrade to the latest supported version of npm for the current node version
+   sudo npm upgrade -g npm|| die "Couldn't update npm"
 
    ## You may also need development tools to build native addons:
    ## sudo apt-get install gcc g++ make


bin/openaps-install.sh
```diff
@@ -46,21 +46,14 @@ if cat /etc/os-release | grep 'PRETTY_NAME="Debian GNU/Linux 8 (jessie)"' &> /de
     echo "Jubilinux 0.2.0, based on Debian Jessie, is no longer receiving security or software updates!"
 fi
 
-#Workaround for Jubilinux to install nodejs/npm from nodesource
```
cluckj, line 49: Is n able to install a compiled x86 version of node?

bin/oref0-setup.sh
```diff
@@ -294,7 +294,7 @@ function move_mmtune () {
 
 function install_or_upgrade_nodejs () {
     # install/upgrade to latest node 8 if neither node 8 nor node 10+ LTS are installed
-    if ! nodejs --version | grep -e 'v8\.' -e 'v1[02468]\.' >/dev/null; then
+    if ! node --version | grep -e 'v8\.' -e 'v1[02468]\.' >/dev/null; then
         echo Installing node 8
         # Use nodesource setup script to add nodesource repository to sources.list.d
         sudo bash -c "curl -sL https://deb.nodesource.com/setup_8.x | bash -" || die "Couldn't setup node 8"
```
cluckj, line 300: Should this also use n?

ChanceHarrison: It don't see why it couldn't. I don't have an edison board so I was a bit hesitant to change things that I couldn't test directly. In theory, this part could be entirely replaced by lines 318-331 of this same file (checking that n is installed, doing so if not, and then installing node).

bin/oref0-setup.sh
```diff
@@ -314,17 +314,21 @@ function install_or_upgrade_nodejs () {
...
+                echo "Installing n and using it to replace the system-provided nodejs"
+                echo "Installing node via n..."
+                curl -L https://raw.githubusercontent.com/tj/n/master/bin/n -o n
+                # Install the latest version of node 10.x.x
+                sudo bash n 10
+                # Delete the local n binary used to boostrap the install
+                rm n
+                # Install n globally
+                sudo npm install -g n
+            fi
```
ChanceHarrison, line 331: I just realized that I don't know how this code will work when node (e.g. from package manager) is already installed. Despite saying it will, it doesn't replace/remove the existing binary. Would `sudo apt remove nodejs npm` or some `dpkg` command work? I would feel uncomfortable leaving a package manager version and trying to write over it, though it's possible that the installed version isn't from the package manager (e.g., what if it's already from nvm?). Or do we need to do something a bit more hacky like what was done before with the symlink on old line 327? This will likely need to change before it can merge.

bin/openaps-install.sh
cluckj: Yep, though it will only install up to v8.

bin/oref0-setup.sh
cluckj: Confirmed it will install via n, but only v8 on the edison.

bin/oref0-setup.sh
cluckj: It's installing to /usr/local/, so it probably won't overwrite apt's install which iirc installs to /usr/bin/

scottleibrand:
Was this designed for and tested on a Raspberry Pi rig? I recently re-flashed and reinstalled all my Edison rigs using `dev`, and that worked fine, so not sure this is needed there. I don't currently have an operational Pi rig, so haven't tested that lately...

ChanceHarrison:
> Was this designed for and tested on a Raspberry Pi rig? I recently re-flashed and reinstalled all my Edison rigs using `dev`, and that worked fine, so not sure this is needed there. I don't currently have an operational Pi rig, so haven't tested that lately...

Yes, this was designed for and tested on a Pi rig (A Pi Zero W, to be precise). Even if the JS changes are not needed for Edison, my hope and expectation is that these changes provide an equally effective solution that would reduce the need to treat the two platforms differently.

scottleibrand:
Ok. LMK when y'all have tested it to your satisfaction on the Pi, and I can re-flash and test it on an Edison.

ChanceHarrison:
I have spent some time thinking about this, and I would appreciate your input on the following:

For our two major architectures, x86 (Edison) and armv6 (Pi 0W), [their support was downgraded to "Experimental"](https://github.com/nodejs/node/blob/master/BUILDING.md#platform-list) as of versions 10 and 12, respectively. So, `n` by default only provides the following versions:

```
$ n --arch x86 --latest
9.11.2
$ n --arch armv6 --latest
11.15.0
```

(On that note, @cluckj can you check and see what the output of `n --latest` says on an Edison? I'm not sure why it only gave you v8 if x86 has support for v9)

I suppose there is no harm in using these old versions and we can leave it at that, but I do want to throw out an alternative for consideration: Use [unofficial builds](https://github.com/nodejs/unofficial-builds/) when they are available (with the old but technically supported versions as a fallback). I believe `n` can be directed to use a [custom source](https://github.com/tj/n#custom-source) and we would  be off to the races with modern Node. Albeit, largely untested and at-your-own-risk Node, which may be antithetical to our goals of building safe and reliable software, but perhaps that would be offset by the value(?) of a modern Node runtime?

Regardless of the decision on that matter, this PR should be ready to move once the following changes are made:

- Updating the call to `n` to install using the `latest` (supported version) alias instead of a specific numbered version
- Uninstalling an existing version (nvm, package manager, etc.) of node/npm if installing via n
- Rebasing my commits to clean up the commit history - is this encouraged?

sharbi:
Hi Chance, unfortunately this build failed on the rpi zero 2. Had exactly the same issues with crashing with each attempt at an npm install. I have the version of npm as 8.1.2 which is failing with node version 10.24.

Global install of latest node via n prior to starting the installation worked. Just wondering about how to do a check for newer  hardware to do a check and install most up to date node.

ChanceHarrison:
Hi @sharbi!

> I have the version of npm as 8.1.2 which is failing with node version 10.24.

That version of npm is too new for that version of node. They won't work together.

I would be eager to figure out how that version of npm got installed. I don't think `n` would install an incompatible version of npm when installing a version of node.

Was this tested on a clean image of Raspbian? If `npm` and/or `node` were already installed from package manager or `nvm`, I would expect them to conflict.

Do you happen to have the output of the script during install?

> Just wondering about how to do a check for newer hardware to do a check and install most up to date node.

I haven't pushed the change yet, but I think the best solution is to just do `n install current`. I don't believe `n` is capable of installing an unsupported version as it does not build from source, it'll grab the latest compatible version for the platform. That should avoid the need for device checks, but it's always a fallback option.

sharbi:
Fresh flash of Raspbian caused the same error. Can't quite figure out how it happens. Output is below:

Installing bash completion script /etc/bash_completion.d/python-argcomplete
npm does not support Node.js v10.24.0
You should probably upgrade to a newer version of node as we
can't make any promises that npm will work with this version.
You can find the latest version at https://nodejs.org/
/usr/local/lib/node_modules/npm/lib/npm.js:32
  #unloaded = false
  ^

SyntaxError: Invalid or unexpected token
    at Module._compile (internal/modules/cjs/loader.js:723:23)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at module.exports (/usr/local/lib/node_modules/npm/lib/cli.js:22:15)
    at Object.<anonymous> (/usr/local/lib/node_modules/npm/bin/npm-cli.js:2:25)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
Couldn't install npm json

Checking versions after running the code shows node 10.24 and unable to see npm version due to node being incompatible. 
Also hasn't globally installed n...so not sure if it's just not going through the if statement. Will have a proper play around later and see if I can figure out what's happening.

ChanceHarrison:
Hi @sharbi 

I think I know why it happens. [This line of openaps-install.sh](https://github.com/openaps/oref0/blob/dfe735b2115b4a035e63c33e8bad529b85b90085/bin/openaps-install.sh#L68) does not download the version of `openaps-packages.sh` on this PR. It downloads it from `$BRANCH` (if provided as an argument when invoking the install script) or `master`. So the changes aren't actually being tested.

My advice: re-test, but instead of piping the install script directly from curl to bash, I would clone the repo, switch to this PR branch (see these [GitHub docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally) for context) (the tl;dr is `git fetch origin pull/{pull_id}/head:{local_branch_name}` where `{pull_id}` is 1419 for this PR, and `{local_branch_name}` can be whatever you want, something like `pr-1419-local` would work. Then you can `git switch` (or `git checkout`, if you prefer) that branch and modify the relevant line:

Old line in openaps-install.sh:
```
curl -s https://raw.githubusercontent.com/openaps/oref0/$BRANCH/bin/openaps-packages.sh | bash -
```
New line:
```
bash {relative or absolute path to openaps-install.sh}
```

Hopefully it is clear what the issue is now and what changes need to be made locally to test them. Ideally this should be a pain point, but we can definitely work on improving this experience if necessary.

sharbi:
Hi cool that all makes sense - should have remembered this from last testing I was doing!

Installation works first time, but then fails setup for some other reason. Strangely then can't re-run installation.

Again n and npm not globally installed and the node version installed was 8.17.0.

ChanceHarrison:
>  fails setup for some other reason (fails with nightscout autoconfigure-device-crud) which feels like a separate issue

Indeed, that does sound like it would be a separate issue, but impossible to say for sure at this point.

> Again n and npm not globally installed and the node version installed was 8.17.0.

Both the version and the installation status defy my expectations. In the current iteration, `n` installs node 10, so I'm not sure how node 8 got installed. Perhaps from package manager? 

If you have the full input/output of a clean installation attempt (e.g. no previous installations) using the scripts from this PR branch, I'd really appreciate it if you could put that in a [GitHub Gist](https://gist.github.com/) and link it here.

sharbi:
Sorry - scrap that. Being a muppet in real time. Solved and now installed with n + node 1.

Setup still not working so will need to start work on that...

bin/oref0-setup.sh
ChanceHarrison: Latest changes include an attempt to uninstall apt and nvm versions. I don't know if that is sufficient, but I would like to think so. Testing would be good.

bin/oref0-setup.sh
ChanceHarrison: Latest changes should work on Edison. I would appreciate a re-test and an update as to what version is installed.

dcacklam:
For buster & bullseye you can just 'apt-get install' it....

ChanceHarrison:
@dcacklam Perhaps the issue has resolved itself with time, or the issue was in my/our implementation all along, but when I opened this PR, the repository versions of `node` and `npm` did not seem to be compatible. Perhaps the wrong packages were being installed or `npm` was being upgraded to the latest version (which would of course be incompatible with the older version of node that gets installed) (see [edit: ][a7c1cd8](https://github.com/openaps/oref0/pull/1419/commits/a7c1cd88630ac57d8ad93dceb9174cbd5321993c)), which I think is an important fix that needs to be merged regardless of the fate of the other changes).

Certainly, perhaps I've fallen prey to the trap that is added complexity, but I do appreciate the platform-agnostic solution that comes with using something like `n`, being almost entirely absolved of concerns re: distro package repository shenanigans. Perhaps it's a short-sighted trade. Perhaps that part can be split into its own PR, as there are other changes here that I believe would be hard to disagree with.

scottleibrand:
@ChanceHarrison Would you like to refactor this to split out the `n` stuff from the other improvements so we can merge it to 0.7.2-dev?

ChanceHarrison:
@scottleibrand Consider it done. Refactored, rebased onto upstream/dev, and history is rewritten to clarify the changes that are being proposed. I don't recommend separating the changes any further, but let's talk if you think otherwise.

I will have to make the time to grab a spare SD card and try installing OpenAPS from scratch again to test my changes once more. When I do so, I'll be sure to write clear-and-concise testing instructions should others want to do the same (at least until the PR is merged into `dev`).

SeregaYakovlev:
I fixed npm and node versions conflict during Openaps 0.7.1 installation: openaps/oref0#1434

ChanceHarrison:
Some time ago, I had the opportunity to test these changes with someone who has an Edision rig who had been attempting to do a clean install of oref0.

I have a few takeaways that I wanted to note here:

- I've shared some `git fetch` commands previously intended to let users pull this PR to use it/test it that had an un-escaped dollar sign that was then interpreted by the terminal as a variable and resulted in creating a local git branch that starts with a dash, which to my knowledge is not an allowed name (and seems to make it non-intuitive to deal with). The fix was to use `git branch -m -- <-old-branch-name> <new-branch-name>`, to rename the branch to something proper, where the `--` clearly delineates command flags from the branch names. Old hat to anyone who is familiar with Linux, but quite a cruel trap for anyone who isn't in the know.
- Attempting to install oref0 from the current master branch is still failing frequently, at least based on messages in the Gitter intend-to-bolus channel. Based on the number of installation failures (independent of rig type) being reported (makes you wonder how many more instances of failed installs are going unreported...), I hope that we can prioritize getting fixes (such as those included in this PR) merged into at least `dev` branch.
- It is difficult for users to test PR requests when they aren't merged into a branch because doing so requires first cloning the openaps/oref0 repo manually, fetching and checking out the PR branch (because GitHub lets you fetch PR directly from the receiving repository, which is pretty cool but is not very intuitive), and then finally editing `openaps-install.sh` as follows before running the install:

Lines 75 has to be changed from:
`curl -s https://raw.githubusercontent.com/openaps/oref0/$BRANCH/bin/openaps-packages.sh | bash -`
to:
`bash [local-path-to-openaps-packages]`
where the bracket portion is replaced with the literal (absolute or relative if you know what the pwd is when that line runs) path is to openaps-packages.sh. If this change is not made, the downloaded file will not have the changes included in the PR (which is only relevant if the PR made changes to the file, which this one has).

(as an aside, I think it might be valuable to instead only have a single file that needs to be downloaded to bootstrap the install instead of having the bootstrap script download another bootstrap script. I think there is value to having the bootstrap script install Git and clone the repo as first order of business so that everything that runs afterwards can use files that have already been stored locally)

Line 76:
`mkdir -p ~/src; cd ~/src && ls -d oref0 && (cd oref0 && git checkout $BRANCH && git pull) || git clone https://github.com/openaps/oref0.git`
can be entirely removed when testing PRs in this manner since the user has already manually cloned the repo and switched to the relevant branch before running the install. If this change is not made and the user does not specify the appropriate (PR) branch when invoking the install, git returns an error saying that checking out $BRANCH would overwrite uncommitted changes. I don't recall if this error causes the script to exit (the script is practically already done by this point anyway).

To re-iterate, I believe the edit to line 76 can be avoided if the appropriate branch name is provided at runtime, but it doesn't completely remove the need (which would be the goal) of manually editing files.

scottleibrand:
Testing this now via:
`curl -s https://raw.githubusercontent.com/openaps/oref0/6fc2e3212497964d52f720aefd515e46dd90b9da/bin/openaps-packages.sh| bash -`

scottleibrand:
That didn't work. Errored out with:
```
ERROR: npm is known not to run on Node.js v8.17.0
You'll need to upgrade to a newer Node.js version in order to use this
version of npm. You can find the latest version at https://nodejs.org/
Couldn't install npm json
```

ChanceHarrison:
> Testing this now via: `curl -s https://raw.githubusercontent.com/openaps/oref0/6fc2e3212497964d52f720aefd515e46dd90b9da/bin/openaps-packages.sh| bash -`

@scottleibrand I believe that success is contingent on the state of the system prior to running setup.

If you did https://github.com/openaps/oref0/pull/1434#issuecomment-1242599247 and then ran this, the error is expected.

The quick workaround is to uninstall any existing nodejs/npm.

Reasoning:

You will find that the setup flow with the proposed changes should succeed in the following cases:

1. There is an existing unbroken`*` installation of nodejs/npm
2. nodejs/npm is not installed

Notably, setup does not adequately detect broken nodejs/npm installations`*`. I already added some logic to try to uninstall nodejs/npm and install known good versions, but that is only invoked as part of `openaps-setup.sh` in the `check_nodejs_timing ` function as per the following block.

```bash
NODE_EXECUTION_TIME="$(\time --format %e node -e 'true' 2>&1)"
    if [ 1 -eq "$(echo "$NODE_EXECUTION_TIME > 10" |bc)" ]; then
    [...prompt the user to uninstall existing node and install with known good process...]
```

I didn't make this the default behavior to avoid fixing what isn't broken (messing with people's existing nodejs/npm installs) , but that might be a worthwhile tradeoff to avoid failed setups. Especially considering that running setup from the current dev/master consistently results in broken nodejs/npm installations`*`.

We can automate the "uninstall nodejs/npm" workaround early in setup to allow for known-good installation to take place. Do you agree?

`*` By a "broken" nodejs/npm installation, I am mostly referring to the case when the latest version of npm is installed when that version does not support whatever version of nodejs is installed. In such a state, npm is not able to fix itself and either nodejs needs to be updated (which isn't feasible for our target platforms; I don't believe they are receiving official nodejs support since v12-ish at the latest) or npm needs to be uninstalled and reinstalled.

scottleibrand:
Ok, trying:
`sudo apt-get remove nodejs`
`curl -s https://raw.githubusercontent.com/openaps/oref0/6fc2e3212497964d52f720aefd515e46dd90b9da/bin/openaps-packages.sh| bash -`

scottleibrand:
That seems to have worked to install node v9.11.2 and 5.10.0, but other stuff I tried with `nvm` seems to have left node v8.17.0 in place as well. Re-flashing the rig and trying again from scratch with your openaps-packages instead of `master`'s.

ChanceHarrison:
You can also take a look at https://github.com/nvm-sh/nvm#uninstalling--removal, nvm seems to make it pretty easy to uninstall, mostly just deleting its directory and removing what it added to `~/.bashrc`. So, whether you go that route or the from-scratch route, I look forward to hearing how it goes.

scottleibrand:
This is working a lot better than `master`, so I'm going to merge it to dev and keep working on getting this rig working. Right now it's erroring out during `go get` with:
```
../go/src/golang.org/x/sys/unix/syscall.go:83:7: undefined: unsafe.Slice
../go/src/golang.org/x/sys/unix/syscall_unix.go:118:7: undefined: unsafe.Slice
../go/src/golang.org/x/sys/unix/sysvshm_unix.go:33:7: undefined: unsafe.Slice
```
That might be an install script issue having to do with updated versions of go, or it might be that this rig's SSD is so far gone it can't keep from corrupting files even during install. I'll probably try another flash and fresh install from dev tomorrow.

PR #1421, opened 2022-01-15
carbsReq pushover improvements

scottleibrand:
This PR cuts down on spurious/duplicate carbsReq pushover noise:

- don't send carbsReq alerts if COB is zero due to clock data parsing error
- run pushover from pump-loop immediately after completing a loop
- check Nightscout for pushover snoozes (created when another rig has already sent a pushover) immediately before sending one

The latter two combine to eliminate the race condition where two rigs would both send pushover alerts if they completed loops within a minute or two of each other (before cron had a chance to run pushover, update the snooze file to Nightscout, and then run ns-loop to download it to the other rig).


Foxy7:
@scottleibrand these changes look good, do you need them testing?

scottleibrand:
I think they are ready for broader testing, yes. We have been using this branch on all our rigs and it’s looking good so far. Have kept the PR open until I stop finding more things to fix. :-)

PR #1423, opened 2022-01-19
Update openaps-install.sh

sharbi:
Include install of node 15.14.0 as current install still forces use of 10.24 when node 15.14 installed
Unable to use npm without updated node version and install crashes despite 15.14 being installed on pi

### DIFF
--- a/bin/openaps-install.sh
+++ b/bin/openaps-install.sh
@@ -76,7 +76,7 @@ fi
 sed -i "s/daily/hourly/g" /etc/logrotate.conf
 sed -i "s/#compress/compress/g" /etc/logrotate.conf
 
-curl -s https://raw.githubusercontent.com/sharbi/oref0/$BRANCH/bin/openaps-packages.sh | bash -
+curl -s https://raw.githubusercontent.com/sharbi/oref0/patch-1/bin/openaps-packages.sh | bash -
 mkdir -p ~/src; cd ~/src && ls -d oref0 && (cd oref0 && git checkout $BRANCH && git pull) || git clone git://github.com/openaps/oref0.git
 echo "Press Enter to run oref0-setup with the current release ($BRANCH branch) of oref0,"
 read -p "or press ctrl-c to cancel. " -r
--- a/bin/openaps-install.sh
+++ b/bin/openaps-install.sh
@@ -58,7 +58,7 @@ apt-get -o Acquire::ForceIPv4=true update && apt-get -o Acquire::ForceIPv4=true 
 
 # Install nvm and update to node 15.14.0
 curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash
-src ~/.bashrc
+source ~/.bashrc
 nvm install 15.14.0
 
 grep "PermitRootLogin yes" /etc/ssh/sshd_config || echo "PermitRootLogin yes" > /etc/ssh/sshd_config
--- a/bin/openaps-install.sh
+++ b/bin/openaps-install.sh
@@ -56,6 +56,11 @@ apt-get -o Acquire::ForceIPv4=true update && apt-get -o Acquire::ForceIPv4=true 
 apt-get -o Acquire::ForceIPv4=true update && apt-get -o Acquire::ForceIPv4=true install -y sudo strace tcpdump screen acpid vim python-pip locate ntpdate ntp
 #check if edison user exists before trying to add it to groups
 
+# Install nvm and update to node 15.14.0
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash
+src ~/.bashrc
+nvm install 15.14.0
+
 grep "PermitRootLogin yes" /etc/ssh/sshd_config || echo "PermitRootLogin yes" > /etc/ssh/sshd_config
 
 if  getent passwd edison > /dev/null; then
--- a/bin/openaps-install.sh
+++ b/bin/openaps-install.sh
@@ -76,7 +76,7 @@ fi
 sed -i "s/daily/hourly/g" /etc/logrotate.conf
 sed -i "s/#compress/compress/g" /etc/logrotate.conf
 
-curl -s https://raw.githubusercontent.com/sharbi/oref0/patch-1/bin/openaps-packages.sh | bash -
+curl -s https://raw.githubusercontent.com/openaps/oref0/$BRANCH/bin/openaps-packages.sh | bash -
 mkdir -p ~/src; cd ~/src && ls -d oref0 && (cd oref0 && git checkout $BRANCH && git pull) || git clone git://github.com/openaps/oref0.git
 echo "Press Enter to run oref0-setup with the current release ($BRANCH branch) of oref0,"
 read -p "or press ctrl-c to cancel. " -r
--- a/bin/openaps-install.sh
+++ b/bin/openaps-install.sh
@@ -60,7 +60,7 @@ apt-get -o Acquire::ForceIPv4=true update && apt-get -o Acquire::ForceIPv4=true 
 curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash
 source ~/.bashrc
 nvm install 15.14.0
-nvm install-latest-npm
+nvm alias default 15.14.0
 
 grep "PermitRootLogin yes" /etc/ssh/sshd_config || echo "PermitRootLogin yes" > /etc/ssh/sshd_config
 
--- a/bin/openaps-install.sh
+++ b/bin/openaps-install.sh
@@ -60,7 +60,7 @@ apt-get -o Acquire::ForceIPv4=true update && apt-get -o Acquire::ForceIPv4=true 
 sudo curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash
 source ~/.bashrc
 nvm install 15.14.0
-npm install -g n
+sudo npm install -g n
 sudo n 15.14.0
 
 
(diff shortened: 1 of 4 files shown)


sharbi:
Still testing this so will remove pull request if any issues

scottleibrand:
Thx for taking this on. When you’re happy with it, LMK and I’ll test it on an Edison rig too.

sharbi:
Hey lots of changes due to testing lots of things but end result is relatively simple.

Needed to update node version using nvm so include the install for this. Problem then becomes difference in path for sudo npm vs just npm. So removed sudo from the npm install in openaps-packages as don't think this is necessary for the install and function still works as expected.

Currently completes initial installation but fails setup as there is another npm call somewhere - just need to find where.

sharbi:
Still some npm calls in the setup that I can't figure out will have a further look later

bin/openaps-install.sh
```diff
@@ -70,7 +76,7 @@ fi
 sed -i "s/daily/hourly/g" /etc/logrotate.conf
 sed -i "s/#compress/compress/g" /etc/logrotate.conf
 
-curl -s https://raw.githubusercontent.com/openaps/oref0/$BRANCH/bin/openaps-packages.sh | bash -
+curl -s https://raw.githubusercontent.com/sharbi/oref0/patch-1/bin/openaps-packages.sh | bash -
```
ChanceHarrison, line 79: This kind of change should not be pushed. It's a fine change to make locally for testing purposes, of course.

bin/openaps-install.sh
```diff
@@ -56,6 +56,12 @@ apt-get -o Acquire::ForceIPv4=true update && apt-get -o Acquire::ForceIPv4=true
 apt-get -o Acquire::ForceIPv4=true update && apt-get -o Acquire::ForceIPv4=true install -y sudo strace tcpdump screen acpid vim python-pip locate ntpdate ntp
 #check if edison user exists before trying to add it to groups
 
+# Install nvm and update to node 15.14.0 also update npm
+sudo curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash
+source ~/.bashrc
+nvm install 15.14.0
```
ChanceHarrison, line 62: Has this line been tested on a Pi 0 (ARMv6) or Edison (x86)?. I fully expect installing any major version 12 or higher (10 on x86) on such architectures with `nvm` will cause it to fall back to building from source. I would call that a non-starter, especially given how long the installation can be already, particularly on low power hardware like the Pi 0. I also haven't heard of or experienced great success with building from source.

I believe the necessity, then, to support older hardware, is to target older versions of node. Ideally we would just install the latest supported version of node, yes?

Out of curiosity, how did you come to the version 15.14.0 in your testing? Why not a lower or higher version?

bin/openaps-packages.sh
```diff
@@ -19,24 +19,11 @@ else
...
-        ## You may also need development tools to build native addons:
-        ##sudo apt-get install gcc g++ make
-fi
 
 sudo pip install -U openaps || die "Couldn't install openaps toolkit"
 sudo pip install -U openaps-contrib || die "Couldn't install openaps-contrib"
 sudo openaps-install-udev-rules || die "Couldn't run openaps-install-udev-rules"
 sudo activate-global-python-argcomplete || die "Couldn't run activate-global-python-argcomplete"
-sudo npm install -g json || die "Couldn't install npm json"
+npm install -g json || die "Couldn't install npm json"
```
ChanceHarrison, line 27: I believe this demonstrates why `nvm` is a sub-optimal; it isn't suited for multi-user/root environments. My solution in #1419 is to use `n` (https://github.com/tj/n) as it does a proper system-wide install that avoids this kind of problem entirely.

bin/oref0-setup.sh
```diff
@@ -745,8 +746,9 @@ if prompt_yn "" N; then
 
     echo Checking oref0 installation
     cd $HOME/src/oref0
+    nvm use 15.14.0
```
ChanceHarrison, line 749: Why wouldn't we just execute this immediately after installing that version?

ChanceHarrison (changes requested):
This seems to largely duplicate the changes I made in #1419, and I am concerned that we are duplicating effort here. I would like to hear your thoughts on building upon the changes that exist already?

To speak to these changes directly; I have some concerns:

- Targeting a fixed version is a dangerous game. As I pointed out in a line comment, I don't expect this to fare well for other platforms. What platform did you develop this on?
- Also mentioned in a line comment, `nvm` doesn't play nicely in multi-user/root environments. On this reason alone I would prefer to use an alternative.

I also have some meta comments on Git matters:

- These changes were based on the `master` branch, which is a long ways behind the `dev` branch. It increases the likelihood that you spend time fixing bugs that have already been fixed or make changes that conflict with changes that are already on `dev`, which complicates merging. It is also hard to compare branches that have different bases (at least, I haven't figured out a good way to do so yet). If you were to continue developing on this branch, I would strongly recommend rebasing your work onto dev.
- As you are making changes, you have a lot of commits, but none of them have a descriptive commit message. It would be helpful to others reviewing your changes if you provided brief descriptions with each commit to describe the specific changes. 
- You may also consider rebasing (likely interactively) your changes to "clean up" your commit history (e.g. squash commits, remove back-and-forth, etc.).

If you're feeling up for it, I would really appreciate the help testing the changes I have made on my branch. My hope is that they have already resolved the issue that you're also trying to fix, but it would be great to know if that's not the case.

sharbi:
Thank you for the feedback, really appreciate it.

Sorry very new to big project + managing commits via git as you can probably tell...

Will work through all your individual points, however, did just look at your suggested changes and agree I've probably doubled up work!

To address individual concerns:
1) Working on an Rpi Zero 2 just now
2) Agree with nvm, the way I was installing n seemed not to work initially so that's why I reverted to nvm

Really appreciate your feedback for git in general, will definitely remove various commits. Also useful to know to work from the dev section. Although sounds like for now there's no need to continue this branch!

Will test what you've done and get back to you :)

bin/oref0-setup.sh
sharbi: This was just me testing that the node version was actually correctly updated, the install should have been working in each shell but for some reason kept coming against npm stating older versions of node being used (which as you've said is an issue with nvm not working for root)

bin/openaps-packages.sh
sharbi: yeah agree, for some reason couldn't get n to work when I was testing

bin/openaps-install.sh
sharbi: This was on a Pi Zero 2, so faster machine for sure.

Agree with it being better to install latest supported, again probably best managed through n.

I actually think I just picked this version up from a passing comment in the intend-to-bolus...not sure why I took it as gospel.

bin/openaps-install.sh
sharbi: will keep in mind! thanks

ChanceHarrison:
Hi @sharbi!

> Sorry very new to big project + managing commits via git as you can probably tell...

No need to apologize! Honestly, I'm pretty new around here too. I think it's awesome that you're taking the time to contribute, and I don't in any way want my feedback to be discouraging. Anyone that is willing to devote their time and effort to supporting a project like OpenAPS is an incredibly valuable resource.

> I've probably doubled up work!

Happens to the best of us! If there's anything in your changes that aren't in the other branch and should be, let's work together to make sure they get there. The last thing I would want is for them to go to waste.

> the way I was installing n seemed not to work initially

Intriguing! Do you happen to remember how you were trying to install it? I'm curious if there are potential pitfalls that I am unaware of regarding our use of `n`, so any insight you might have there would be helpful. Knowing more about it, I (or anyone else) may be able to help you figure out why it wasn't working.

> Really appreciate your feedback for git in general

Glad to hear! It's a bit light on detail, so please feel welcome to reach out if you would like some clarification and/or resources on doing any of what I mentioned.

> Also useful to know to work from the dev section.

You certainly can't be faulted for not knowing. I feel like this is something that should be written in the contribution guidelines/readme (perhaps along with some Git best-practices?).

> Although sounds like for now there's no need to continue this branch!

Certainly no harm in keeping it open for now, especially if there are some changes that aren't reflected in the other branch. At the same time, you have the power to close this (and any other PR you open) as you see fit.

Thank you!

sharbi:
Feedback is much appreciated, just trying to wrap my head around a few things! So it's useful to have the guidance. Probably need to do some practice with working locally and not pushing every change...!!

Think the changes you've made cover everything I was trying to resolve so that's all good.

For some reason when I did curl n_location sudo will still say n not a command? Doesn't seem to have been an issue with your updates so hopefully was just something I was missing in my code. 

Thanks again for your time!

ChanceHarrison:
> Feedback is much appreciated, just trying to wrap my head around a few things!

Totally valid! If there's anything that I (or others) can do to help you in wrapping your head around things, feel free to reach out.

> Probably need to do some practice with working locally and not pushing every change...!!

A lot of practice can be done locally, but there's some parts of the process that inherently involve moving past local and interacting with changes/input from others. I'm sure it will come with time. 

Pushing changes regularly isn't inherently a bad thing, At the same time, I do think it is good to be reasonably sure/happy about your changes before pushing them. Personally, I like to think of it like turning in an assignment of sorts or marking a checkpoint (on top of the atomic checkpoints that are my commits); I want what I push to have good presentation/quality and be reasonably complete. This is likely too high of a bar for several scenarios (e.g., sharing work-in-progress code for collaboration purposes), but may be a helpful perspective nonetheless.

> For some reason when I did curl n_location sudo will still say n not a command?

So when you download `n` from a remote source and try to run it, I believe you need to change the permissions on it to make it executable. It also isn't going to be on your path when you just download it (e.g., to your downloads folder, home directory, whatever), so to run it (until you use npm to do a global install of `n`) you would need to do `sudo ./n {arguments}` to explicitly use the local executable. Let me know if that works!

> Thanks again for your time!

Likewise!

PR #1425, opened 2022-02-03
Update get_profile.py

kakoni:
Fixing get_profile to work with latest nightscout api authentication


scottleibrand:
Does anyone who uses Nightscout token auth want to test this before I merge it to 0.7.2-dev?

PR #1426, opened 2022-02-21
Delete killall-g.sh

Angelica421:


scottleibrand:
Why?

Angelica421 (commented):
True

scottleibrand:
Why are you proposing to delete this file?

scottleibrand (changes requested):
This kind of change needs significantly more explanation, testing, etc.

PR #1427, opened 2022-02-25
Add support for Bullseye (address Python3 default/no python2 pip)

dcacklam:
This is being proposed for merge to master because 'master' is where new-installs (even of dev branch) start.

The changes are as follows:
1) Wait to install pip until in 'openaps-packages.sh'
2) When in openaps-packages.sh, check for 'bullseye' OS version, and if found do the following: change the default python to python2, then manually install pip & use it to install all subsequent python packages (a/o APT which will be installing py3 versions of them). 

A similar, parallels pull will be proposed to address the same issues in DEV, plus to enable aarch64 based GO in the event that someone is running raspbian-bullseye64.


dcacklam:
Errors included. Will resubmit

PR #1428, opened 2022-02-25
Fixes to the install scripts for bullseye & 64bit Raspberry Pi OS installs. 

### LINKED ISSUE #1422: Install on Debian Bullseye fails due to python2 requirements
Bug description:
Using most recent Raspbian Version 11 (Bullseye) the install fails during the line 56:
https://github.com/openaps/oref0/blob/acb0150021e7bf284b6741a9510076a13f690747/bin/openaps-install.sh#L56
Throwing the error: E: Package 'python pip' has no installation candidate
Due to python2 being removed on this version of Raspbian.

**To Reproduce**
1. Flash bullseye to new SD card
2. Run usual installation process
3. See error

**Expected behavior**
Expected installation to complete or at least get through python installations

**Setup Information (please complete the following information):**
* Pump type: Medtronic 722 
* CGM type: Libre with xDrip+
* Rig type: Pi Zero 2 running Raspbian Bullseye
* oref0 version: 0.62 master

**Additional context**
Is python2 required for installation? Or would it be possible just by updating the installation code to python3-pip?

Thanks!

dcacklam:
Fixes the following issues:
openaps-setup trying to install python-pip, which does not exist under bullseye

openaps-packages trying to install python, python-pip, and python-numpy expecting Python2, when Bullseye is python3-based and does not include the python-pip package (just python3-pip).

oref0-setup not having an 'elif' for aarch64/not installing go because of this.  Note that armv61l go will not work on an aarch64 OS - you need the arm64 version.

closes #1422

### DIFF
--- a/bin/openaps-install.sh
+++ b/bin/openaps-install.sh
@@ -53,7 +53,7 @@ fi
 #dpkg -P nodejs nodejs-dev
 # TODO: remove the `-o Acquire::ForceIPv4=true` once Debian's mirrors work reliably over IPv6
 apt-get -o Acquire::ForceIPv4=true update && apt-get -o Acquire::ForceIPv4=true -y dist-upgrade && apt-get -o Acquire::ForceIPv4=true -y autoremove
-apt-get -o Acquire::ForceIPv4=true update && apt-get -o Acquire::ForceIPv4=true install -y sudo strace tcpdump screen acpid vim python-pip locate ntpdate ntp
+apt-get -o Acquire::ForceIPv4=true update && apt-get -o Acquire::ForceIPv4=true install -y sudo strace tcpdump screen acpid vim locate ntpdate ntp
 #check if edison user exists before trying to add it to groups
 
 grep "PermitRootLogin yes" /etc/ssh/sshd_config || echo "PermitRootLogin yes" > /etc/ssh/sshd_config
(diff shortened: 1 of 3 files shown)


cluckj:
Do we also need python3 installed?

dcacklam:
Yes.
Python3 and python3-pip are installed during oref0-setup if not already
present (even on buster and earlier).

The issue is that on Buster and earlier, python2 was the 'default python',
with python3 accessed via the python3 executable not the python one.
On bullseye the 'default' is 3 unless you install  the python-is-python2
and python-dev-is-python2 meta-packages.

Eventually the distros will drop python2 compltely (or at least the python
project wants them to). But for now the above works.

danamlewis:
@dcacklam is this repeat of #1428 or something separate?

dcacklam:
It merges the same changes into master that 1429 does into dev.

The reasoning I had for doing it that way is that when following the
documented install process, the install scripts I was fixing always pull
from master to start even if they are going to be used to install dev.

danamlewis:
Ok, thanks for confirming! I merged the other one to dev after we released 0.7.1; we can include this in the next release once someone else tests. Closing this since #1429 achieves the same thing. Thanks again!

dcacklam:
It will take slightly different steps to test without 1428, as the install
routine will fail on Bullseye or with 64bit OS before it reaches the code
in 1429....

PR #1434, opened 2022-08-10
Node and npm versions conflict fix during the installation of 0.7.1 o…

SeregaYakovlev:
…penaps

Installing the nodejs 8.x package puts npm version 6.13.4. The installation line of the latest version of npm leads to a conflict between the versions of npm and nodejs. Too new npm cannot run on too old nodejs. So I just removed the npm update to the latest version so that the Openaps installation script would not crash.

### DIFF
--- a/bin/openaps-packages.sh
+++ b/bin/openaps-packages.sh
@@ -30,9 +30,6 @@ if ! nodejs --version | grep -e 'v8\.' -e 'v1[02468]\.' &> /dev/null ; then
    else
      sudo apt-get install -y nodejs npm || die "Couldn't install nodejs and npm"
    fi
-   
-   # Upgrade npm to the latest version using its self-updater
-   sudo npm install npm@latest -g || die "Couldn't update npm"
 
    ## You may also need development tools to build native addons:
    ## sudo apt-get install gcc g++ make


scottleibrand:
What OS did you see this problem on? Jubilinux? Raspbian?

SeregaYakovlev:
@scottleibrand, jubilinux 0.3.0

ChanceHarrison:
This is not a Jubilinix-specific issue. It occurs on Raspbian too. Actually, I can't think of any installation that _wouldn't_ run into this issue.

Calling `sudo npm install npm@latest` installs the latest version of npm without any regard for if it is compatible with the installed version of nodejs. This results in npm being in a broken state (when the versions aren't compatible at least, which is happening with greater frequency as the latest versions of npm don't support the old versions of nodejs that are available through package repos for a given rig architecture) and failed installations.

Do note that this has been fixed in #1419 since January (approx.) by changing the command to update npm in a manner which respects compatibility constraints with the installed version of nodejs. Those changes have received a fair bit of testing, too.. I could see this being merged sooner due to its simplicity, but it is somewhat redundant.

scottleibrand:
The reason I asked about the OS is that I haven’t had the same problem flashing and installing fresh rigs with jubilinux. So I assumed it was a Raspbian-specific issue. But if you’re seeing the problem on jubilinux and I’m not, we must be doing something differently.

ChanceHarrison:
I am interested as to why you would not be seeing the issue too. My hypothesis is that the special-case logic that exists currently for [handling Jubilinux nodejs install via nodesource](https://github.com/openaps/oref0/blob/88cf032aa74ff25f69464a7d9cd601ee3940c0b3/bin/openaps-packages.sh#L27) is immune to this issue. I'm not entirely sure.

Are you able to share what version of nodejs (and npm) gets installed when you do a clean Jubilinux install (or even what is running on an existing rig of yours)? I'd be especially interested in what the output of this line running (the subject of this one-liner PR) is for you https://github.com/openaps/oref0/blob/88cf032aa74ff25f69464a7d9cd601ee3940c0b3/bin/openaps-packages.sh#L35

scottleibrand:
My currently running rigs all show:
```
# node -v; npm -v
v8.17.0
6.13.4
```
uname -a shows `3.10.98-jubilinux-edison #3 SMP PREEMPT Sun Aug 13 04:22:45 EDT 2017 i686 GNU/Linux`
and lsb_release -a shows `Debian GNU/Linux 9.13 (stretch)`

ChanceHarrison:
@scottleibrand Thanks for sharing. Are you able to speak to the last part of my previous comment?
> I'd be especially interested in what the output of this line running (the subject of this one-liner PR) is for you https://github.com/openaps/oref0/blob/88cf032aa74ff25f69464a7d9cd601ee3940c0b3/bin/openaps-packages.sh#L35

scottleibrand:
I didn’t run that because I didn’t want to change anything on my working rigs. Is there a read-only command to get the info you’re looking for?

ChanceHarrison:
@scottleibrand Appending `--dry-run` will give the output without making any changes.

ChanceHarrison:
For example, when I run that command on my Pi rig:
```
# npm install npm@latest -g --dry-run
+ npm@8.17.0
added 71 packages from 26 contributors, removed 292 packages and updated 134 packages in 186.328s
```
It updates npm to the latest version. Sure. But on the next invocation of npm:
```
# npm
ERROR: npm is known not to run on Node.js v11.15.0
You'll need to upgrade to a newer Node.js version in order to use this
version of npm. You can find the latest version at https://nodejs.org/
```
npm has to be reinstalled once it's in that broken state.

So, I'm very intrigued if the command would have a different result on your machine. Because my expectation is that this behavior occurs all the same regardless of rig type or how nodejs/npm was installed.

scottleibrand:
```
# sudo npm install npm@latest -g --dry-run
npm WARN notsup Unsupported engine for npm@8.18.0: wanted: {"node":"^12.13.0 || ^14.15.0 || >=16"} (current: {"node":"8.17.0","npm":"6.13.4"})
npm WARN notsup Not compatible with your version of node/npm: npm@8.18.0

+ npm@8.18.0
added 67 packages from 21 contributors, removed 292 packages and updated 138 packages in 57.207s
```

scottleibrand:
I finally have to flash and install a new rig, and ran into this problem. Apparently the node/npm maintainers deprecated something since my last install.

When I try to run your workaround, I get:
```
root@edison-eb5 ~ # sudo npm install npm@latest -g
ERROR: npm is known not to run on Node.js v8.17.0
You'll need to upgrade to a newer Node.js version in order to use this
version of npm. You can find the latest version at https://nodejs.org/
```

ChanceHarrison:
@scottleibrand I believe there has been a misunderstanding.

`sudo npm install npm@latest -g` is _not_ the workaround; it is the flawed line that exists in the scripts currently.

A line that achieves the desired result without potentially resulting in the installation of a version of npm that is not compatible with the installed version of nodejs is: `sudo npm upgrade -g npm`. _That_ is the workaround.

scottleibrand:
That doesn't work either:
```
root@edison-eb5 ~ # sudo npm upgrade -g npm
ERROR: npm is known not to run on Node.js v8.17.0
You'll need to upgrade to a newer Node.js version in order to use this
version of npm. You can find the latest version at https://nodejs.org/
```

ChanceHarrison:
@scottleibrand Once npm is in that state, no npm command (that I know of) will fix it. You will have to reinstall npm manually, update to a sufficiently new version of nodejs (not likely feasible), or just uninstall nodejs/npm and run my updated setup (the setup on master/dev will just cause the same issue again because of `npm install npm@latest -g`).

ChanceHarrison:
I believe this can be closed since #1419 was merged?

PR #1437, opened 2022-10-23
fix(lib/iob): Move value checks up to index.js

### LINKED ISSUE #1436: Excessive limitation checks and logging when Custom Peak time below limit of active insulin curve
**Describe the bug**
When running on an insulin curve = rapid-acting and enabling a custom peak time of 48 min, which is under the defined minimum of 50mg/dl excessive logging will start at every loop displaying :
`JavaScript log: Setting minimum Insulin Peak Time of 50m for,rapid-acting,insulin`
150 times every second for up to 60s resulting in 75% failed loops.
This happened when I switched back from Lyumjev and missed disabling the custom peak time.

**Expected behavior**
change once and display once per loop cycle

**Setup Information (please complete the following information):**
* Pump type Omnipod
* CGM type G6
* Rig type iPhione
* oref0 v0.7.1-dev on FreeAPS-X

**Additional context**
the logging happens in the denomination with OpenAPS, so it is a result of determine-basal in conjunction with the other js-files  meal iob etc, where  I found above mentioned string.

ChanceHarrison:
(Mostly) fixes #1436

The root cause of the above issue is that the logic for using and validating custom insulinPeakTime (and by extension, the errors that are logged from said logic) was all the way down (in the sense of a call stack) in `calculate.js`. This means that the validation logic was running many more times than necessary (and in the case of when the values fall out of the tolerable range, an absurd amount of log messages).

Here's what the typical flow might look like for that:

> `oref0-pump-loop.sh` defines and invokes `calculate_iob`
`calculate_iob` calls `run_remote_command 'oref0-calculate-iob [...]`
`oref0-calculate-iob.js` uses `../lib/iob` as `generate`
`lib/iob/index.js` uses `lib/iob/total.js`(called either 1 or 48 times depending on bool `currentIOBOnly`)
`lib/iob/total.js` uses `lib/iob/calculate.js` (called for each non-future treatment that is more recent than dia_ago)

The changes I made move such value handling (both that in `calculate.js` and in `total.js`; thought the former was the source of pain, both were moved for consistency) to `index.js`, where the handling can be done once and the values can be used many times in `calcualte.js` and `total.js`.

As this change asserts, why would we fetch the values, validate them, and log on error in `calculate.js` (which can, in the worst case be called 48 * the number of treatments that are being evaluated in the DIA times) when we can do so _once_ in `lib/iob/index.js` and just pass the values down?

This is a much better situation, because we only get one log message for each invocation of `lib/iob`.

The fix isn't perfect (and why I say it only mostly fixes #1436), because invoking `oref0-meal` results in somewhere between 15 and 20 `lib/iob` invocations, and thus an equivalent number of log messages. Not ideal, but much more palatable than the original number. 

>`oref0-pump-loop.sh` defines and invokes `refresh_pumphistory_and_meal`
`refresh_pumphistory_and_meal` calls `run_remote_command 'oref0-meal [...]`
`oref0-meal.js` users `../lib/meal` as `generate`
`lib/meal/index.js` uses `lib/meal/total.js` (called once)
`lib/meal/total.js` uses `lib/determine-basal/cob.js` as `detectCarbAbsorption` (called for each treatment + 1)
`lib/determine-basal/cob.js` uses `lib/iob` as `get_iob` (once for each bucketed_data value, seems to be between 15 to 20 times on average)

Arguably a similar (if not more effective) fix for the logging issue would be to have just commented out the lines that were doing the logging entirely, but getting rid of one of the only ways that the user might know their config has out-of-range values didn't feel good.

On that note, I think it might be worth considering how we can get the log message into the pump-loop log instead of just the shared-node log, since I don't think there is much (if any) documentation that directs users to look in the shared-node log.

That, and maybe the pump-loop should just fail on out-of-range config values? It would certainly be more noticeable, but maybe also not desirable since it's an issue that we can correct programmatically by using the min/max allowed value.


scottleibrand:
@mountrcg can you test this and confirm whether it helps with your issue, and otherwise works properly?

ChanceHarrison:
I should probably mention that I'm running with these changes on my rig. I haven't noticed any issues, but I'm happy to run it for a few days and confirm that is still the case.

mountrcg:
just saw this now and will test, great stuff @ChanceHarrison

mountrcg:
warnings have been reduced to:
```
2022-12-08 10:32:53.279824+0100 FreeAPS[36957:2018777] [OpenAPS] JavaScriptWorker.swift - createContext() - 23 DEV: JavaScript log: Warning: adjusting insulin peak time from,48,to a minimum of 50m for,rapid-acting,insulin
2022-12-08 10:32:53.281234+0100 FreeAPS[36957:2018777] [OpenAPS] JavaScriptWorker.swift - createContext() - 23 DEV: JavaScript log: Warning: adjusting insulin peak time from,48,to a minimum of 50m for,rapid-acting,insulin
2022-12-08 10:32:53.282051+0100 FreeAPS[36957:2018777] [OpenAPS] JavaScriptWorker.swift - createContext() - 23 DEV: JavaScript log: Warning: adjusting insulin peak time from,48,to a minimum of 50m for,rapid-acting,insulin
2022-12-08 10:32:53.282804+0100 FreeAPS[36957:2018777] [OpenAPS] JavaScriptWorker.swift - createContext() - 23 DEV: JavaScript log: Warning: adjusting insulin peak time from,48,to a minimum of 50m for,rapid-acting,insulin
2022-12-08 10:32:53.283486+0100 FreeAPS[36957:2018777] [OpenAPS] JavaScriptWorker.swift - createContext() - 23 DEV: JavaScript log: Warning: adjusting insulin peak time from,48,to a minimum of 50m for,rapid-acting,insulin
2022-12-08 10:32:53.284770+0100 FreeAPS[36957:2018777] [OpenAPS] JavaScriptWorker.swift - createContext() - 23 DEV: JavaScript log: Warning: adjusting insulin peak time from,48,to a minimum of 50m for,rapid-acting,insulin
2022-12-08 10:32:53.454443+0100 FreeAPS[36957:2018777] [OpenAPS] JavaScriptWorker.swift - createContext() - 23 DEV: JavaScript log: Warning: adjusting insulin peak time from,48,to a minimum of 50m for,rapid-acting,insulin
```
but thats on the simulator, I might not have enough treatments on it. will test on live data shortly.

mountrcg:
same result on live system, 7 warnings only

mountrcg:
lovely, fixed in my book!

scottleibrand (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "oref0",
-  "version": "0.7.0",
+  "version": "0.7.1",
```
scottleibrand, line 3: I think this is an artifact? Will need to check it before merging.

ChanceHarrison: I'm not sure! All I can say is that it wasn't in my original changeset.

scottleibrand: It was because you based your change on master instead of dev, so shows up as a diff after I rebased to dev.

ChanceHarrison: Ah, of course. My bad on the base branch. Fortunately, it doesn't seem to make a significant difference in this context (beyond the version diff in `package.json`). Let me know if you would like me to do anything on my end.

scottleibrand: I'll probably `git reset --hard origin/dev` and cherry-pick your commits, then git push -f it, unless you beat me to it.

scottleibrand:
Merged via https://github.com/openaps/oref0/commit/abf9c6be91c2726482391a7e40c836e271c10c50

scottleibrand:
I had to revert this: https://github.com/openaps/oref0/commit/10f9e2745a2a3d7de0176fe16275a58e178d07ab

It is causing future IOB predictions to be incorrect (highly negative).
That causes the predBG lines to swoop up and to the right.
That results in (slightly) too much insulin being dosed in some situations.

scottleibrand:
To repro, run oref0-calculate-iob and look at the resulting iob.json. Note the iob values at the very end of the json file. With current dev (and the version previous to the regression), they'll be near zero. With the regression, they'll be highly negative.
`oref0-calculate-iob monitor/pumphistory-24h-zoned.json settings/profile.json monitor/clock-zoned.json settings/autosens.json | jq . > iob.verbose.json`
```
-    "iob": -1.52,
-    "activity": -0.0196,
-    "basaliob": -1.53,
+    "iob": 0.012,
+    "activity": 0.0006,
+    "basaliob": 0.002,
     "bolusiob": 0.01,
-    "netbasalinsulin": -4.45,
+    "netbasalinsulin": 0.2,
```

mountrcg:
re-introduces bug #1436

PR #1438, opened 2022-12-08
convert_bg missing in many statements for rT

mountrcg:
I changed many rT statements for BG's to include `convert_bg function`. However for the `rT.predBGs` in the for-Loop that doesn't work. I would require some help there.

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -695,7 +695,7 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
         }
 
         // set eventualBG based on COB or UAM predBGs
-        rT.eventualBG = eventualBG;
+        rT.eventualBG = convert_bg(eventualBG,profile);  // for FAX needs to be in mg/dL
     }
 
     console.error("UAM Impact:",uci,"mg/dL per 5m; UAM Duration:",UAMduration,"hours");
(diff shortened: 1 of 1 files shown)


lib/determine-basal/determine-basal.js
```diff
@@ -695,7 +695,7 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
         }
 
         // set eventualBG based on COB or UAM predBGs
-        rT.eventualBG = eventualBG;
+        rT.eventualBG = convert_bg(eventualBG,profile);  // for FAX needs to be in mg/dL
```
scottleibrand: What is FAX (edit: FreeAPSX?)?

Are you intending to comment out this line in light of #1439? Looks like it's still uncommented.

scottleibrand: If you want to comment out / revert this one for now until we figure out #1439, the other console.error lines are perfectly safe to change, and I can merge this PR with just those.

mountrcg: yes FAX = FreeAPS-X, there the rT variables are used to populate the swift code

mountrcg: do you have an idea how to convert the prediction values UAM, ZT and the like as well?

mountrcg: > if you want to comment out / revert this one for now

ok, forgot about it - yes, I took it out now

scottleibrand (approved):

scottleibrand:
Thx. All changes are now cosmetic, and should only affect mmol/L users, so merging.

sulkaharo:
Was this tested with Nightscout btw? 😅

scottleibrand:
@sulkaharo which changes? AFAICT this only affects console.error logging, not anything uploaded.

sulkaharo:
Ah right, misread the code and assumed some data is also uploaded

PR #1443, opened 2022-12-18
Compress data

tzachi-dar:
Using nightscout on google cloud there is a limit on the BW that can be used for free.
The maximum free amount that can be used is 1G per month.
It seems that two rigs use that BW in a day, not in a month.

This checkin decreases the amount of data that is used significantly by doing two things:
1) Compressing in a few places where curl was not used to compress.
2) Compress the js usage of the code.
3) Decrease the number of device status entries that are brought for pushover from 100 to 20. It seems to me that 20 entries should cover one hour of data, but I'm not 100% sure how things work.

Without this changes each pushover takes 250k and there is one of this calls every minute. After compressing, we get to ~10k of data. This is still too much, so I have changed this to be 20 entries only.
If that does not make sense, please let me know.

### DIFF
--- a/bin/oref0-pushover.sh
+++ b/bin/oref0-pushover.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+﻿#!/usr/bin/env bash
 
 source $(dirname $0)/oref0-bash-common-functions.sh || (echo "ERROR: Failed to run oref0-bash-common-functions.sh. Is oref0 correctly installed?"; exit 1)
 
@@ -87,7 +87,7 @@ fi
         CURL_AUTH='-H api-secret:'${API_SECRET}
     fi
 
-    if snooze=$(curl -s ${CURL_AUTH} ${URL} | jq '.[] | select(.snooze=="carbsReq") | select(.date>'$(date +%s -d "10 minutes ago")')' | jq -s .[0].date | noquotes | grep -v null); then
+    if snooze=$(curl --compressed -s ${CURL_AUTH} ${URL} | jq '.[] | select(.snooze=="carbsReq") | select(.date>'$(date +%s -d "10 minutes ago")')' | jq -s .[0].date | noquotes | grep -v null); then
         #echo $snooze
         #echo date -Is -d @$snooze; echo
         touch -d $(date -Is -d @$snooze) monitor/pushover-sent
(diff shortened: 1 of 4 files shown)


bin/oref0-pushover.sh
```diff
@@ -80,14 +80,14 @@ fi
 
 #function pushover_snooze {
 # check Nightscout to see if another rig has already sent a carbsReq pushover recently
-    URL=$NIGHTSCOUT_HOST/api/v1/devicestatus.json?count=100
+    URL=$NIGHTSCOUT_HOST/api/v1/devicestatus.json?count=20
```
scottleibrand: Do you use pushover? If not, should we disable this if pushover isn't configured?

scottleibrand: I'm reluctant to reduce this value, as for our use case, with multiple rigs all uploading devicestatus data, we might get more than 20 devicestatus entries between pushovers. I'd have to check. If we can just disable it for people who don't use it, that'd be easier.

sulkaharo:
Note the REST API supports if-modified-since headers, use of which will hugely reduce bandwidth consumption. This is being used in loading the glucose values but could be added to other data as well?

tzachi-dar:
Thanks for all the responders.

Here is some more info about the issue:
After moving to google cloud and running there for 3 weeks without the free tier, I have realized that data consumption is going to be more than 30G in a month. Since the free tier is limited to 1G of traffic, it seems that I would have to pay around 4$ a month for this BW.

That said, I have started looking for where the extra BW is coming from. I have used the nginx logs for that. It was very clear that the highest usage of data was coming from the device status command asking for a status of 100 and also non compressed. I have also changed all other places that are used in my use case to be compressed.
This has reduced the BW (as nginx) reports it to ~0.3G a month.

To be honest, I'm not using pushover, and I have removed this calls from my devices.
Still, Google is charging me for about 3G per month (which is about 10 times better that what I have started with but 10 times worth than what nginx is counting). I have used tcpdump to get a sense of what is going on, and it seems that google calculation are right. There is a very high overhead to opening an SSL connection. 
See attached snipset for a typical https connection:
image

So, it seems that getting to a smaller number requires using less connections and not further compression.

All that said (if my PR will be expected as is), it seems that the expected price is going to be ~30 cents per month. So ROI is going to be very hard.
Looking at the pushover calls, it seems that data is filtered on the last 10 minutes, and that can be done on the server.
I'll change my PR to do that.
If there is a simple way to know if pushover is used, please let me know, and I'll also add that.
Going further probably means that we need to keep one connection alive and do multiple requests on it, but I have a filling this will add complexity to the project.
Another option might be to create a tunnel from the rig to the cloud and use html inside that rig. I'm not sure if it actually works or would save BW. 

So, to sumarize:
If there is a way to know if pushover is used, please let me know.
I'll change the pushover call to ask for 10 minutes of data instead of 20 objects.
Any other ideas that will not make the project more complicated are welcomed. (probably calling less than one minute is the simplest but this might also have influence on total rig performance, so I don't like the idea)

scottleibrand:
Looks like oref0-pushover never gets to the point of doing the NS calls if you don't have pusover_token and pushover_user set in your preferences.json: https://github.com/openaps/oref0/blob/d553394da0294da811a94d91bab2e5469243d93a/bin/oref0-pushover.sh#L21-L24

So if you don't use Pushover, you can remove the pushover optimizations and then we can merge this with the `--compressed` flags etc.

tzachi-dar:
Thanks @scottleibrand I see your point about not calling if pusover_token  is not set.
I have removed this part from the PR.

scottleibrand (approved):

PR #1460, opened 2024-03-25
Fix perms check

inventor96:
Previous to this change, my setup script would fail because the perms checking was limited to only the first permissions group received from Nightscout. In my case, I have to share a token, between xDrip+ and my OpenAPS rig in order to facilitate offline functionality. Rather than create one role with everything in it, I assigned multiple roles to the token.

This change allows the permissions to exist in any of the permissions groups received from Nightscout.

For some context, here's the details of my use case...

Here are the role definitions:
image

I had a token with the `roles` string set to "activity, careportal, cgm, openaps, readable". I tried moving the "openaps" role to be the first one, but Nightscout sorts them alphabetically, so that's not an option.

When the script would make a request to `{MY_HOST}/api/v2/authorization/request/token={MY_TOKEN}`, it would look like this:
```json
{
	"token": "{JWT}",
	"sub": "{SUBJECT}",
	"permissionGroups": [
		[
			"api:activity:create"
		],
		[
			"api:treatments:create"
		],
		[
			"api:entries:create"
		],
		[
			"api:devicestatus:create",
			"api:devicestatus:read",
			"api:entries:create",
			"api:entries:read",
			"api:profile:create",
			"api:profile:read",
			"api:treatments:create",
			"api:treatments:read"
		],
		[
			"*:*:read"
		]
	],
	"iat": 1711357690,
	"exp": 1711386490
}

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -1,6 +1,6 @@
 {
   "name": "oref0",
-  "version": "0.7.0",
+  "version": "0.7.1",
```
inventor96: I originally based this branch on master, but should I rebase it on dev to avoid this difference?

scottleibrand: Yes.

bin/oref0_nightscout_check.py
```diff
@@ -92,21 +92,26 @@ def startup_checks(args):
...
             if perm not in pg:
-              missing.append(perm)
+                missing.append(perm)
 
-        if len(missing)>0:
-           logging.error("The following permissions are missing in Nightscout: %s" % missing)
-           logging.error("Please follow instructions at https://openaps.readthedocs.io/en/master/docs/walkthrough/phase-1/nightscout-setup.md#switching-from-api_secret-to-token-based-authentication-for-your-rig")
-           sys.exit(1)
-        logging.info("All permissions in Nightscout are ok")
+    if len(missing)>0:
```
scottleibrand: Why does this need to be promoted out of the else block? If pg==["*"] doesn't that always mean that no perms will be missing by definition?

inventor96: It doesn't :stuck_out_tongue_winking_eye: In my first local iteration, I also promoted the `missing` array, and this went with it. The promotion is no longer needed in the latest iteration.

inventor96: On this note, should I refactor the admin check to `if "*" in pg:` so the excessive permissions are caught even if there are other perms, too?

scottleibrand (approved):

scottleibrand:
This LGTM. I don't use this functionality, so can't test it myself. Does anyone else want to do so? If not, I'm fine merging it in a week or so based on @inventor96's testing and my code review.

bin/oref0_nightscout_check.py
scottleibrand: I don't know if anyone relies on that functionality, so I'd rather stick to fixing what's broken unless we get a volunteer to help test any further changes.

PR #1462, opened 2024-04-13
Add 2 variables

Jon-b-m:
This PR has 2 objectives.

1: introduce 2 new variables.

2: This will make the Oref0 modules identical to the Oref0 code used in newest version of iAPS (v 4.0.0).Newest iAPS version 4 will be using clean Oref0 code.
Currently only difference is these 2 variables, in lib/determine_basal.js and lib/profile/index.js).

Variable 1: SMB Delivery Ratio.
Introduces a variable for the SMB Delivery Ratio, default is 0.5 (50 % of insulinReq) as before. Min/Max 0.1 to 1.

Variable 2: Minimum Glucose Threshold.
This variable allows to increase the very minimum glucose threshold of 65 mg/dl. 

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -329,8 +329,8 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
 
     // But if profile.threshold_setting is over 65 increase the very minimum glucose threshold
     var minimum_threshold = 65;
-    if (minimum_threshold) {
-        Math.min(profile.threshold_setting, 65);
+    if (profile.threshold_setting) {
+        minimum_threshold = Math.min(profile.threshold_setting, 65);
     }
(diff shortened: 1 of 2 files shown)


lib/determine-basal/determine-basal.js
```diff
@@ -326,7 +326,14 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     }
 
     // min_bg of 90 -> threshold of 65, 100 -> 70 110 -> 75, and 130 -> 85
-    var threshold = min_bg - 0.5*(min_bg-40);
+
+    // But if profile.threshold_setting is over 65 increase the very minimum glucose threshold
+    var minimum_threshold = 65;
+    if (minimum_threshold) {
```
yurique: Is this right? 
The `if` will always pass, and the `Math.min` expression will not have any effect (not assigned to anything).

Jon-b-m: Thanks. I fixed it now

PR #1463, opened 2024-04-13
2 new variables

Jon-b-m:
This PR has 2 objectives.

1: introduce 2 new variables.

2: This will make the Oref0 modules identical to the Oref0 code used in newest version of iAPS (v 4.0.0).Newest iAPS version 4 will be using clean Oref0 code.
Currently only difference is these 2 variables, in lib/determine_basal.js and lib/profile/index.js).

Variable 1: SMB Delivery Ratio.
Introduces a variable for the SMB Delivery Ratio, default is 0.5 (50 % of insulinReq) as before. Min/Max 0.1 to 1.

Variable 2: Minimum Glucose Threshold.
This variable allows to increase the very minimum glucose threshold of 65 mg/dl.

### DIFF
--- a/lib/profile/index.js
+++ b/lib/profile/index.js
@@ -75,6 +75,8 @@ function defaults ( ) {
     , target_bg: false // set to an integer value in mg/dL to override pump min_bg
     , edison_battery_shutdown_voltage: 3050
     , pi_battery_shutdown_percent: 2
+    , threshold_setting: 65 // The minimum glucose threshold
+    , smb_delivery_ratio: 0.5 // Ratio of insulinReq, up to maxBolus, to deliver as SMB, when enabled. Default is 1/2 of insulinReq (0.5)
   }
 }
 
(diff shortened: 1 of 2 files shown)


lib/profile/index.js
```diff
@@ -98,6 +100,8 @@ function displayedDefaults (final_result) {
     profile.offline_hotspot = allDefaults.offline_hotspot;
     profile.edison_battery_shutdown_voltage = allDefaults.edison_battery_shutdown_voltage;
     profile.pi_battery_shutdown_percent = allDefaults.pi_battery_shutdown_percent;
+    profile.threshold_setting = allDefaults.profile.threshold_setting;
+    profile.smb_delivery_ratio = allDefaults.smb_delivery_ratio;
```
scottleibrand, line 104: These don't necessarily need to be displayedDefaults for rig users. I believe adding them to defaults above would be sufficient to allow iAPS to display them (or not) as y'all see fit?

Does anyone have an opinion on whether they should also be displayedDefaults for OpenAPS rig users? We generally have trie to avoid putting too many customizations "up front" for people to feel they should tweak them. Do the defaults here seem good enough for most users that these can be buried deeper to change if people really need to? Or do we need to make them more visible to encourage people to experiment with / tweak them?

PR #1464, opened 2024-04-18
rT.reason missing colons (:)

Jon-b-m:
In some places a space is used instead of a colon. This probably has no impact on OpenAPS users, but for iAPS the colon can be used as a separator when reading the reasons-string (the variables): reasons.components(separatedBy: ":")

This would also make the format of the OpenAPS variables in the NS OpenAPS pill more homogenous.

This change isn't super important as I'm currently using a workaround for IAPS to read also the reasons with space separator, but at the same time this PR wouldn't brake anything.


scottleibrand (approved):
This looks like a simple and straightforward consistency/readability improvement to me, which should be strictly cosmetic, so I'm good with merging it straight to dev to test there.

danamlewis:
LGTM for dev as-is

PR #1465, opened 2024-04-18
2 new variables

Jon-b-m:
This PR has 2 objectives.

1: introduce 2 new variables.

2: This will make the Oref0 modules identical to the Oref0 code used in newest version of iAPS (v 4.0.0), as the newest iAPS version 4 will be using "clean" Oref0 code. Currently only difference is these 2 variables, in lib/determine_basal.js and lib/profile/index.js).

Variable 1: 
SMB Delivery Ratio. Introduces a variable for the SMB Delivery Ratio, default is 0.5 (50 % of insulinReq) as before. Min/Max 0.1 to 1.

Variable 2: 
Minimum Glucose Threshold. This variable allows to increase the very minimum glucose threshold of 65 mg/dl.

Updated the PR with removal of the variables from allDefaults.

### DIFF
--- a/lib/determine-basal/determine-basal.js
+++ b/lib/determine-basal/determine-basal.js
@@ -326,7 +326,14 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     }
 
     // min_bg of 90 -> threshold of 65, 100 -> 70 110 -> 75, and 130 -> 85
-    var threshold = min_bg - 0.5*(min_bg-40);
+
+    // But if profile.threshold_setting is over 65 increase the very minimum glucose threshold
+    var minimum_threshold = 65;
+    if (profile.threshold_setting) {
+        minimum_threshold = profile.threshold_setting;
+    }
+    var threshold = Math.max(min_bg - 0.5*(min_bg-40), minimum_threshold);
+    process.stderr.write("Glucose threshold: " + convert_bg(threshold,profile));
 
 // If iob_data or its required properties are missing, return.
 // This has to be checked after checking that we're not in one of the CGM-data-related error conditions handled above,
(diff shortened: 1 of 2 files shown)


scottleibrand:
Thanks for updating it to remove the new preferences from displayedDefaults vs. https://github.com/openaps/oref0/pull/1463: that simplifies things.

I don't have any objections to allowing people with familiarity with the code to override these two settings in this way.

The code for how these are implemented (such as the Min/Max 0.1 to 1 for smb_delivery_ratio and only being able to increase the Minimum Glucose Threshold, not decrease it) both look like sensible safety guardrails to me.

Does anyone want to do on-rig testing before merging this to dev? If not, I'm happy to merge it to dev and test it there, as it looks quite straightfoward.

scottleibrand (approved):

danamlewis:
I would prefer an off-body rig tester with this to ensure no edge cases pop up before this goes into dev. 

(Please feel free to step up and volunteer to check this, otherwise it'll go on my/our list to check but it's a long list so if another volunteer has time/attention that will speed this up)

lib/determine-basal/determine-basal.js
```diff
@@ -326,7 +326,14 @@ var determine_basal = function determine_basal(glucose_status, currenttemp, iob_
     }
 
     // min_bg of 90 -> threshold of 65, 100 -> 70 110 -> 75, and 130 -> 85
-    var threshold = min_bg - 0.5*(min_bg-40);
+
+    // But if profile.threshold_setting is over 65 increase the very minimum glucose threshold
+    var minimum_threshold = 65;
```
MikePlante1: Should this be 60? I think the lowest you can set `min_bg` to is `80`, and `80 - 0.5*(80-40) = 60`

lib/profile/index.js
```diff
@@ -75,6 +75,8 @@ function defaults ( ) {
     , target_bg: false // set to an integer value in mg/dL to override pump min_bg
     , edison_battery_shutdown_voltage: 3050
     , pi_battery_shutdown_percent: 2
+    , threshold_setting: 65 // The minimum glucose threshold
```
MikePlante1: 60 here too? (see my other comment)

lib/determine-basal/determine-basal.js
Jon-b-m: Changed to 60 now

lib/profile/index.js
Jon-b-m: changed to 60 now

PR #1466, opened 2024-04-28
use archive.debian.org for stretch sources.list

scottleibrand:
The sources.list that shipped with jubilinux-stretch is no longer valid, so we have to use archive.debian.org instead.


scottleibrand:
Tested successfully on a freshly flashed rig via:
```
curl https://raw.githubusercontent.com/openaps/oref0/stretch-archive/bin/openaps-install.sh > /tmp/openaps-install.sh
bash /tmp/openaps-install.sh
```

PR #1468, opened 2024-05-04
WIP: Make glucose-get-last faster

thomasvargiu:
Make `glucose-get-last` extremely with minimum loops and ignoring older useless data.


mountrcg:
Do you have a time estimate of the gain @thomasvargiu ?

thomasvargiu:
@mountrcg not really. I tried it on iAPS, I thought it was very slow but currently it isn't a real problem...
But the function was something like O(n^3) and now it's just O(n).

PR #1471, opened 2024-06-03
Changed to python3-pip

inventor96:
When setting up a new version of my rig, I encountered an error about python-pip being replaced by python3-pip. For reference, here's some applicable terminal interactions:

```
pi@loop:~ $ sudo apt install python-pip
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Package python-pip is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  python3-pip

E: Package 'python-pip' has no installation candidate
```
```
pi@loop:~ $ python --version
Python 3.9.2
```
```
pi@loop:~ $ cat /etc/os-release 
PRETTY_NAME="Raspbian GNU/Linux 11 (bullseye)"
NAME="Raspbian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=raspbian
ID_LIKE=debian
HOME_URL="http://www.raspbian.org/"
SUPPORT_URL="http://www.raspbian.org/RaspbianForums"
BUG_REPORT_URL="http://www.raspbian.org/RaspbianBugs"
```

I changed the package name, and it seems to be working so far.

### DIFF
--- a/bin/openaps-packages.sh
+++ b/bin/openaps-packages.sh
@@ -12,7 +12,7 @@ apt-get install -y sudo
 sudo apt-get update && sudo apt-get -y upgrade
 ## Debian Bullseye (Raspberry Pi OS 64bit, etc) is python3 by default and does not support python2-pip.
 if ! cat /etc/os-release | grep bullseye >& /dev/null; then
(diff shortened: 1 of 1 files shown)


inventor96:
Just added a few more changes to continue support for the existing code. I had to explicitly install these additional packages in order for the `bin/bt-pan` script to work.

